Issue #15, opened 2015-07-23
About local.h


kanflo:
I proposed the following change to esp_iot_rtos_sdk wrt wifi SSID and password management. I find it rather attractive as there will be a file containing dummy SSID/pw following a git clone and changes to that file will be ignored by git. I'd be happy to incorporate that change here if more people than me likes the idea :)

https://github.com/kanflo/esp_iot_rtos_sdk/commit/3d0f7be119388ebdc3517a534b3adb2f895e59b0

projectgus:
Hi,

Sorry for the slow reply on this, I wanted to think it over before I replied. It's an interesting approach, I never thought about keeping the file in git like this and just telling the local client to ignore it.

I am a bit uncomfortable with the way local.h is implemented at the moment - you may have seen already, but the makefile checks if it's present and then adds it to the preprocessor flags:
https://github.com/SuperHouse/esp-open-rtos/blob/master/common.mk#L246

... which is kind of icky "magical behaviour", which I don't like. And the makefile is already plenty complex and doesn't really need more weird one-off sections! This approach seems more explicit, which I like.

The other nice thing about it is that "PROGRAM_DIR/include" is placed ahead of root/include/ in the include flags, so if you want per-program ssid_config.h then you can just create one and put it there, and the search path will find it before the global one.

So, yes - I'd happily accept a PR that removed local.h and used this instead. Thanks!


Angus

kanflo:
Great, what do you think of this? https://github.com/SuperHouse/esp-open-rtos/pull/16 
I like that you can have separate credentials for  different projects.

Johan

projectgus:
(Closing this issue as it's covered by #16)

PR #16, opened 2015-07-28
WiFi credentials

kanflo:
PR for issue https://github.com/SuperHouse/esp-open-rtos/issues/15

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,11 +21,14 @@ git clone --recursive https://github.com/Superhouse/esp-open-rtos.git
 cd esp-open-rtos
 ```
 
-* To build any examples that use WiFi, create a file "local.h" in the top-level directory and add two macro defines to it:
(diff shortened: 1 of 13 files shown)


projectgus:
This looks really good, thanks.

The only thing I noticed is that the #warning line in ssid-config.h is commented out. Is that by accident? I think the warning is a good idea.

Cheers,

Angus

kanflo:
Well that was rather sloppy of me, the warning should be there. I'll revisit the PR later this week.

kanflo:
I amended my previous commit, removing the #warning comment from include/ssid_config.h

projectgus:
Thanks again for these, sorry for the slow reply!

Have merged the header fix and the WiFi credentials fix. I'll open a new PR to discuss/merge the UART changes.

Issue #17, opened 2015-07-30
C++ support and MQTT


mikejac:
This it not an issue as such :-)

1) How do you feel about having the ability to use C++ within this project? I have forked esp-open-rtos (https://github.com/mikejac/esp-open-rtos) and made some initial steps to get C++ supported. However, I'm in no way an Makefile/make expert, so would you take a look at the /common.mk file (lines 183 - 187 I believe) and perhaps think of a way to get .cpp files compiled with their own compiler flags (CXXFLAGS).

2) How about adding the Eclipse Paho MQTT client library to esp-open-rtos? Already got it "ported" locally.

kanflo:
I wouldn't mind C++ support. Do global constructors get called automagically? I saw someone else add C++ support and included code for this.

Adding MQTT sounds great. @projectgus is working on adding an "extras" or "components" directory where a project can select modules/drivers to be included. When that work is done, an addition of MQTT would fit nicely.

mikejac:
AFAIK, no, global constructors does not get called. @jcmvbkbc had some ideas here: http://www.esp8266.com/viewtopic.php?f=9&t=478&sid=e39885fb994d80e87a2fb24c56e58e8f&start=5.

I've used that technique "outside" esp-open-rtos. There seem to be something with overridden virtual functions that doesn't work 100% - needs more testing.

projectgus:
Sorry @mikejac, I thought I replied to this post on Friday but it seems my reply got eaten somewhere. :(

> AFAIK, no, global constructors does not get called

Most of the building blocks for open sourcing the startup code (#2) are in place, so once we have that it will be possible to call global constructors properly at the right place in the startup sequence, and everything should "just work". At the moment the best solution is probably calling them manually from user_start, as mentioned.

>I'm in no way an Makefile/make expert, so would you take a look at the /common.mk file (lines 183 - 187 I believe) and perhaps think of a way to get .cpp files compiled with their own compiler flags (CXXFLAGS).

Yeah, common.mk (especially the non-recursive-make dynamic generation parts) gets pretty `$$($(nasty))`!

In 30877f4ef056 (ota branch) I recently added support for assembling .s files. C++ support will probaby look similar, you may need to generate a `$$($(1)_CXX_ARGS)` with the C++ compiler flags though (you can probably model this on `$(1)_CC_ARGS`. Does that help?

[...]

Issue #2, opened 2015-06-02
Open source startup code


projectgus:
At the moment the startup code is provided in binary libmain. The reset vector is in ROM which performs low level initialisation then maps the SPI flash and calls into the entry point at `call_user_start`.

Startup code includes:
* "standard" libc-style setup of memory & interrupts
* calling into libpp for radio calibration
* setup of the FreeRTOS task which handles wifi functions.
* Other things?

As a first step, this process needs to be documented in detail.

It should be possible to write open source routines that give more fine-grained control to the program author.

Some more details (ROM & IoT SDK based, but could be extended to document RTOS startup code):
https://github.com/esp8266/esp8266-wiki/wiki/Boot-Process

projectgus:
First stage is enabling SPI flash cashing so functions from flash can be called (obviously).

@pvvx has a decompiled reconstruction of that sequence here (in Russian):
http://esp8266.ru/forum/threads/revers-sdk-espressif-i-rom-bios-dlja-sozdanija-otkrytogo-sdk.292/#post-5190

pfalcon:
> setup of the FreeRTOS task which handles wifi functions.

Can we make FreeRTOS be optional at all? Both because it's extra bloat and because of its questionable licensing.

pfalcon:
Another effort: http://richard.burtons.org/2015/05/17/decompiling-the-esp8266-boot-loader-v1-3b3/

projectgus:
> Can we make FreeRTOS be optional at all? Both because it's extra bloat and because of its questionable licensing.

I think this is something that might become possible. there are quite a few FreeRTOS-specific calls in the binary libraries, but most of them are in libmain. Some shims some other symbols (mostly mutexing) and it might be possible.

I've thought that Contiki would be interesting to try to port at some point in the future, so having an OS-agnostic base some day would be great from that point of view too.
 
At the very least, the current set of goals would form a solid base for that kind of porting.

Re: OTA bootloader support & rBoot, yes this is an excellent reference thanks. I'll add another issue to track OTA support.

pfalcon:
> because of its questionable licensing.

To clarify, by "questionable licensing" of FreeRTOS I of course mean not GPL, but exception 2 to it, which, arguably, make the RTOS license not an open-source license. Summary of my concerns are expressed here: https://github.com/micropython/micropython/issues/1284#issuecomment-105486744

[...]

PR #24, opened 2015-08-10
C++ Support

projectgus:
This is a PR based on @mikejac's work mentioned in #17, plus tweaks to the Makefile and support for global constructors (currently hacked in just before the FreeRTOS scheduler starts, can be put in a better place once #2 is done.)

I don't usually do C++ on microcontrollers so I'm not sure what else is necessary for this to be useful - new/delete operators? STL support?


mikejac:
Great :-)

1) Yes to new/delete operators (have them locally - and here: http://pastebin.com/Wj1Cyvt6).
2) Don't think so for STL support. Would be really nice, but that's probably too much for a device like the ESP8266.

igrr:
Some parts of STL work quite well even on cores with much less memory than 8266. In the 8266 arduino core and libraries i use `<algorithm>` for some algorithms, <memory> for `std::shared_ptr`, and <function> for `std::function`, `std::bind` and other goodness. Others have even used `std::vector` without apparent issues.

Apart from things you already have, and new/delete operators mentioned above, you might also need some error-handling functions like `__throw_bad_function_call`, `__throw_length_error`, `__throw_bad_alloc`, `__cxa_deleted_virtual`, `__cxa_pure_virtual` — just make them call `abort` or something equivalent. If you want people to use local static variables with non-zero initializers you will also need `__cxa_guard_acquire`, `__cxa_guard_release`, and `__cxa_guard_abort`.

projectgus:
Thanks @igrr, appreciate the insight as always. :)

I see Arduino compiles with -fno-exceptions as well, same as the support @mikejac contributed. Just curious, is this just for performance & code size reasons or do C++ exceptions break the current gcc call0 ABI support?

jcmvbkbc:
> Just curious, is this just for performance & code size reasons or do C++ exceptions break the current gcc call0 ABI support?

C++ exceptions work in call0 ABI, with the only possible exception (put not intended) of stack unwinding over the signal frame, which must be implemented in OS-specific way anyway.
But they pull a lot from libgcc, so that in my tests empty program that only throws one exception doesn't fit into esp8266 IRAM. It may work if libgcc is relocated to FLASH though.

projectgus:
Interesting, thanks @jcmvbkbc. The esp-open-rtos linker script only puts the [integer support functions (*i3.o) into IRAM](https://github.com/SuperHouse/esp-open-rtos/blob/master/ld/eagle.app.v6.ld#L235), so it's probably OK from the libgcc point of view.

Still might not be a super useful feature in the esp8266 context, but I'll give it a try anyhow.

projectgus:
Realised we could fix the weak linking problem by just being intelligent about library order in the linker arguments. Not sure why I didn't realise that earlier.

If you rebase this pull against current master with 0090153, it will link correctly without any init function tricks because the strong-linked _read_r is seen by the linker first.

However to be honest I'm still not sure about merging this, I think a driver that uses the FIFO buffer registers will probably meet most people's needs (128 byte buffer) and has basically no resource overhead.

projectgus:
(Sorry, commit 0090153 and the follow-up comment were meant for pull #31 not this one. Please ignore.)

Issue #8, opened 2015-06-02
Register access header


projectgus:
Flesh out https://github.com/SuperHouse/esp-open-rtos/blob/master/include/esp8266.h to provide register masks, values for known ESP8266 registers - GPIO, UART, etc.

Try to avoid clunky WRITE_PERI_REG-style macros, it should be possible to write clean-looking code that accesses registers directly.

pfalcon:
Please consider going data-driven way of providing a description of hardware registers, from which various "sources" and "headers" can be generated. https://github.com/pfalcon/libperipha is dedicated to such approach, a description example: https://github.com/pfalcon/libperipha/blob/master/arm/cortex-m/nrf51/gpio.yaml

projectgus:
Hi Paul,

Thanks for the suggestion and the link. That's a really nice way of generating headers in different styles.

The only sticking point for me is that it adds another dependency, Python. Not a big deal for Linux users like us, but a minor hassle for Windows users and possibly OS X users depending on their version.

I'm hoping that esp-open-rtos will only require GNU Make and a few base utils, besides the cross-toolchain. This makes it easiest for the most number of people to get up and running and contribute.

Once the known hardware register set becomes reasonably stable then maybe this is something that could be revisited and adopted, with the idea being that you'd only need Python if you wanted to change the register headers.

Cheers,

Angus

pfalcon:
> The only sticking point for me is that it adds another dependency, Python.

It adds a dependency on a Python in the same way as gcc adds a dependency on dejagnu. This is 100% true, right? Dejagnu is required to run testsuite for gcc, so you cannot imagine gcc without dejagnu. But most people don't even know about that - they're given gcc binaries and they run them. Just the same here, layman people don't even need to know that headers are generated from a description, they just compile using them.

 > I'm hoping that esp-open-rtos will only require GNU Make and a few base utils, besides the cross-toolchain. 

Cool, just like I write above.

> Once the known hardware register set becomes reasonably stable then maybe this is something that could be revisited and adopted, with the idea being that you'd only need Python if you wanted to change the register headers.

Sorry, it doesn't work like that. When something will be reasonably stable (which is usually never for amatuer reverse engineering project - hardware gets obsolete first), then nobody will spend any time on it. wm8505 (which is first time we talked, to remind) is "stable", and nobody does anything on it.

[...]

PR #27, opened 2015-08-20
Convert headers to new *_regs.h format

foogod:
As discussed in Issue #8, this change moves to a parameterized-macro and struct-based approach to defining MMIO access registers.  It also renames a number of structs and bit constants (and a couple of related typedefs) to use more consistent naming, etc.

I also went ahead and broke registers.h out into individual .h files for each peripheral type (registers.h now just #includes all the individual *_regs.h files)

### DIFF
--- a/core/include/esp/gpio_regs.h
+++ b/core/include/esp/gpio_regs.h
@@ -59,12 +59,9 @@ struct GPIO_REGS {
     uint32_t volatile RTC_CALIB_RESULT; // 0x70
 } __attribute__ (( packed ));
 
-/* Double-check the structure size to make sure the compiler hasn't done
- * something strange (or somebody typoed in the struct definition, etc)
- */
 _Static_assert(sizeof(struct GPIO_REGS) == 0x74, "GPIO_REGS is the wrong size");
 
-/* Bit mapping for CONF[i] registers */
+/* Details for CONF[i] registers */
 
 /* GPIO.CONF[i] control the pin behavior for the corresponding GPIO in/output.
  *
@@ -109,7 +106,7 @@ typedef enum {
     GPIO_INTTYPE_LEVEL_HIGH = 5,
 } gpio_inttype_t;
 
-/* Bit mapping for PWM register */
+/* Details for PWM register */
 
 #define GPIO_PWM_ENABLE          BIT(16)
 #define GPIO_PWM_PRESCALER_M     0x000000ff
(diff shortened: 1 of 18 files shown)


core/include/esp/gpio_regs.h
```diff
@@ -0,0 +1,130 @@
...
+ *
+ * GPIO_CONF_WAKEUP_ENABLE (boolean)
+ *     Can an interrupt contion on this pin wake the processor from a sleep
+ *     state?
+ *
+ * GPIO_CONF_INTTYPE (multi-value)
+ *     Under what conditions this GPIO input should generate an interrupt.
+ *     (see gpio_inttype_t enum below for values)
+ *
+ * GPIO_CONF_DRIVER_ENABLE (boolean)
```
projectgus: I see you changed this from ...DRIVER_PUSH_PULL back to DRIVER_ENABLE, which I think is what the SDK calls it (?)

Do you have any other thoughts about this? AFAIK what it does is best described as setting a push/pull mode (ie drive high or drive low)? Are you aiming to keep terminology close to what Espressif call each register?

core/include/esp/registers.h
```diff
@@ -14,295 +14,33 @@
...
+#include "esp/wdt_regs.h"
+#include "esp/dport_regs.h"
 
 /* Register base addresses
 
    You shouldn't need to use these directly.
  */
 #define MMIO_BASE   0x60000000
-#define DPORT_BASE  0x3ff00000
+//#define DPORT_BASE  0x3ff00000
```
projectgus: Are you commenting these out as they are migrated to the new register system?

projectgus:
This looks fantastic, thank you! Definite improvement.

I left a couple of minor comments in commits above, but I think this is basically ready to merge. Unless anyone has other comments then I'll probably pull it down and run it on some hardware here later today, then merge.

core/include/esp/gpio_regs.h
foogod: I don't actually fully remember my reasoning for this one.. it may just have been a side-effect of the fact I was sorta merging both your existing stuff and the Espressif headers to make sure nothing got left out..  The more I think about it, I think I sorta agree that "DRIVER_ENABLE" is not the most descriptive name for what it's doing..  I have no problem switching it back to PUSH_PULL (and will see about updating the PR to include that).

On a side-note, when looking at this more closely, I just realized that it looks like this setting in the existing esp-open-rtos headers actually has the values reversed from the Espressif headers (GPIO_DRIVE_OPEN_DRAIN sets bit 2 and GPIO_DRIVE_PUSH_PULL clears it, whereas Espressif labels having bit 2 set as GPIO_PAD_DRIVER_ENABLE).  Was this intentional or accidental?

core/include/esp/registers.h
foogod: Yup.. the *_BASE defines for those are now included in the corresponding *_regs.h file.  I considered just deleting them, but I think commenting them out makes for a handy "checklist" until we finish with all of them, so decided to just leave them commented..

PR #30, opened 2015-08-25
Added bmp180_i2c example

Angus71:
Hi,

I've written a more "advanced" example I wanted to share with the public, it also contains a driver for the BMP180/BMP085 digital pressure sensor, which can be used as an example how to use external devices and distribute data between tasks.

Let me know if this is okay with you (License, reference to external I2C lib etc.)

I'm not sure if it makes more sense to create an external repository for the BMP180 driver (Like the one for the I2C driver) and link it also into the example, but this would lead to an even more polluted .gitmodules file.

Regards

  Angus71

### DIFF
--- a/examples/bmp180_i2c/Makefile
+++ b/examples/bmp180_i2c/Makefile
@@ -1,0 +1,3 @@
+PROGRAM=BMP180_Reader
+EXTRA_COMPONENTS = extras/i2c extras/bmp180
+include ../../common.mk
--- a/examples/bmp180_i2c/README.md
+++ b/examples/bmp180_i2c/README.md
@@ -1,0 +1,7 @@
+# I2C / BMP180 Example
+
+This example references two addtional drivers [i2c](https://github.com/kanflo/esp-open-rtos-driver-i2c) and [bmp180](https://github.com/Angus71/esp-open-rtos-driver-bmp180), which are provided in the `../../extras` folder.
+
+If you plan to use one or both of this drivers in your own projects, please check the main development pages for updated versions or reported issues.
+
+To run this example connect the BMP085/BMP180 SCL to GPIO0 and SDA to GPIO2.
--- a/examples/bmp180_i2c/bmp180_i2c.c
+++ b/examples/bmp180_i2c/bmp180_i2c.c
@@ -1,0 +1,132 @@
+/* Simple example for I2C / BMP180 / Timer & Event Handling
+ *
+ * This sample code is in the public domain.
+ */
+#include "espressif/esp_common.h"
+#include "espressif/sdk_private.h"
+
+#include "FreeRTOS.h"
+#include "task.h"
+#include "timers.h"
+#include "queue.h"
+
+// BMP180 driver
+#include "bmp180/bmp180.h"
+
+#define MY_EVT_TIMER  0x01
+#define MY_EVT_BMP180 0x02
+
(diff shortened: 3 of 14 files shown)


kanflo:
Nice! Although the wiki hasn't been updated to reflect this yet, drivers and other additions should go into the ```extras``` directory with an example under ```examples```. The i2c driver. will soon be added to ```extras```. 

I understand this might feel like a bummer as you probably feel you have completed your commit, but could I ask you to please move the driver part to ```extras``` and add the driver as a reference to the example makefile with a line like this: ```EXTRA_COMPONENTS=extras/bmp180```.

Regards,
Johan

Angus71:
Hi Johan, 

No problem at all ...

I let you know, when I'm finished ... I will move the i2c reference into the extras folder and create a separate github repository for the bmp180 and create also a reference in extras.

Regards,

Frank

projectgus:
Thanks for all this effort @Angus71 & @kanflo. A bmp180 driver will be neat. I have one lying around somewhere on a breakout, so I will fire it up!

Regarding "drivers", keeping them in `/extras/drivername` is definitely the preference. Makes it easy for others to link it into their program without copying any code.

Regarding git submodules, the idea behind having submodules for projects like lwip or axTLS is that these are upstream projects that move at their own pace, so they're going to change independently of esp-open-rtos and we want to keep up with those changes with a minimum of fuss. Hence tracking their git repositories as submodules.

For things that are written specifically for esp-open-rtos, it's a bit different. I don't mind adding them as submodules if you want, but if they're small projects that are closely associated with esp-open-rtos then I'm happy to just include them directly in the esp-open-rtos repository. In the long run I think that's probably less fuss to maintain.

Angus

Angus71:
Hi Angus,

As I'm currently in the process of "migrating" the changes into the repository. I will add at least the bmp180 driver as "source" and not as a git submodule. If it's okay with @kanflo I will also add the I2C driver as source. This also secures that the examples will work, even if the drivers are changed in the futures.
I will also add a note in the example, where to find the current source of the drivers referenced, so that users know where to look for "updated" versions/issues etc.

After all these are only "examples" for users to see what is possible and how others have solved certain problems.

Regards
Frank

projectgus:
Sounds good to me, thanks!

Angus71:
Hi @projectgus, @kanflo,

I've updated the repository as discussed and hope the README.md's I've added are okay with you.

Regards

Frank

projectgus:
This all looks good to me. I have some more comments for the 'extras' readme, but I'll add them after merging.

@kanflo, any comments?

@Angus71, are you able to squash these commits together? The commit history has been getting a bit hairy so I'd like to squash PRs before merging where it makes sense to do so. Never mind if not,  I can squash them from my end and merge that way.

Angus71:
@projectgus, the commits are now squashed ...

kanflo:
@projectgus, @Angus71, no comments, go ahead and merge. I think it's good you added he README.md, I can make the necessary trimmings to it to reflect its newly found home.

projectgus:
I just went to merge this but I have a couple of code style questions I didn't notice the first time around:

* How would you feel about changing LOCAL to static?
* Similarly, would you mind changing the DEBUG macro to something specific like BMP_DEBUG or BMP180_DEBUG?

Angus71:
No problem at all ... Done ...

projectgus:
Thanks!

PR #32, opened 2015-08-26
Remove packed attribute from register structs

foogod:
Per my comment on issue #8, `__attribute__ (( packed ))` can apparently cause GCC to do strange things with accesses under `-O0`, so this removes them (they shouldn't really be necessary anyway).

This also includes a patch to the VAL2FIELD / SET_FIELD macros to remove an AND operation which is probably rarely needed and makes them slightly less efficient than they need to be in general (I added VAL2FIELD_M / SET_FIELD_M which have the old behavior in case anything down the line actually does need it).

### DIFF
--- a/core/include/common_macros.h
+++ b/core/include/common_macros.h
@@ -20,11 +20,20 @@
  * and shift) constants.  Used primarily with ESP8266 register access.
  */
 
-#define VAL2FIELD(fieldname, value) (((value) & fieldname##_M) << fieldname##_S)
+#define VAL2FIELD(fieldname, value) ((value) << fieldname##_S)
 #define FIELD2VAL(fieldname, regbits) (((regbits) >> fieldname##_S) & fieldname##_M)
 
 #define FIELD_MASK(fieldname) (fieldname##_M << fieldname##_S)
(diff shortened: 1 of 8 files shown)


projectgus:
Hi Alex,

Thanks for this. Nice catch with the -O0 access thing, that's really weird. I wish I knew about more arcana about the C standard so I could guess if this was a compiler bug or just some standards-compliant oddness!

I've cherry-picked e2e841c and pushed it.

The other change with the AND, I feel a bit reluctant about. Most of the time the value passed will be a constant, so the AND will be optimised out anyhow. In the rare occasions when it's not, it seems like a wise thing to keep in as a safety check. Do you feel strongly about removing it?

foogod:
Well, it occurred to me that in most cases, if the supplied value doesn't fit, then ANDing out just the lower bits of it are probably going to be the wrong fix anyway, and will just result in unpredictable behavior, which is really the same result you're likely to get without it, so it's not really likely to be a very useful safety check anyway, and just adds overhead.

If we want a real safety check here, we should probably do it with an assert or something, and make it a compile-time option.. That's a larger question/issue, though, I think..

projectgus:
Makes sense. Merged!

PR #33, opened 2015-08-26
Fix broken UART(i) definition

foogod:
Looks like I had a rather grievous cut/paste error in my previous uart_regs.h submission..


projectgus:
Oops, thanks!

PR #34, opened 2015-08-27
Compile by default with -ffunction-sectons -fdata-sections

projectgus:
Shrinks size of 'http_get' example by about 3.5%.

Adding as a PR instead of merging directly because I'm a little suspicious about potential downsides. Looking at disassembly it looks fine, but I have a feeling in some circumstances there may be relaxations the compiler would otherwise have avoided, that now can't be avoided with -ffunction-sections because the target is in a different section... I haven't found that happening much yet, though.

@foogod, @jcmvbkbc, @igrr - do any of you have experience/thoughts regarding this? Do you think it's a good option to use with lx106?

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -94,8 +94,18 @@ OWN_LIBC ?= 1
 # Note: you will need a recent esp
 ENTRY_SYMBOL ?= call_user_start
 
+# Set this to zero if you don't want individual function & data sections
+# (some code may be slightly slower, linking will be slighty slower,
+# but compiled code size will come down a small amount.)
+SPLIT_SECTIONS ?= 1
+
 CFLAGS		= -Wall -Werror -Wl,-EL -nostdlib -mlongcalls -mtext-section-literals -std=gnu99 $(CPPFLAGS)
 LDFLAGS		= -nostdlib -Wl,--no-check-sections -Wl,-L$(BUILD_DIR)sdklib -Wl,-L$(ROOT)lib -u $(ENTRY_SYMBOL) -Wl,-static -Wl,-Map=build/${PROGRAM}.map $(EXTRA_LDFLAGS)
+
+ifeq ($(SPLIT_SECTIONS),1)
+  CFLAGS += -ffunction-sections -fdata-sections
+  LDFLAGS += -Wl,-gc-sections
+endif
 
 ifeq ($(FLAVOR),debug)
     CFLAGS += -g -O0


projectgus:
One interesting thing, this change doesn't generate individual sections for IRAM functions as the attribute sets the section name (whcih doesn't change).

I have a version where the IRAM attribute generates a section name, but iram code size got slightly (~20 bytes) larger so I haven't pushed it.

I _think_ the reason is that the IRAM functions are fairly cherry-picked to only be the things needed by the base system and the NMI handler, so it's hard not to end up with them compiled in.

jcmvbkbc:
> in some circumstances there may be relaxations the compiler would otherwise have avoided, that now can't be avoided with -ffunction-sections because the target is in a different section...

I can only think of the assembler always emitting l32r + callx0 (the compiler will still emit plain call0 despite -mlongcalls or -ffunction-sections) instead of call0 when the call definitely reaches. But these l32r + callx0 will be folded back to call0 by the linker if the call target is reachable.

foogod:
I don't have any real objections to this change per se, but I'm actually not sure I'm in favor of doing it _now_..

The reason is mainly that we're still in the middle (and will probably be getting even moreso) of ripping apart/tweaking/rewriting all of the existing libraries, which is already complicated enough without adding extra compiler/linker magic to the whole mix.  I actually just ended up pulling my hair out for about half a day because of a linker script bug (issue #35) when working on the open-startup-code stuff (and as a result, had occasion to look into the ld scripts we're currently using, and was a bit dismayed at how overcomplicated they seem to be in general, actually)..

It's also somewhat hard to say at this point what the impacts of this change will really be once we rewrite large portions of the way the libraries do things, so it's possible that something like this that seems like a good idea now might actually end up resulting in worse results later on, when nobody's paying attention anymore.

This sort of linking change really seems like a "fine tuning"/"polishing" sort of thing we should be reserving until after we've got most of the code the way we want it, so I'd kinda prefer to just keep it on the "to-do shelf" for a while, personally..

foogod:
On a related note, though, when we do get to the point where we want to do this sort of optimization, it seems to me we should also be compiling with `-Os` by default (not just for size:  With the unusually large cache-miss penalties of IROM code, I suspect that even small improvements in code locality might potentially speed up the execution of some code considerably)..

projectgus:
Thanks @jcmvbkbc for the insight into where gcc adds/removes literals.

I'm going to take foogod's advice and close this for now, though. Can look at it again another day.

Issue #35, opened 2015-08-27
.irom0.literal segment not linked properly


foogod:
The `IROM` macro causes a data definition to be placed in the .irom0.literal segment, however there's nothing anywhere in the linker scripts to cause .irom0.literal (or any other .irom/.irom0 segment that's not part of the sdklibs) to actually be linked into the final binary.

This results in very bad/confusing things if you actually use the `IROM` macro anywhere in your code..

projectgus:
Sorry again about that one. Merged with comment in #36.

PR #36, opened 2015-08-27
Fix linker script handling of irom/irom0 segments

foogod:
(Re: Issue #35)

Segments tagged with .irom or .irom0 should always be linked into the irom region, whether they're part of the sdklib or not..


projectgus:
Sorry about this, it looks like a regression introduced in a566a51c when I removed the objcopy --rename-sections steps from the build process. A good example of where a regression test would have been handy.

Issue #37, opened 2015-08-27
PWM output


gpascualg:
First of all, I do not know if this is the correct place to post this (is there a forum, mailing list, w/e?). Sorry in case it is not.

I've been trying to understand how the GPIO works here, and so far it is quite intuitive. But when it comes to PWM, I have no clue on how to enable it.
I successfully did so on the bare Espressif SDK, but had no luck so far with this RTOS.

Any help would be really appreciated.

Thank you!

projectgus:
It's the correct place for now. There is no forum or mailing list yet, but I'm working on it...

PWM is not something that the ESP hardware does automatically, it's implemented in the Espressif SDK using a timer interrupt and some software.

esp-open-rtos doesn't have this yet. :( Someone needs to write a PWM implementation that uses one of the timer interrupts, and contribute it under `extras/pwm` or similar.

Sorry! I'm going to keep this issue open as a feature request though, because it's something we definitely should have but don't.

blipi:
Oh I see, if that's the case let me take a look at Espressif's implementation and see if I can get something to work.

I can't really test now, as I don't have the ESP with me, but I'll keep you updated in case I do any progress.

Don't worry, it's nice to have such support and effort put in this project ;) I'll try to contribute too.

projectgus:
I'm going to close this issue and move discussion to the PR in #40

PR #40, opened 2015-08-28
[WIP] Basic PWM implementation

gpascualg:
As per #37 I've tried a basic implementation of PWM.

Might be wrong and might not be optimized, so please feel free to comment!

Explanation:

Timer FRC1 is used as the duty timer. An interruption is fired when the GPIO pin should be toggled.
Althought more than one GPIO might be specified, all of them will follow the same period and duty cicle, that's why I'm implementing (not done yet) the "divider".

That would allow a single pin to act slower, for instance, if period is 50us, having a divider of 2 for a specific pin would make it behave as if it was of 100us.

### DIFF
--- a/extras/pwm/pwm.c
+++ b/extras/pwm/pwm.c
@@ -1,0 +1,160 @@
+#include "pwm.h"
+
+#include <espressif/esp_common.h>
+#include <espressif/sdk_private.h>
+#include <FreeRTOS.h>
+#include <esp8266.h>
+
+typedef struct PWMPinDefinition
+{
+    uint8_t pin;
+    uint8_t divider;
+} PWMPin;
+
+typedef enum {
+    PERIOD_ON   = 0,
+    PERIOD_OFF  = 1
+} pwm_step_t;
+
+
+typedef struct pwmInfoDefinition
+{
+    uint8_t running;
+
+    uint16_t freq;
+    uint16_t dutyCicle;
+
+    /* private */
+    uint32_t _maxLoad;
+    uint32_t _onLoad;
+    uint32_t _offLoad;
+    pwm_step_t _step;
+
+    uint16_t usedPins;
+    PWMPin pins[8];
+} PWMInfo;
+
+static PWMInfo pwmInfo;
+
+void frc1_interrupt_handler(void)
+{
+    uint8_t i = 0;
+    bool out = true;
+    uint32_t load = pwmInfo._onLoad;
+    pwm_step_t step = PERIOD_ON;
+
+    if (pwmInfo._step == PERIOD_ON)
+    {
+        out = false;
+        load = pwmInfo._offLoad;
+        step = PERIOD_OFF;
+    }
+
+    for (; i < pwmInfo.usedPins; ++i)
+    {
+        gpio_write(pwmInfo.pins[i].pin, out);
+    }
+
+    timer_set_load(FRC1, load);
+    pwmInfo._step = step;
+}
+
+void pwm_init(uint8_t npins, uint8_t* pins)
+{
+    /* Assert number of pins is correct */
+    if (npins > MAX_PWM_PINS)
+    {
+        printf("Incorrect number of PWM pins (%d)\n", npins);
+        return;
+    }
+
+    /* Initialize */
+    pwmInfo._maxLoad = 0;
+    pwmInfo._onLoad = 0;
+    pwmInfo._offLoad = 0;
+    pwmInfo._step = PERIOD_ON;
+
+    /* Save pins information */
+    pwmInfo.usedPins = npins;
+
+    uint8_t i = 0;
+    for (; i < npins; ++i)
+    {
+        pwmInfo.pins[i].pin = pins[i];
+
+        /* configure GPIOs */
+        gpio_enable(pins[i], GPIO_OUTPUT);
+    }
+
+    /* Stop timers and mask interrupts */
+    pwm_stop();
+
+    /* set up ISRs */
+    _xt_isr_attach(INUM_TIMER_FRC1, frc1_interrupt_handler);
+
+    /* Flag not running */
+    pwmInfo.running = 0;
+}
+
+void pwm_set_freq(uint16_t freq)
+{
+    pwmInfo.freq = freq;
+
+    /* Stop now to avoid load being used */
+    if (pwmInfo.running)
+    {
+        pwm_stop();
+        pwmInfo.running = 1;
+    }
+
+    timer_set_frequency(FRC1, freq);
+
+    if (pwmInfo.running)
+    {
+        pwm_start();
+    }
+}
+
+void pwm_set_duty(uint16_t duty)
+{
+    pwmInfo.dutyCicle = duty;
+    pwm_restart();
+}
+
+void pwm_restart()
+{
+    if (pwmInfo.running)
+    {
+        pwm_stop();
+        pwm_start();
+    }
+}
+
+void pwm_start()
+{
+    pwmInfo._maxLoad = timer_get_load(FRC1);
+    pwmInfo._onLoad = pwmInfo.dutyCicle * pwmInfo._maxLoad / UINT16_MAX;
+    pwmInfo._offLoad = pwmInfo._maxLoad - pwmInfo._onLoad;
+    pwmInfo._step = PERIOD_ON;
+
+    // Trigger ON
+    uint8_t i = 0;
+    for (; i < pwmInfo.usedPins; ++i)
+    {
+        gpio_write(pwmInfo.pins[i].pin, true);
+    }
+
+    timer_set_load(FRC1, pwmInfo._onLoad);
+    timer_set_reload(FRC1, false);
+    timer_set_interrupts(FRC1, true);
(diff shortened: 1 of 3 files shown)


extras/pwm/pwm.c
```diff
@@ -0,0 +1,160 @@
...
+    uint32_t _offLoad;
+    pwm_step_t _step;
+
+    uint16_t usedPins;
+    PWMPin pins[8];
+} PWMInfo;
+
+static PWMInfo pwmInfo;
+
+void frc1_interrupt_handler(void)
```
jcard0na: I would make this static to restrict its scope to this module.

jcard0na:
Hi Guillem,

This is a great contribution.  I wanted to help by testing it on my ESP8266 Huzzah but could not get it to work.  The code is pasted below.

```c
void task1(void *pvParameters)
{
    printf("Hello from task1!\r\n");
    uint32_t count = 0;
    while(1) {
        vTaskDelay(1000);
        count++;
        if (count % 2) {
            printf("pwm_stop()\n");
            pwm_stop();
        } else {
            printf("pwm_start()\n");
            pwm_start();
        }
    }
}

void user_init(void)
{
    uint8_t pins[1];
    uart_set_baud(0, 115200);

    printf("SDK version:%s\n", sdk_system_get_sdk_version());

    printf("pwm_init(1, [14])\n");
    pins[0] = 14;
    pwm_init(1, pins);
    printf("pwm_set_freq(1000)     # 1 kHz\n");
    pwm_set_freq(1000);
    printf("pwm_set_duty(UINT16_MAX/2)    # 50%%\n");
    pwm_set_duty(UINT16_MAX/2);

    xTaskCreate(task1, (signed char *)"tsk1", 256, NULL, 2, NULL);
}
```
This codes freezes when calling ```pwm_set_freq(1000);```.  After commenting that out, the code runs and the task toggles pwm on and off.  But this eventually triggers a reset (watchdog timer?).  See the output below.

```
ESP-Open-SDK ver: 0.0.1 compiled @ Nov 25 2015 00:22:37
phy ver: 273, pp ver: 8.3

SDK version:0.9.9
pwm_init(1, [14])
// pwm_set_freq(1000)     # 1 kHz
pwm_set_duty(UINT16_MAX/2)    # 50%
mode : sta(18:fe:34:fd:93:53)
add if0
Hello from task1!
scandone
add 0
aid 2
cnt 

connected with cookabit, channel 11
dhcp client start...
ip:192.168.1.127,mask:255.255.255.0,gw:192.168.1.1
pwm_stop()
pwm_start()

 ets Jan  8 2013,rst cause:2, boot mode:(3,6)
```

After the pwm_start() and before the reset, I do see gpio 14 cycling at 50% duty cycle.  With a scope I measured the frequency to be 152 kHz, which I am assuming it's de default frequency you would get when pwm_set_frequency() is not called.

Sorry I could not get to the bottom of this, but I thought my tests results might be useful.

Cheers,

extras/pwm/pwm.c
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    /* Stop timers and mask interrupts */
+    pwm_stop();
+
+    /* set up ISRs */
+    _xt_isr_attach(INUM_TIMER_FRC1, frc1_interrupt_handler);
+
+    /* Flag not running */
+    pwmInfo.running = 0;
```
jcard0na: This is redundant as the preceding pwm_stop() will already clear this flag.

jcard0na:
More progress on this...

Turns out that just by replacing the call to ```pwm_set_frequency(1000)``` by ```timer_set_frequency(FRC1, 1000)``` my example code works, no resets are ever triggered.  Except that the frequency value is incorrect: it is around 150 kHz.

After that I tried to lower the frequency to 10Hz.  With the logic analyzer I observed that the first pwm_start() produces the correct frequency (10 Hz) , but the frequency is doubled after each restart (ie. to 20 Hz, 40 Hz and so on). 



This seems to be caused by an error in calculating the timer load values.  Changing the duty cycle changes the rate at which frequency increases.  With a duty cycle of 25%, the frequencies grow as (10 kHz, 13.33 kHz, 17.77 kHz, 23.7 kHz



So I would say there are two problems with the code.  

1. The mysterious freeze when calling pwm_set_frequency().  I ruled out stack overflow: a call to xPortGetFreeHeapSize() returns 41500 bytes of available stack/heap.  I'm out of ideas on this.

2. The frequency increase after each pwm_start(), pwm_stop().  This seems to be an easy fix.

Cheers,

Javier

jcard0na:
OK, I think I figured it out.  

### Issue 1

Regarding the freeze, it turns out that ```timer_set_frequency(FRC1, 1000)``` ends up calling one of two functions depending on whether the compiler can tell if the arguments are constant at compile time or not.  If gcc determines they are constant, the function works.  If gcc cannot tell, the function freezes in this infinite loop:

```c
bool _timer_set_frequency_runtime(const timer_frc_t frc, uint32_t freq)
{   
    return _timer_set_frequency_runtime(frc, freq);
}
```

When my code invoked ```timer_set_frequency()``` directly, I would get the "good" function.  When my code would call ```timer_set_frequency()``` via ```pwm_set_frequency()```, the "shot-yourself-on-the-foot-over-and-over" version of the function would get called instead.

I do not really why such function exists.  Maybe the author of that module, @projectgus, can shed some light and suggest a clean approach to implement ```pwm_set_frequency()```?

### Issue 2

This is resolved by saving the load value only after the frequency is set:

```diff
--- pwm.orig.c	2015-11-25 16:57:45.282585549 -0800
+++ pwm.c	2015-11-25 16:57:13.666586940 -0800
@@ -108,6 +108,7 @@
     }
 
     timer_set_frequency(FRC1, 10);
+    pwmInfo._maxLoad = timer_get_load(FRC1);
 
     if (pwmInfo.running)
     {
@@ -132,7 +133,6 @@
 
 void pwm_start()
 {
-    pwmInfo._maxLoad = timer_get_load(FRC1);
     pwmInfo._onLoad = pwmInfo.dutyCicle * pwmInfo._maxLoad / UINT16_MAX;
     pwmInfo._offLoad = pwmInfo._maxLoad - pwmInfo._onLoad;
     pwmInfo._step = PERIOD_ON;
```

gpascualg:
Hey! Thank you for the input.

I can not do it now but tomorrow I will be implementing your suggestions.

Thank you fir the fix too. At the time of writing that I had no ESP module to test it into. Now that I have it, I have no time to wire it all up... :(

Anyway, tomorrow I'll do some commits and squash, thank you so much for such a detailed debugging!

yesco:
can this code be added as an example?

one if the biggest issues with e-o-r is we don't know what is tested and
verified to work...

thanks

jcard0na:
@gpascualg thanks for following up.  It's a holiday (Thanksgiving) on my side of the world, so I won't be doing much either in the next few days.  But I might be able to test your submissions.  
Also, I'd love to know what is the maximum frequency you get out of your esp module.  I could not go above ~40kHz.

@yesco sure, i can add test code if @gpascualg does not include it in his updated pull request.  I have to say, though, that esp-open-rtos is above all other ESP sdk's regarding sample code.  I am using it in my project because I could very run a bunch of examples covering many features.  I don't know who is responsible, but kudos to them!

Cheers,

gpascualg:
It's seems like I've deleted my fork :-1: I'll have to fork again and create another Pull Request. As soon as I get home I will do it, sorry for any inconvenience.

As for tests, I had none on mind. If you have any ideas I don't mind including them in further commits.

projectgus:
Oh no! You can get your fork back and re-synced with this pull request though, all the commits are stored against the pull request in the esp-open-rtos repo as well. This should work to start from scratch:

* First re-fork in github, then run these commands:

> git clone git@github.com:gpascualg/esp-open-rtos.git
> cd esp-open-rtos
> git remote add upstream git@github.com:SuperHouse/esp-open-rtos.git
> git fetch upstream pull/40/head
> git reset --hard FETCH_HEAD
> git push --force origin master

Those steps assume you don't have a local copy of your fork. If you do then you can probably just do the last step (git push --force) to push to your recreated fork and update your branch to the same as the PR.

I think github will automatically re-associate your master branch with this pull request once again, it sees it reappear (although I could be wrong about that.)

BTW, great to see all this discussion about the PWM implementation everyone. I haven't had time to follow closely but please @-mention me if you have any specific queries I might be able to help with.

gpascualg:
@projectgus Good to know! I didn't know I was able to "recover" a lost pull request. I'll do it now and update.

Don't worry about that, thought it would be nice if your could lighten up issue 1 on: https://github.com/SuperHouse/esp-open-rtos/pull/40#issuecomment-159770193

gpascualg:
Nope, it doesn't pick my new fork again. I've tried both using master branch and another one, `pwm`, but no luck... I leave it up to you whether to create a new PR or what to do.

Sorry again!

projectgus:
Drat. :( I guess starting a new PR (and reference this one in the comment) is the way to go, then.

gpascualg:
Please see https://github.com/SuperHouse/esp-open-rtos/pull/72 for follow up. I'm closing this due to problems with my fork.

PR #46, opened 2015-09-13
Feature/dhcp server

projectgus:
DHCP server feature, as discussed on mailing list:
https://groups.google.com/forum/#!topic/esp-open-rtos/4BlYYrYRfH8

I would be much happier finding an embeddable third party DHCP server that we can make use of instead of writing yet another half baked one, so if anyone knows of one please let me know and I'll move to it.

Otherwise... enjoy this half baked one!

Please report back if you successfully use this on different platforms. I've personally tested the access_point example only on my Debian 8 laptop and my Android 5 phone. Both worked, although the phone disconnects after ~10 seconds after it can't find a route to the internet.

If a few people can confirm this works for them across different platforms then that'll be a great start, and I'll merge it into master.

(Other comments/critique/code review also welcome. I've probably violated the RFC in at least a couple of places...)


projectgus:
Does anyone have any comments on this before I merge it?

It probably needs more testing, but the way to do this may be to merge it first so people see that it exists. I can target it early-on in the automated test writing spree that I plan to go on soon.

foogod:
I have no objections.. I never really did get around to giving this a try, but I don't really see a reason not to merge what you have so far, even if it's not fully tested..  We can always keep refining it..

projectgus:
thanks @foogod .

The main reason I hadn't merged it is that I'd been hoping to find someone else's more mature code that we could use instead, because writing your own crippled DHCP server seemed like a "what not do". But there doesn't seem to be one, so here we are!

PR #49, opened 2015-09-20
Open startup

### LINKED ISSUE #2: Open source startup code

projectgus:
I'd like to look at merging this, it seems pretty stable in the testing I've done. Closes #2.

Some discussion about next stages here:
https://groups.google.com/forum/#!topic/esp-open-rtos/D7SPgMrqc9E


foogod:
Sorry, I promised to do this PR a few days back and then completely neglected to do it..

I've updated my branch with the new newlib stuff and made the necessary changes to the printf format strings so it now builds correctly against the current master, so I think it should be OK to merge at this point.

projectgus:
Looks great. I did a few more tests with client and AP modes, and the ota_basic example, everything seemed perfect.

Issue #52, opened 2015-10-05
Reading from UART is broken


projectgus:
As per reports on mailing list and in comments of #31.

The current UART reading code in master doesn't seem to work.

Fix will be to implement our own UART wrapper functions and test whatever initialisation stages are necessary to set the UART RX FIFO up correctly and read/buffer characters in hardware on input.

kanflo:
I have started work on using the HW FIFO in #31 and guess that can be reused here. Just in case we don't want to duplicate our efforts.

projectgus:
Hi kanflo,

That'd be great, thanks! Are there more changes you've got in progress right now?

Angus

foogod:
I haven't had a chance to look at any of the existing code, but I'm not sure what you mean about initialization of the UART FIFO.. there should be no initialization required except setting up the baud rate divisor and whether it uses RTS/CTS handshaking.  the UART automatically uses the hardware FIFO (and I don't think there's even a way to turn it off), so there should be no special configuration needed..

I've actually successfully used the UART with no special initialization even with bare code (no SDK, just booting straight from the ROM into my own entry point), so I know this to be true..

Are you sure you aren't overthinking this whole thing?  I suspect the problem may be much simpler (like just not reading the FIFO status right or something)

projectgus:
I could easily be overthinking it and doing something silly.

Just to be clear though, with your experience are you talking about UART RX or TX? I have no problems with UART TX (ie putc/printf), but working with UART RX seems trickier somehow...

foogod:
(non-interrupt-driven) RX and TX are almost exactly the same, really.. in the case of TX, you check the TX FIFO status to make sure it's not full (number returned is less than 127), block if necessary/appropriate until it's ok, then write the byte to the FIFO register.  RX, you check the RX FIFO status to make sure there's something in it (number returned is not 0), block if necessary/appropriate, then read a character from the FIFO register.

That's really all there should be to it..

I can take a look at the current state of things in more detail and try to figure out what's going on when I get home this evening, if you want..

projectgus:
[...]

PR #53, opened 2015-10-06
Fix issue #52.  Add esp/uart.h

foogod:


projectgus:
Looks great!

* One thing I noticed is that the behaviour of the SDK to sneak in a CR ahead of any LF printed to the output is changed here. I think that's fine, it's weird to implicitly mess with stdout, but just wanted to acknowledge it's changing.

* I snuck in a commit with another UART-related improvement I wanted to make here, changing the default baud rate to 115200 bytes (and providing non-SDK baud rate functions). I think we discussed this before(?) but I think it's a good improvement, most serial adapters don't support 74906 at all.

* I want to quickly check a couple of other behavioural details before merging this, at least to make sure what I was seeing last night was me making mistakes in my uart.h. And also I want to check if any locks are held when _read_r() is called.

projectgus:
> And also I want to check if any locks are held when _read_r() is called.

No. The file descriptor is locked, but the system-level locking mechanism is only used to lock access to the file descriptor's lock. So it's fine the way it is.

projectgus:
> I want to quickly check a couple of other behavioural details before merging this, at least to make sure what I was seeing last night was me making mistakes in my uart.h.

Yeah, that must have been all me - stuff that I was sure wasn't working last night is working fine now.

+1 to merge from my end, although you probably want to review the other commits I added.

foogod:
> I snuck in a commit with another UART-related improvement I wanted to make here, changing the default baud rate to 115200 bytes [...]. I think we discussed this before(?)

Yeah, and I was opposed to it at the time..

> but I think it's a good improvement, most serial adapters don't support 74906 at all.

I think that's actually completely wrong.  Pretty much every USB serial adapter I've seen (which is what almost everybody is using these days) handles oddball rates like this just fine.  Some older serial apps/drivers don't like it, but if you're hardware hacking on something like an ESP8266 you're probably savvy enough to sort out those sorts of software issues.

The main reason I'm opposed to this is that with this change it's now pretty much impossible for anybody to actually get full boot information out of the thing in a single pass, because now (without anybody telling it to or having any way to prevent it from happening if they don't want it to) it switches baud rate part way through the boot sequence in a way that's basically guaranteed to make sure either the first part or the second part of the boot comes across as garbage.  At least prior to this change, if you really wanted to see everything that was going on during boot you could set your serial rate to 74906 and leave the divisor alone in your app and everything would work nice and cleanly.  This now irreversibly breaks that option.

Given that in my opinion the only thing wrong with the 74906 baud rate is a personal aesthetic issue, I think practicality and not removing functionality should be more important than stylistic concerns here.

projectgus:
> Yeah, and I was opposed to it at the time..

Sorry, I forgot the discussion. Change reverted.

> The main reason I'm opposed to this is that with this change it's now pretty much impossible for anybody to actually get full boot information out of the thing in a single pass,

I have a proposal that may allow for fixing this in another way, but I'll leave explaining it for another time.

> I think that's actually completely wrong. Pretty much every USB serial adapter I've seen (which is what almost everybody is using these days) handles oddball rates like this just fine.

I must have bad luck. I have CP2102 (NodeMCU v1.0), some cheap "FT232R"s, and a Freetronics USBSerial (atmega16u2 based), and none of them seem to be able to set ~74900 under Linux. The only adapter I have on hand that works for me at 74900 is a CH341 (NodeMCU v0.9) and it has other problems (kernel driver baud rate setting bugs, and a high bit error rate when flashing at high speed).

I think genuine FTDIs should be able to do it, but the only ones of these I have are integrated into other devices.

>Actually, this also brings up a larger issue, which I think I'll open a separate ticket for: We should also have a way for people to override the default stdin/stdout if they want to (for example, redirect stdout to UART(1) instead, or to some completely non-UART thing).

This is a good feature to raise an issue for. I'm happy to bring back `\r\n` for now, with a goal of making it a configurable choice (via swapping in a different uart "driver") in the future.

foogod:
Are we good to merge this?  I think I'm fairly happy with it at this point if you are..

foogod:
Oh, except I forgot we still haven't resolved the baud rate issue..  Can we revert that bit and merge the rest of it to actually fix the (more serious) stdin bug, and then continue the debate separately from this PR?

Regarding device support, I've never had any problems with the CP210x devices at 74906 (that's what I use most of the time).  That makes me wonder if you've got old drivers or something.  Likewise my CH341 is fine with it (though I agree that the CH341 is crap in general and will probably never buy another one).  I haven't tried it specifically with this, but I also know from past experience that the FTDI chips will do pretty much any baud rate (though I personally don't patronize FTDI any more as a matter of principle).  I have to admit I don't know anything about the Freetronics one..

To be honest, I'm not sure why Espressif has that code in there to set the divisor to begin with, and don't see why it doesn't just leave it the way the boot ROM has already set it (I'd be fine just taking that bit out of app_main.c entirely, really)..  Regarding the larger issue, my general philosophy on things like this, though, is that we should give the app developer as much control over the platform behavior as we can.  With the current method, if you don't want the baud rate changed, you can leave it at 74906, and if you do, you can explicitly change it.  If we change it to 115200 that early in the SDK startup process, then if the app designer doesn't want that behavior, they have no control or choice over it, which in by book counts as removing choice/functionality that was previously present..

projectgus:
The baud rate change was reverted already in 7823a28. Can continue discussing that on the mailing list.

I pushed two more commits, to do with disabling stdout line buffering as this was also discussed in #53. I'm +1 for merging this if it all looks OK to you.

foogod:
Ah, ok, sorry.. I apparently missed that commit, and I saw that the lines were still being changed in the cumulative diff output but apparently didn't look closely enough to see that the number was now 74906 instead of 115200..

I'll go ahead and merge, then..

PR #59, opened 2015-10-26
IRQ driven UART0 driver using UART HW FIFO

kanflo:
Continuing here from PR #31.


kanflo:
Changed examples/terminal to do something useful as interactive GPIO flipping.

extras/serial-driver/serial_driver.h
```diff
@@ -0,0 +1,33 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+#ifndef __I2C_H__
```
projectgus: Minor thing, but this is copy-pasta :)

projectgus:
Thanks for re-submitting this.

I _really_ like the new terminal example! That's a very nifty example, thank you. It's great to have examples that demonstrate non-trivial things, a lot of the stuff we have is pretty "hello world" at this stage.

I'm not sure about the name `serial-driver` for the component. I think a more suitable name would be something like `stdin_uart_interrupt` or maybe some variation on that? As that's what it really is, it maintains the same basic stdin UART functionality but makes it interrupt driven.

(Also the underscores-not-hyphens thing seems to be a semi-convention for directory names, except for `extras/rboot-ota` which needs changing to match the others.)

To make sure I understand - the terminal program would still work if you removed the `extras/serial-driver` component and used the built-in stdin uart reader - yes? But I can see that in some circumstances you may want a high priority serial task, that responds immediately when serial data is available but doesn't starve other tasks when serial data isn't available. So this driver is very useful for that. :)

The other thing I'd suggest adding is just a README in the `extras/serial-driver` directory explaining what the driver is for, why you'd use it, how to enable it in the Makefile, and to look at the 'terminal' example for an example that uses it.

I added a couple of other code comments inline above.

@foogod , do you have any other review comments for this PR?

Angus

extras/serial-driver/serial_driver.h
kanflo: Oups!

kanflo:
@projectgus, good comment. I did explain the idea behind the driver in a PR thread and the same explanation should go into the readme. The name change you suggested sounds good. And yes, the terminal program would work without this driver.

kanflo:
I renamed the module to stdin_uart_interrupt and added a README and also did some squashing. The August commit (9d80dc4) is still tagging along even though all the SET_PERI stuff and SW ring buffer have been superseded. I feel a bit anxious about going back that far to rewrite the git history.

projectgus:
Thanks Johan, this looks pretty good. I want to test it on some real hardware as well, but that might not happen until after next week. Sorry! (The curse of doing projects in your spare time.)

kanflo:
No worries, looking forward to it.

projectgus:
Rebased and merged as a single commit 4cfe40d34885, with minor cleanup and a bugfix as mentioned above.

Sorry for the delay in testing this, thanks again for contributing it! :)

kanflo:
Great!

PR #72, opened 2015-11-27
[WIP] Basic PWM implementation

gpascualg:
*Please see https://github.com/SuperHouse/esp-open-rtos/pull/40 for previous talk.*

As per #37 I've tried a basic implementation of PWM.

Might be wrong and might not be optimized, so please feel free to comment!

Explanation:

Timer FRC1 is used as the duty timer. An interruption is fired when the GPIO pin should be toggled.
Althought more than one GPIO might be specified, all of them will follow the same period and duty cicle, that's why I'm implementing (not done yet) the "divider".

That would allow a single pin to act slower, for instance, if period is 50us, having a divider of 2 for a specific pin would make it behave as if it was of 100us.

### DIFF
--- a/extras/pwm/component.mk
+++ b/extras/pwm/component.mk
@@ -1,0 +1,9 @@
+# Component makefile for extras/pwm
+
+INC_DIRS += $(ROOT)extras/pwm
+
+# args for passing into compile rule generation
+extras/pwm_INC_DIR =  $(ROOT)extras/pwm
+extras/pwm_SRC_DIR =  $(ROOT)extras/pwm
+
+$(eval $(call component_compile_rules,extras/pwm))
--- a/extras/pwm/pwm.c
+++ b/extras/pwm/pwm.c
@@ -1,0 +1,160 @@
+#include "pwm.h"
+
+#include <espressif/esp_common.h>
+#include <espressif/sdk_private.h>
+#include <FreeRTOS.h>
+#include <esp8266.h>
+
+typedef struct PWMPinDefinition
+{
+    uint8_t pin;
+    uint8_t divider;
+} PWMPin;
+
+typedef enum {
+    PERIOD_ON   = 0,
+    PERIOD_OFF  = 1
+} pwm_step_t;
+
+
+typedef struct pwmInfoDefinition
+{
+    uint8_t running;
+
+    uint16_t freq;
+    uint16_t dutyCicle;
+
+    /* private */
+    uint32_t _maxLoad;
+    uint32_t _onLoad;
+    uint32_t _offLoad;
+    pwm_step_t _step;
+
+    uint16_t usedPins;
+    PWMPin pins[8];
+} PWMInfo;
+
+static PWMInfo pwmInfo;
+
+void frc1_interrupt_handler(void)
+{
+    uint8_t i = 0;
+    bool out = true;
+    uint32_t load = pwmInfo._onLoad;
+    pwm_step_t step = PERIOD_ON;
+
+    if (pwmInfo._step == PERIOD_ON)
+    {
+        out = false;
+        load = pwmInfo._offLoad;
+        step = PERIOD_OFF;
+    }
+
+    for (; i < pwmInfo.usedPins; ++i)
+    {
+        gpio_write(pwmInfo.pins[i].pin, out);
+    }
+
+    timer_set_load(FRC1, load);
+    pwmInfo._step = step;
+}
+
+void pwm_init(uint8_t npins, uint8_t* pins)
+{
+    /* Assert number of pins is correct */
+    if (npins > MAX_PWM_PINS)
+    {
+        printf("Incorrect number of PWM pins (%d)\n", npins);
+        return;
+    }
+
+    /* Initialize */
+    pwmInfo._maxLoad = 0;
+    pwmInfo._onLoad = 0;
+    pwmInfo._offLoad = 0;
+    pwmInfo._step = PERIOD_ON;
+
+    /* Save pins information */
+    pwmInfo.usedPins = npins;
+
+    uint8_t i = 0;
+    for (; i < npins; ++i)
+    {
+        pwmInfo.pins[i].pin = pins[i];
+
+        /* configure GPIOs */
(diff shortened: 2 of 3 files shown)


jcard0na:
Hi Guillem,

Thanks for incorporating my suggestions.  As for the pending open issue, I thought I'd copy it over from #40 so it does not get lost, as I don't think your implementation will work until it is resolved.

### Issue 1

Regarding the freeze, it turns out that ```timer_set_frequency(FRC1, 1000)``` ends up calling one of two functions depending on whether the compiler can tell if the arguments are constant at compile time or not.  If gcc determines they are constant, the function works.  If gcc cannot tell, the function freezes in this infinite loop:

```c
bool _timer_set_frequency_runtime(const timer_frc_t frc, uint32_t freq)
{   
    return _timer_set_frequency_runtime(frc, freq);
}
```

When my code invoked ```timer_set_frequency()``` directly, I would get the "good" function.  When my code would call ```timer_set_frequency()``` via ```pwm_set_frequency()```, the "shot-yourself-on-the-foot-over-and-over" version of the function would get called instead.

I do not really why such function exists.  Maybe the author of that module, @projectgus, can shed some light and suggest a clean approach to implement ```pwm_set_frequency()```?

projectgus:
> I do not really why such function exists. Maybe the author of that module, @projectgus, can shed some light and suggest a clean approach to implement pwm_set_frequency()?

This is called "be too clever for your own good, then shoot yourself in the foot by not testing every possibility."

I've just pushed a commit that removes that "decide to inline or not at compile time" pattern entirely from timers.h. Please let me know if any of the timer functionality still doesn't seem to work.

projectgus:
(To be clear, I'm the one too clever for their own good here!)

jcard0na:
Thanks Angus.  I just looked at the patch and I believe it will resolve the issue.  I'll try to test it later today.

Thanks!

jcard0na:
Testing looks good.

You can see here my test report: http://jcard0na.github.io/Testing-PWM-implementation-for-esp8266/
For convenience I copy here the list of issues:

1. The library crashes when the duty cycle is set to 0 (pwm_set_duty(0)). It is probably a problem with the underlying timers, but I would suggest detecting that in the pwm_set_duty() function and just call pwm_stop().

2. When setting the duty cycle to 100%, the output is not truly 100% (i.e. constant high) but rather 99.65% I would also suggest to detect that scenario in pwm_set_duty() and set the output to constant high.

3. The library/target does not work at high frequencies. When setting the frequency to 50kHz the observed output is 40kHz.

4. Every change of duty cycle resets the output to HIGH. This might introduce glitches in applications where the duty cycle changes frequently (e.g. audio)

I also made a pull request with my example/test code.

Cheers,

jcard0na:
Hi @projectgus , is there anything preventing you from merging this (and the follow up #73)?
IMHO, none of the issues I reported are a blocker.  The implementation provided by @gpascualg is quite usable.

Best,

projectgus:
Hi @jcard0na,

Sorry for the slow reply, I had a busy few weeks. This generally looks great, there's just a couple of things I'd like to see before merging:

- A fix for issue (1) would be great, so the API call doesn't cause a crash.
- There's no Copyright/LICENSE information on the source files. You can just copy the short blurb from any of the other source files for the header, [ie here](https://github.com/SuperHouse/esp-open-rtos/blob/master/core/esp_gpio.c#L1). Feel free to change the copyight to yourself, of course!

If you're interesting in doing some of this yourself @jcard0na, please go ahead. You can actually check this pull request and apply changes to do it, then submit a new pull request. (This also allows you to merge #73 into this branch so they can go in as a single PR.)

The syntax for checking out a Pull Request locally looks like this:
```
git fetch origin pull/72/head:pwm_impl
```

(Where 'origin' is the name of the remote, 72 is the number of the PR, and 'pwm_impl' at the end is the name of the branch that you want to check the PR out as.)

projectgus:
(Sorry, I got a bit mixed-up in the previously comment - seems copyright belongs to @gpascualg as original author not @jcard0na, so any copyright coment should reflect that.)

jcard0na:
Hi @projectgus,

I'd be happy to help out and modify the pull request as suggested, assuming that @gpascualg is OK with me adding his copyright notice to his contribution.  

BTW, do I need any specific permissions to be able to checkout the pull request?

I'm able to pull master, but when I try to fetch a pull request, I get the following error:

```
~/dev/esp-open-rtos$ git fetch origin pull/72/head:pwm_impl
fatal: Couldn't find remote ref pull/72/head
```

I am assuming origin should be the *target repository* for this pull request, right?  I also tried with the source repo, just for kicks, with the same result.

```
~/dev/esp-open-rtos$ git remote -v
gpascualg	git@github.com:gpascualg/esp-open-rtos.git (fetch)
gpascualg	git@github.com:gpascualg/esp-open-rtos.git (push)
origin	git@github.com:jcard0na/esp-open-rtos.git (fetch)
origin	git@github.com:jcard0na/esp-open-rtos.git (push)
```

Cheers,

gpascualg:
Hey, sry @jcard0na and @projectgus. I've been really busy lately and had no time no update it.

It is still a WIP, I wanted to implement that "divider" thing to multiplex the number of IO pins used as PWM at different "simulated" frequencies.

I'd be glad if anyone can continue it, and worry not about copyrights. This is a BSD licensed project and so was the intention on this PR.

@jcard0na Unless you've set up ssh keys, try using the https version (https://github.com/SuperHouse/esp-open-rtos.git)

jcard0na:
Thanks @gpascualg.  I must have had something misconfigured in my local clone.  I did a clean git clone and now I'm able to fetch pull requests.

Would you object at merging what you have submitted so far (including the minor changes I suggested for [issues 1 and 2 here](https://github.com/SuperHouse/esp-open-rtos/pull/72#issuecomment-160373123)?  

I will not have time to implement the divider you mention, and it seems as you might not have time for it right now.  But the code as-is would be a significant improvement over what's there now.

projectgus:
Closing as I think this was covered by #77 (let me know if I'm misreading.)

PR #73, opened 2015-11-29
Example code for PWM library

jcard0na:
This is example code for [pull request 72](https://github.com/SuperHouse/esp-open-rtos/pull/72) from @gpascualg.

@projectgus: Please only merge after PR72.


jcard0na:
As mentioned above, this PR depends on PR72, so it is expected that the continuous integration tests fail.

PR #77, opened 2015-12-11
Initial PWM implementation from gpascualg

jcard0na:
This pull request merges #72 and #73 plus the following changes:

1. Handle duty cycle 0% so it does not crash
2. Handle duty cycle 100% so output remains constant on
3. Add license notices
4. Modify test code to cover 0% and 100% duty cycle.

Compiled and tested on adafruit's Huzza esp8266 board.  Requires an LED or an oscilloscope to verify that the test code works as intended.

### DIFF
--- a/examples/pwm_test/Makefile
+++ b/examples/pwm_test/Makefile
@@ -1,0 +1,4 @@
+# Simple makefile for simple example
+PROGRAM=pwm_test
+COMPONENTS = FreeRTOS lwip core extras/pwm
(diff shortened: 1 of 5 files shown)


projectgus:
Sorry for the delay in replying, this looks good. Eventually it would be good to expand the software PWM capabilities, but this is a useful initial implementation.

gpascualg:
Glad to see it merged! Thank you @jcard0na for continuing my job!

Issue #55, opened 2015-10-09
Add ability to install custom stdin/stdout handlers


foogod:
The Espressif SDK includes the option for an app to install a custom "putc" function if they want stdout to go somewhere other than UART0.  Currently with the newlib stuff, etc, esp-open-rtos does not actually support this, but it would be a good thing to bring back (and shouldn't be too hard).

(As a side-note, it might also be interesting to add support for registering a stderr handler as well)

projectgus:
I linked #61 to here as supporting POSIX read, write, close, fcntl transparently on both LWIP & UARTS (and someday probably filesystem operations) overlaps with this work as well.

Something that was added to mainline LWIP (much newer than our revision) this year is the ability to add an offset to the fds that LWIP allocates, which makes it pretty straightforward to identify whether a file descriptor belongs to newlib or LWIP:
http://git.savannah.gnu.org/cgit/lwip.git/commit/?id=8155b8cfb31e877c5c51cc539fa121347193f6ce
https://savannah.nongnu.org/patch/?7702

kanflo:
This is a feature I need and if no-one else has started looking into this, I would like to. Regarding the LWIP ```LWIP_SOCKET_OFFSET ```offset, could we apply that patch you liked to @projectgus to the existing EOR LWIP repository (for better or for worse) or is that too hawkish?

Regarding an optional replacement of ``putc``, it would be easy to install a custom handler as you mention @foogod. I think it would be useful to have the option of several (at least two) handlers for both local UART debug output as well as (say) UDP based output for remote debugging.

kanflo:
Also, when pulling the LWIP patch, we need to decide on ranges for the various file descriptors part from stdin/out/err: LWIP, SPIFFS, ...

PR #83, opened 2016-01-19
Added the ability to select stdout UART by modyfying STDOUT_UART.

doragasu:
I have modified the sources to be able to select UART1 TX (GPIO2) instead of UART0 to output messages using printf/os_printf. This way I can leave UART0 free for other tasks (like talking to a microcontroller using a binary protocol).

Not the most elegant patch you have seen for sure (involves modifying common.mk to pass a define to the compiler), but as syscalls must be modified for this to work, I couldn't think of a better way.

Feel free to merge the code if you find this useful (for me it is essential!).

And thanks for the great project, keep up the FOSS good work!

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -48,6 +48,9 @@ ESPTOOL ?= esptool.py
 ESPPORT ?= /dev/ttyUSB0
 ESPBAUD ?= 115200
 
+# Set STDOUT_UART to the UART number that will be used to output using printf/os_printf
+STDOUT_UART ?= 0
+
 # Set OTA to 1 to build an image that supports rBoot OTA bootloader
 #
 # Currently only works with 16mbit or more flash sizes, with 8mbit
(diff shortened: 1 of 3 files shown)


projectgus:
Hey,

Sorry for the slow reply and thanks for sending this PR!

I'm very keen to support reconfiguring stdout to be more flexible than just UART0. #55 covers the long-term plans for this.

Because we want to (soon) do this in a fully generalised way, I'm reluctant to merge this PR - as it adds additional overhead to roll back, and I really don't want to complicate common.mk any more than it already is.

However, there should be no reason that you can't keep using UART1 in this way - until we get more flexible configuration support.

doragasu:
Thanks for the feedback! No problem, your planned implementation looks way better and more flexible than mine. I'm looking forward to seeing it!

PR #88, opened 2016-02-18
Add (again) ds18b20 driver and examples.

hetii:
Cause I produce a bit mess in my last pull request I add here a clear version of commit 72f30ad.


foogod:
Thanks.. It probably would have been better if this stayed as two separate commits instead of squashing them, just because it would be easier to see what changes you made in response to the earlier comments (but in this case it's still small enough it's not a huge deal, I think)..

This looks pretty good to me..  I'm not sure whether we really need a separate copy of the MIT license text in each directory (it would probably just be enough to state in each file that it's MIT-licensed), but I suppose it doesn't hurt anything at this point, and is better than not having enough license info..  @projectgus, what do you think?  Are your earlier concerns all addressed with this version?

 (I also have a couple of ideas for improvements/enhancements, but will submit my own PR later for that stuff)

foogod:
Actually, now that I think about it.. the actual onewire and ds18b20 bits that were ported from other projects are, of course, subject to the licensing of those other projects, but the example programs you included are entirely your own work, right?  If so, would it be OK if we licensed those bits under the same license as the rest of esp-open-rtos (3-clause BSD instead of MIT, more or less the same but with different wording) so we don't need as many different LICENSE files all over the place?

I'd like to keep everything under one consistent license as much as we can, at least for original contributed work..

hetii:
@foogod 
From my point of view you can put there any license you need.
Examples are written by me, based on the knowledge that I found over Internet.

extras/ds18b20/ds18b20.h
```diff
@@ -0,0 +1,20 @@
...
+typedef struct {
+	uint8_t id;
+    uint8_t major;
+    uint8_t minor;
+} ds_sensor_t;
+
+// Scan all ds18b20 sensors on bus and return its amount.
+// Result are saved in array of ds_sensor_t structure.
+// Cause printf in esp sdk don`t support float,
+// I split result as two number (major, minor).
```
foogod: I hadn't noticed this comment before and had been wondering about this..

I've opened bug #95 for the printf issue, so hopefully we can get that fixed..  In the meantime, though, this really isn't the right way to work around it anyway.  We should really just return a float from these routines, and then the caller can do a conversion *if* they want to print it..  This way makes it easier to print but substantially harder to use for any other purpose.

extras/ds18b20/ds18b20.c
```diff
@@ -0,0 +1,105 @@
...
+        uint8_t temp_lsb = get[0]; // Temp data plus lsb
+        uint16_t temp = temp_msb << 8 | temp_lsb;
+        
+        float temperature;
+
+        temperature = (temp * 625.0)/10000;
+        //printf("Got a DS18B20 Reading: %d.%d\n", (int)temperature, (int)(temperature - (int)temperature) * 100);
+        result[sensor_id].id = sensor_id;
+        result[sensor_id].major = (int)temperature;
+        result[sensor_id].minor = (int)(temperature) - (int)temperature * 100;
```
foogod: As mentioned elsewhere, we shouldn't be doing this conversion here anyway. But I also wanted to point out that it looks like the parentheses are wrong here, so I'm pretty sure this can't be returning the right value anyway..

foogod:
@hetii: So I've gone ahead and done some fixes for the couple of comments I mentioned above, as well as cleaning up some of the license files, and I've submitted a PR from my own repo into your branch.  If you go ahead and merge that PR, the updates should also show up here, and then I think this PR should be ready to merge as well.

hetii:
@foogod 
Ok thx for your fixex, all is merged from my side.
Regards

foogod:
Great! @projectgus, unless you have any objections, I'm going to go ahead and merge this one..

examples/ds18b20_broadcaster/ds18b20_broadcaster.c
```diff
@@ -0,0 +1,120 @@
+#include <string.h>
```
projectgus: Nice example! Can you put a header and a copyright notice (ideally public domain/CC0) on it please?

projectgus:
Looks great to me, apart from the one inline comment above.

As an aside. at some point I think we'll need to scoop up all the small peripheral drivers in `extras/` and group them under one component - `extras/peripherals` or something. But that can be done separately at some point in the future.

hetii:
@projectgus Ok I add required by you additional information.

foogod:
I'm gonna go ahead and merge this..  I think we've made @hetii jump through enough hoops :) and it'll be handy to have this part of the mainline (I have some additional enhancements I'm going to be putting together and submitting once this is merged).

projectgus:
:+1: and thank you @hetii for being patient and helpful :)

PR #92, opened 2016-02-20
Added possibility to enable internal pull-up resistors

Nurgak:
I was working on an 5V I2C device (LCD screen) which had pull-up resistors on the SCL an SDA lines, since the ESP is not 5V tolerant I removed them. 3.3V is apparently enough in my application to make the slave recognise a logical 1.

This modification lets the user choose to enable the ESP internal pull-up resistors for the SCL and SDA lines with a new function: `i2c_pullups(boolean)`. This way the circuit does not need more external parts. ~~I have tested the code and checked the signal on an oscilloscope, despite the SCL line not rising to 3.3V it works well.~~ Edit: this is actually not reliable, but could work at slower speeds...

I2C using ESP internal pull-up resistors

~~There's a risk in driving the SDA and SCL lines to Vcc. In my example I use a ESP8266-01 board where only the GPIO0 and 2 are available, just enough for I2C. GPIO0 must be connected to 0V when programming, but as soon as the program starts and drives the line to 3.3V it basically shorts Vcc to 0V via this GPIO0, an open-drain output allows for safety in this situation.~~

~~Ultimately driving the SCL and SDA lines to Vcc is a bad idea and should be removed I think (at least as the default configuration). If the slave drives the line to 0V for some reason it shorts out and the ESP probably dies, open-drain output is basically the whole point of I2C.~~

Never mind, the lines were never driven to Vcc, however internal pull-up resistors is still a neat feature.

### DIFF
--- a/extras/i2c/i2c.c
+++ b/extras/i2c/i2c.c
@@ -36,14 +36,21 @@
 #define CLK_STRETCH  (10)
 
 static bool started;
+static bool pullups;
 static uint8_t g_scl_pin;
 static uint8_t g_sda_pin;
 
 void i2c_init(uint8_t scl_pin, uint8_t sda_pin)
 {
     started = false;
+    pullups = false;
     g_scl_pin = scl_pin;
     g_sda_pin = sda_pin;
+}
+
+void i2c_pullups(bool enable)
+{
+	pullups = enable;
 }
 
 static void i2c_delay(void)
@@ -54,14 +61,14 @@ static void i2c_delay(void)
 // Set SCL as input and return current level of line, 0 or 1
 static bool read_scl(void)
 {
-    gpio_enable(g_scl_pin, GPIO_INPUT);
+    gpio_enable(g_scl_pin, pullups ? GPIO_INPUT_PULLUP : GPIO_INPUT);
     return gpio_read(g_scl_pin); // Clock high, valid ACK
 }
 
 // Set SDA as input and return current level of line, 0 or 1
 static bool read_sda(void)
 {
-    gpio_enable(g_sda_pin, GPIO_INPUT);
+    gpio_enable(g_sda_pin, pullups ? GPIO_INPUT_PULLUP : GPIO_INPUT);
     // TODO: Without this delay we get arbitration lost in i2c_stop
     i2c_delay();
     return gpio_read(g_sda_pin); // Clock high, valid ACK
--- a/extras/i2c/i2c.h
+++ b/extras/i2c/i2c.h
@@ -49,3 +49,7 @@ bool i2c_slave_read(uint8_t slave_addr, uint8_t data, uint8_t *buf, uint32_t len
 // devices where the i2c_slave_[read|write] functions above are of no use.
 void i2c_start(void);
 void i2c_stop(void);
+
+// Use internal pull-up resistors instead of external ones,
+// by default they are disabled.
+void i2c_pullups(bool enable);


foogod:
Seems reasonable.. I would worry a bit about whether the internal pull-up is strong enough to reliably meet timing requirements in many situations, but it's probably fine for short lines to a single device..

I'm actually wondering whether there's any situation where it would be bad to have this enabled anyway.. I'm not sure exactly what resistance the ESP8266's internal pull-ups are (and can't seem to easily find anything that will tell me) but I suspect they're high enough that even if you needed to add an external one having the internal one enabled too probably wouldn't affect the value much..

projectgus:
This looks good to me, thanks @Nurgak.

The internal pullup values are between 30K and 100K, according to an Espressif post here: http://bbs.espressif.com/viewtopic.php?t=1079#p4097

The only situation I could think of where you wouldn't want the soft pullups is if you're pulling up to a different VCC, maybe for 1.8V logic. Worst case scenario, a hard 4.7K pullup to 1.8V and a soft 30K pullup to 3.3V would give a high voltage of 2.04V. I don't think this would be harmful (not that high voltage, and a high impedance current source), but it might be unexpected.

That said, I'd be happy to have soft pullups default to "on", given most people use them?

Another thing I noticed, it would be possible to do away with the 'pullup' variable entirely if that's desirable. The flags for setting output mode and for enabling pullups are separate bit flags on the iomux register - IOMUX_PIN_OUTPUT_ENABLE and IOMUX_PIN_PULLUP. 

We just don't expose them in the gpio.h interface for now.

It might make even sense to add another mode for the `gpio_enable` call - GPIO_OUT_OPEN_DRAIN_PULLUP, which gives you an open drain output with a soft pullup on it? The i2c interface could then use this, and it's probably useful to other people too?

... anyhow, I'm just throwing out some ideas here. I think this is a good PR as-is, so if noone wants to change anything then I'd be more than happy to merge it. Let me know. :)

foogod:
Good point about the possibility of other bus voltages.. I agree it'd probably be fine to make it on by default, though..

Actually, I was just running into exactly the same thing with the DS18B20 stuff, and was also thinking that this is something we might just want to find a way to handle independent of each individual driver having to have its own code for it..  I was thinking about maybe just adding some gpio_* call to allow the app writer to specify (persistently) for any given pin whether GPIO_INPUT should enable the pullup or not (so the drivers don't need any changes and it all happens automatically)

projectgus:
How about a `gpio_enable_pullup(int, bool)` that just sets/clears the IOMUX_PIN_PULLUP bit?

That should work OK as long as nothing else clobbers that bit in the iomux register. Probably the difficult thing is to know how it interacts with gpio_enable(GPIO_INPUT) vs gpio_enable(GPIO_INPUT_PULLUP)  - if GPIO_INPUT_PULLUP enables pullups then you'd expect GPIO_INPUT to clear them?

We could do away with GPIO_INPUT_PULLUP entirely I guess, recommend people use both functions if they want that mode?

foogod:
That wouldn't be enough for some things (like the DS18B20 stuff, which needs to switch back and forth between input and output, and we don't want the pullup enabled during output, so there needs to be some way to remember that it should be turned back on when it switches back to input mode)..

Regarding the fate of GPIO_INPUT_PULLUP, we could alternately leave it in and add another GPIO_INPUT_NOPULLUP, so somebody could use those two to override whatever the pin configuration has been set to for specific cases, or (normally) use GPIO_INPUT when they don't care and want to let the app developer decide..

projectgus:
GPIO_INPUT_NOPULLUP sounds good

Is leaving the pullup on in output mode really a no-go? When driving low, the gate will sink an additional 100uA of current through a 30K pullup. It'll probably slow the falling edge of the signal a very tiny bit when driving low, but I can't think of too many circumstances where that would matter?

foogod:
Actually, after I said that I thought about it some more and, yeah, I guess there isn't really any problem with having the pullup enabled in that case.. given that, your suggested approach would probably be the most straightforward way to deal with this issue..

If we're going to do that, though, then GPIO_INPUT_PULLUP and GPIO_INPUT_NOPULLUP wouldn't be temporary overrides, but would permanently change the setting, and thus we'd have two (potentially confusing) ways to change things, which I'm not a huge fan of, so I'm actually leaning towards just dropping them and only having GPIO_INPUT, GPIO_OUTPUT, and GPIO_OUT_OPEN_DRAIN for `gpio_enable`..

Nurgak:
Pull-up resistors really do not have any function when the pin is low impedance, as you say it will only slow down the edge of the signal when pulled the opposite way. To me `GPIO_INPUT` and `GPIO_INPUT_PULLUP` are explicit enough. There are pull-down resistors on the ESP as well, might be nice to have a new case in `esp_gpio.h`:

    case GPIO_INPUT_PULLDOWN:
        iomux_flags = IOMUX_PIN_PULLDOWN;
        break;

I think the user should be aware of the pin states and understand what is being done at the hardware level, there's no way to make this monkey-proof. `GPIO_INPUT_NOPULLUP` is not necessarily very explicit for people who do not know what a pull-up resistor is... `GPIO_INPUT` is easier to understand IMHO.

By the way what are `IOMUX_PIN_PULLUP_SLEEP`, `IOMUX_PIN_PULLDOWN_SLEEP`, `IOMUX_PIN_FUNC_LOW_M`... in `iomux_regs.h` used for?

foogod:
On a related note, though, I just noticed that the existing i2c code is using GPIO_OUTPUT when writing.. shouldn't those be GPIO_OUT_OPEN_DRAIN instead?

Nurgak:
`GPIO_OUTPUT` is good the way it is used now, but only when driving the pins low. Never actively drive pins high on an I2C bus, another element might try to drive its pin low and something will give up its magic smoke.

foogod:
@Nurgak The problem with that approach, though, is what we've just been discussing:  Since the pullup (or pulldown) setting is intermingled with the direction in the `gpio_enable` call, each driver has to have custom code to handle whether or not the pullups should be used.  If we separate out the pullup configuration into a separate interface, then the app writer can just set it themselves on the pin however they want, and the driver doesn't have to care, it just uses `GPIO_INPUT` / `GPIO_OUTPUT` / etc and the pullup stays however the user wants it to be (which means we don't need driver-specific hacks like this one then).

`IOMUX_PIN_FUNC_LOW_M` is one part of the mask/shift constants used for setting the function for the pin.  You generally won't need to mess with it directly (that's part of what `gpio_enable` does).

I've never actually tried them myself, but if I understand the ESP8266 capabilities correctly, I believe `IOMUX_PIN_PULLUP_SLEEP` and `IOMUX_PIN_PULLDOWN_SLEEP` determine whether the pullup/pulldown resistors are active when the chip is in sleep mode or not.

These other options you mention are actually another good argument for splitting pullup/pulldown out into a separate interface, as otherwise the `gpio_enable` options could get unnecessarily complicated (`GPIO_INPUT_PULLUP_WITH_PULLDOWN_DURING_SLEEP`, etc).  If we had a separate `gpio_set_pullup(pin, type, sleep_type)`, then we could have type settings of `GPIO_PULLUP`, `GPIO_PULLDOWN`, or `GPIO_NONE` (or whatever) and not pollute the `gpio_enable` options with all the multitude of combined configuration possibilities (though `gpio_set_pullup` may not be the best name if it also supports pull-down.. not sure what a good name would be).

Regarding using `GPIO_OUTPUT`, I suppose you're right if they are only ever driven low, but it still seems sloppy to me.  Worse, though, is that there's nothing in this code to prevent a context switch between setting the pin to output mode and the call to pull it low, which means in theory, when it gets set to output mode, it could end up being driven high initially, then a context switch to another task could cause it to stay driven high for a fair bit of time before it is finally set low.  If something else on the bus is pulling things low at the time this happens, it could cause real problems.  Using the correct mode (`GPIO_OUT_OPEN_DRAIN`) would avoid this problem.

foogod:
Actually, another advantage of open drain mode is that I believe you can actually read the pin as an input (and thus read the line level) even when the output drivers are enabled, which means we could just set it to `GPIO_OUT_OPEN_DRAIN` once at the beginning of things and then never need to change it back to `GPIO_INPUT` at all.

Nurgak:
@foogod I see now. What you say makes sense, it's probably a more elegant way to future-proof this part.

I come from Arduino/Atmel where these two functions are intermingled at a hardware level: the only way the pull-up resistors can be enabled on some ATmega MCUs is to set the pin mode to input (DDR = Data Direction Register) and the pin value (PORTx) to high, essentially there are only 3 states possible on an ATmega (input, output and input+pull-up). On an ESP things are more interesting with more possibilities, separating those functions makes sense as they are separated on register/hardware level.

Nurgak:
@foogod I'm not very familiar with `GPIO_OUT_OPEN_DRAIN` mode and can't make much sense from the code. If you say you can supposedly read the pin in `GPIO_OUT_OPEN_DRAIN` mode then what is the difference with `GPIO_INPUT_PULLUP`?

I'm trying it out right now and in `GPIO_OUT_OPEN_DRAIN` mode the lines are not being pulled to 3.3V when calling `gpio_write(g_sda_pin, 0);`.

Nurgak:
A bit off topic, but somewhat related to what I said in the beginning: the screen I'm using (HD44780 with a PCF8574 I/O I2C Expander) has stopped working when only using internal pull-ups... it seems the internal pull-up resistors are not small enough to reliably drive the I2C bus, so I was really lucky it worked at first.

I tested with another ESP with the original code and got the same results, so this did not damage the pins. With external 1K pull-ups everything works again.

If there was a to slow down the communication on the I2C bus it might solve this issue...

foogod:
`GPIO_OUT_OPEN_DRAIN` is exactly the same as `GPIO_OUTPUT` except if you set it "high" it doesn't drive the pin high, but just lets it float (so you can drive it down, but not up).  It's designed to be used for applications like this that will have a pull-up resistor on the line to pull it high.

Unfortunately, at the moment, `gpio_enable(pin, GPIO_OUT_OPEN_DRAIN)` does not enable the internal pull-up resistor, so the bit I was saying about reading the bus state without having to switch to `GPIO_INPUT` / `GPIO_INPUT_PULLUP` will probably only work if you have an external pullup resistor on the line (but this could be fixed if we do what we were talking about about controlling pullups separately from `gpio_enable`).  I believe it will work, but I haven't played with this stuff in a while.  I should have a little bit of time later today to play with all this and maybe implement what we were talking about and verify that this all works as I think it should.

foogod:
Ok, a couple of updates from my testing:

1. The internal pull-up resistors in the ESP8266 are 47k (this was determined by setting a pin's internal pull-up to enabled, pulling the pin to ground, and measuring the resulting current (about 680 uA for a 3.2V supply voltage)..
2. Setting either `IOMUX_PIN_PULLUP` or `IOMUX_PIN_PULLDOWN` both seem to have the same effect, and both actually enable a *pull-up* resistor.  There does not appear to be any way to do internal pull-down on the ESP8266 GPIO lines (this actually jives with what the datasheet claims (it only claims pull-up capability, not pull-down) and with what some other users on various forums appear to have experienced).  Given this, there doesn't seem to be any point to exposing the pull-down bits in our API..
3. As I suspected, I have confirmed that it is possible to read the actual state of the pin with `gpio_read` even when in `GPIO_OUT_OPEN_DRAIN` mode, so it is not necessary to switch back and forth between that and `GPIO_INPUT` (just set the pin high in open drain mode and you can read it like a normal input).

I've just about finished an update that splits out the pullup configuration from `gpio_enable` into a `gpio_set_pullup` call as previously described, and should be putting together a PR soon.. Hopefully this will make the whole thing more flexible in general and remove the need for things like this PR..

projectgus:
Nice work!

1) Nice measurement. I think one should still consider the 30K to 100K range posted by Espressif as the possible range though (with 47K "nominal"), because there may be fabrication tolerances (or differences between series of chip) that we're not considering.

2) My understanding is that normal GPIOs have soft pullups and no pulldown, and GPIO16 has a soft pulldown but no pullup. My tests confirmed the first behaviour (same as yours), but I haven't tested GPIO16. As you probably know, GPIO16 is controlled by the RTC register bank and at the moment we don't have support for it at all in `gpio.h`.

3) That's excellent news!

One request if you're making changes on this stuff anyhow, can you please add some comments in gpio.h about details like the pullup values. It's not the best place for documentation, but it is a place where people will look!

foogod:
Do you have a reference for that "30K - 100K" info?  I can't find it (or anything else) anywhere in any of Espressif's docs..  (I would certainly hope "fabrication tolerances" aren't *that* bad for this stuff.. I suppose it's possible that they could change it for different revisions of the chip, but seems unlikely they would..)

I can certainly add some info about this in the headers (I'm also gonna update the wiki) but would really like to have source info for anything I do that with..

projectgus:
The place I got it from (linked above) was a forum post by an Espressif employeee. http://bbs.espressif.com/viewtopic.php?t=1079#p4097 Which is about as official as most of Espressif's information releases get. ;) 

I guess I really meant tolerances between revisions or runs in different fab facilities. Reason I was thinking that way is that I've seen other MCU vendors also give fairly hand-wavey values for their internal pullups. The other factors I can think of is that maybe variation is high across the operating temperature range, and/or that heat dissipation from the resistor itself causes significant temperature variation due to localised heating if you leave the pullup on for a while? I'm just guessing though, really...

and thanks! re: headers & wiki. :)

projectgus:
Veering a bit O/T here, but I was curious about making resistors in silicon. This lecture goes over a few methods used for analog cmos http://www.aicdesign.org/SCNOTES/2010notes/Lect2UP070_%28100419%29.pdf
... says large resistance values are often made as 'n-well resistors', +/-40% absolute accuracy. I don't know enough about chip fabrication to know what techniques would apply to ESP8266 though. 30K-100K is +/-54% anyhow, which is more than what's mentioned for any of these techniques (even once you add a few percent across the temp range). ???

foogod:
FYI, I've submitted PR #94 with the mentioned changes to `gpio_enable`, etc..

projectgus:
#94 is merged, thanks for that.

@Nurgak, are you able to update the i2c driver here to use GPIO_OUT_OPEN_DRAIN and gpio_set_pullup()? It should not be necessary to change the mode to GPIO_INPUT in order to read the bus state.

Nurgak:
I tried with the `GPIO_OUT_OPEN_DRAIN`, still using external pull-ups and it turns out bit-banging this way is much faster, way too fast for my HD44780 screen, but good enough for my I2C GPIO expander, so I don't get the "i2c arbitration lost" messages, but the screen won't display anything.

I guess speed control is another task, but it really needs to be implemented as well...

foogod:
With the changes in #94, is there a need to keep this PR open?  It sounds like we still want to do some work with `GPIO_OUT_OPEN_DRAIN`, etc, but that will require some more work to get it working reliably, and should probably be a separate PR from this anyway..

Nurgak:
I agree, also the ~100kbit/s is not valid anymore with `GPIO_OUT_OPEN_DRAIN` as there's no that much overhead anymore. How would one even change the bus speed in this situation?

PR #94, opened 2016-02-22
Separate pullup config out of `gpio_enable()`

foogod:
Created `gpio_set_pullup` to configure pullups independently of direction.
Removed GPIO_INPUT_PULLUP direction type.
Added misc other helper functions in iomux.h

### DIFF
--- a/core/include/esp/gpio.h
+++ b/core/include/esp/gpio.h
@@ -17,13 +17,20 @@ typedef enum {
     GPIO_INPUT,
     GPIO_OUTPUT,         /* "Standard" push-pull output */
     GPIO_OUT_OPEN_DRAIN, /* Open drain output */
-    GPIO_INPUT_PULLUP,
 } gpio_direction_t;
 
-/* Enable GPIO on the specified pin, and set it to input/output/ with
- *  pullup as needed
+/* Enable GPIO on the specified pin, and set it to input or output mode
  */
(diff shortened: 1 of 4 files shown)


core/include/esp/gpio.h
```diff
@@ -36,6 +43,19 @@ static inline void gpio_disable(const uint8_t gpio_num)
     *gpio_iomux_reg(gpio_num) &= ~IOMUX_PIN_OUTPUT_ENABLE;
 }
 
+/* Set whether the specified pin continues to drive its output when the ESP8266
+ * goes into sleep mode.  Note that this setting is reset to off whenever
+ * gpio_enable is called, so this must be called after calling that function.
+ */
```
projectgus: Is this confirmed working by anyone? Does it work in "deep sleep"?

foogod: I actually don't know..  I also haven't actually verified the `IOMUX_PIN_PULLUP_SLEEP` stuff either.  It's one of those things on my list to look into someday, but I've still never even gotten as far as playing around with the sleep stuff.

I added the function for completeness, though, because otherwise (assuming it does work) there was a way to set it before and isn't anymore (at least using the `gpio_*` routines)..  We can take it out (or annotate it) if it's a concern, though.

projectgus: I think it's probably a good idea to mark it as something that's untested, then. I think I played with a bit once without success, but I didn't keep notes.

Other than that, looks great to me!

PR #98, opened 2016-02-23
Add driver and example for Samsung vfd16lf01 display.

hetii:


extras/vfd16lf01/LICENSE
```diff
@@ -0,0 +1,7 @@
+This is a driver for Samsung 16LF01 series VFD.
+
+Originate written by Daniele Napolitano, distributed under GPL.
```
foogod: Sorry, this is a problem..  We can't accept any code that's under the GPL, because it contaminates our own BSD licensing.  You'll need to find or make something under a more compatible license..

hetii: What In the case if I will get permission from author to use this code under other license ?

foogod: If you can get permission from all of the original authors (in this case it looks like there's only one, which makes it easier) to distribute this under a more permissive license (i.e. BSD or MIT), then yeah, that would work..

If you do get that permission, you should change this bit to say something like "originally distributed under the GPL, but special permission given to license this code under BSD for esp-open-rtos"

hetii: Ok but to be honest I don`t get it.

Driver like this is in fact a wrapper on data-sheet of IC. There is no special magic behind of it.
Even if I will build driver from scratch it will looks in the same way (or at least very similar) cause its a way how this display communicate with the world.

projectgus: Unfortunately that's Copyright law. :(

You can recreate it from scratch, but you can't derive it from this source code without it falling under GPL.

sheinz:
I'll close it for now. If the situation with licensing changes feel free to reopen this or open a new one.

PR #100, opened 2016-02-24
fix sdk_bss_info STAILQ_ENTRY

jkent:
I'm certain this was an oversight... This is a needed change for the cgiwifi stuff in esphttpd.


projectgus:
Yes, that's an oversight! Oddly it can still work, I think you need to add an unnecessary cast though.

PR #102, opened 2016-02-28
Fix delays in DS18B20 driver

darksv:
There is a bug in the driver for DS18B20: instead of delaying for 750**ms** it waits only for 750**us**, which obviously causes that read temprature is not valid (it's default 85 degrees).

### DIFF
--- a/extras/ds18b20/ds18b20.c
+++ b/extras/ds18b20/ds18b20.c
@@ -1,3 +1,6 @@
+#include "FreeRTOS.h"
+#include "task.h"
+
 #include "onewire/onewire.h"
 #include "ds18b20.h"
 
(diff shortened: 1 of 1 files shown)


foogod:
Ah, good catch.. On the couple of devices I have, it doesn't seem to return bad values, but I suspect it may just be returning the results from the previous read instead of the one that was just initiated.. Might explain why it seemed to be a little slow to react to changes sometimes..

MightyPork:
ah, so `sdk_os_delay_us()` is how you do precise delays. I've been asking this yesterday on IRC. Will be handy for the ws2812b driver I'm building..

projectgus:
@foogod - are you OK to merge this then?

@MightyPork - yes, that's the current method. I keep meaning to re-implement it as it should be pretty easy to write our own one. Bear in mind that the NMI (Non-Maskable-Interrupt) means precise delays can be less precise than you hoped!

foogod:
Yeah.. I had intended to do a quick double-check on my own DS18B20 setup before merging, but then got sidetracked and didn't actually follow through.. sorry about that.

Looks good to me, and does appear to make things substantially more stable.

Issue #82, opened 2016-01-17
Support for 160MHz CPU frequency


projectgus:
As reported on IRC, there's no function prototype for sdk_system_cpu_freq(uint8_t freq) in the headers, although the function exists in the libraries.

esp-open-rtos should either implement a prototype for that function, or (better) write a new implementation that sets the appropriate register.

elmorejd:
That solved the issue. Added "void sdk_system_update_cpu_freq( uint8_t )" to "espressif/esp_system.h" which seemed to be the most accurate place to put it. I will keep you updated if anything wonky happens.

projectgus:
These functions are now part of `espressif/esp_system.h`.

Leaving this issue open as it'd be good to have a from-scratch reimplementation of CPU frequency setting and the relevant registers.

sheinz:
I'll close it as `sdk_system_update_cpu_freq`  is reimlemented [here](https://github.com/SuperHouse/esp-open-rtos/blob/master/open_esplibs/libmain/user_interface.c#L429)

PR #104, opened 2016-02-28
Added definition of SDK functions

UncleRus:


projectgus:
Thanks for those, looks good. This gets us part-way towards #82 as well, although I'll leave that issue open as it'd be good to just re-implement the CPU speed switching function.

include/espressif/esp_system.h
```diff
@@ -57,6 +58,32 @@ bool sdk_system_rtc_mem_read(uint8_t src, void *dst, uint16_t n);
...
+   Voltage range 0 .. 1.0V. RF must be enabled.
+   Example:
+      printf ("ADC voltage %.3f", 1.0 / 1024 * sdk_system_adc_read());
+ */
+uint16_t sdk_system_adc_read(void);
+/*
+   Measure power voltage. Returns 10 bits ADC value (1/1024V).
+   TOUT pin must be floating and RF must be enabled.
+   Example:
+      printf ("Power voltage %.3f", 1.0 / 1024 * sdk_system_get_vdd33());
```
projectgus: I just noticed this post-merge, but is there a scaling factor we can apply to this example to get VDD in volts?

I'm guessing the ADC full-scale voltage is a bandgap reference (usually 1.1V but might be 1.0V), so measuring VDD must involve dividing the input voltage down and comparing it to the bandgap value?

UncleRus:
I guess reference ADC voltage is 1V and we can't change it.
Anyway, voltage divider on the TOUT pin can be used, like in NodeMCU.

PR #105, opened 2016-02-29
Removed sdk_system_get_vdd()

UncleRus:
sdk_system_get_vdd() was added by mistake in #104


projectgus:
Ah you're right, that function is missing from the version of the binary SDK we have. We'd need to recreate it from source.

PR #106, opened 2016-03-01
WS2812B driver & example

MightyPork:
Now we have a »NeoPixel« support!

The demo displays a fancy animated rainbow generated on-the-fly.

### DIFF
--- a/examples/ws2812_rainbow/Makefile
+++ b/examples/ws2812_rainbow/Makefile
@@ -1,0 +1,6 @@
+# Makefile for the ws2812 Rainbow example
+
+PROGRAM=ws2812_rainbow
+EXTRA_COMPONENTS = extras/ws2812
+
+include ../../common.mk
--- a/examples/ws2812_rainbow/ws2812_rainbow.c
+++ b/examples/ws2812_rainbow/ws2812_rainbow.c
@@ -1,0 +1,155 @@
+/**
+ * @file   es2812_rainbow.c
+ * @author Ondřej Hruška, 2016
+ *
+ * @brief  Example of a rainbow effect with
+ *         WS2812 connected to GPIO2.
+ *
+ * This demo is in the public domain.
+ */
+
+#include "espressif/esp_common.h"
+#include "FreeRTOS.h"
+#include "task.h"
+#include "esp/uart.h" // uart_set_baud
+#include <stdio.h> // printf
+#include <stdint.h>
+
+#include "ws2812.h"
+
+
+#define delay_ms(ms) vTaskDelay((ms) / portTICK_RATE_MS)
+
+
+/** GPIO number used to control the RGBs */
+static const uint8_t pin = 2;
+
+
+/**
+ * @brief "rainbow" animation with a single RGB led.
+ */
+void demo_single(void)
+{
+    // duration between color changes
+    const uint8_t delay = 25;
+
+    ws2812_rgb_t x = {.num = 0xFF0000}; // RED color
+
+    while (1) {
+        // iterate through the spectrum
+
+        // note: This would be _WAY_ easier with HSL
+
+        while(x.c.g < 0xFF) { x.c.g++; ws2812_set(pin, x.num); delay_ms(delay); } // R->RG
+        while(x.c.r > 0x00) { x.c.r--; ws2812_set(pin, x.num); delay_ms(delay); } // RG->G
+        while(x.c.b < 0xFF) { x.c.b++; ws2812_set(pin, x.num); delay_ms(delay); } // G->GB
+        while(x.c.g > 0x00) { x.c.g--; ws2812_set(pin, x.num); delay_ms(delay); } // GB->B
+        while(x.c.r < 0xFF) { x.c.r++; ws2812_set(pin, x.num); delay_ms(delay); } // B->BR
+        while(x.c.b > 0x00) { x.c.b--; ws2812_set(pin, x.num); delay_ms(delay); } // BR->R
+    }
+}
+
+
+/**
+ * @brief "rainbow" effect on a RGB strip (30 pixels - can be adjusted)
+ *
+ * This example shows how to use the "procedural generation" of colors.
+ *
+ * The pixel colors are calculated on the fly, which saves RAM
+ * (especially with large displays).
+ */
+void demo_strip(void *pvParameters)
+{
+    const uint8_t anim_step = 10;
+    const uint8_t anim_max = 250;
+
+    // Number of your "pixels"
+    const uint8_t pixel_count = 30;
+
+    // duration between color changes
+    const uint8_t delay = 25;
+
+    ws2812_rgb_t color = WS2812_RGB(anim_max, 0, 0);
+    uint8_t step = 0;
+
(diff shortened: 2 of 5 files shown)


extras/ws2812b/ws2812b.h
```diff
@@ -0,0 +1,210 @@
...
+
+/**
+ * @brief Struct for easy manipulation of RGB colors.
+ *
+ * Set components in the xrgb.c.r (etc.) and you will get
+ * the hex in xrgb.num.
+ */
+typedef union {
+
+    /** Struct for ccess to individual color components */
```
projectgus: Tiny typo :)

extras/ws2812b/ws2812b.h
```diff
@@ -0,0 +1,210 @@
...
+        uint8_t b;
+        uint8_t g;
+        uint8_t r;
+        uint8_t reserved; /**< padding field, could be used for brightness (?) */
+    } c;
+
+    /** RGB color as a single uint32_t */
+    uint32_t num;
+
+} xrgb_t;
```
projectgus: Is it possible to namespace these and the functions that manipulate them, somehow? ws2812b_rgb_t if it's not too long? Or could abbreviate ws2812 if that's easier.

extras/ws2812b/ws2812b.h
```diff
@@ -0,0 +1,210 @@
...
+
+
+/**
+ * @brief Macro to compose a xrgb_t constant.
+ */
+#define rgb2xrgb(r, g, b) {.num = (((r) & 0xFF) << 16) | (((g) & 0xFF) << 8) | ((b) & 0xFF)}
+
+#define hex2xrgb(rrggbb) {.num = (rrggbb) & 0xFFFFFF}
```
projectgus: Namespacing these would be good, to. Also we generally follow the C convention of uppercase macro names - they're ugly but as least you know when you're calling them.

Is it possible to just use C99 syntax for these though? ie

const xrgb_t rgb_val = { .r = 127, .g = 0, .b = 64 };

const xrgb_t as_rgb = { .num = 0x7F0080 };

 ? I can't remember how much anonymous union support you get under 'gnu99', maybe this isn't quite possible?

extras/ws2812b/ws2812b.h
```diff
@@ -0,0 +1,210 @@
...
+ * the hex in xrgb.num.
+ */
+typedef union {
+
+    /** Struct for ccess to individual color components */
+    struct __attribute__((packed)) {
+        uint8_t b;
+        uint8_t g;
+        uint8_t r;
+        uint8_t reserved; /**< padding field, could be used for brightness (?) */
```
projectgus: I don't think you need a padding field here, sizeof(xrgb_t) will be the size of the largest thing in the union (ie the uint32_t).

extras/ws2812b/ws2812b.h
```diff
@@ -0,0 +1,210 @@
...
+
+#define hex2xrgb(rrggbb) {.num = (rrggbb) & 0xFFFFFF}
+
+/**
+ * @brief Set a single RGB LED color, and display it.
+ *
+ * @param gpio_num : data line GPIO number
+ * @param rgb : RGB color - 0x00RRGGBB
+ */
+void ws2812b_set(uint8_t gpio_num, uint32_t rgb);
```
projectgus: What's the rationale behind having these take uint32_t instead of xrgb_t?

extras/ws2812b/ws2812b.h
```diff
@@ -0,0 +1,210 @@
...
+#endif
+
+// 800 ns
+#ifndef WS_LONG_DELAY
+#define WS_LONG_DELAY()  for (volatile uint32_t __j = 4; __j > 0; __j--)
+#endif
+
+
+// inline attribute for functions that need zero overhead.
+#define INLINEFUNC static inline __attribute__((always_inline))
```
projectgus: Does this turn out to be necessary? I was a bit hooked on this kind of macro, a habit I picked up with old versions of avr-gcc that would fail to inline very obviously inlinable things!

But @foogod told me off and I went and pulled them all out of the codebase, without any major side effects that I can find. :)

projectgus:
Hey,

This is really cool, thanks for submitting it! I was hoping we'd end up with RGB LED support sooner or later. :) :)

I left some comments in the diffs - they're all pretty minor things, nothing super important.

I think it is important to explain to people (maybe in the header file) that even though we can disable interrupts we can't (yet) reliably disable the NMI, so there is always a chance (especially under heavy WiFi traffic) that the NMI will fire and mess up the WS2812B bit bang timing -so you'll see glitching.

Try ping flooding the example sketch (while it's on the WiFi) and I think you'll almost definitely see this happening, unfortunately. :( :(

Eventually it should be possible for us to disable NMI for at least short periods, enough to drive short chains of LEDs at least - hopefully longer. But it's not possible now.

It may be possible at the moment to detect (using the CCOUNT register, or by setting a flag) when an NMI has occured so at least know that you have to start over. But I'm happy to merge this without any such gold plating, we can always gold plate it later. :)

Yay blinkenlights!

extras/ws2812b/ws2812b.h
MightyPork: I was worried it might be adjusted to the wrong side, but just guessing here. Will try if it works without.

extras/ws2812b/ws2812b.h
MightyPork: you can use RGB hex directly this way, the xrgb_t stuff is optional. Much easier if you just want to set an indicator color.

```
ws2812b_set(0xFFFF00); // yellow
```

extras/ws2812b/ws2812b.h
MightyPork: do you mean the `__attribute__((always_inline))` or the whole thing?
The idea is to avoid overhead from calls, will it be inlined if they're regular functions in the .c file?
There's not much room for call overhead here.

extras/ws2812b/ws2812b.h
MightyPork: I tried & the former does not work, it would have to be:

```c
ws2812_rgb_t color = {.c.r = 127, .c.g = 0, .c.b = 64};
```

which is pretty ugly. The second option works fine.
Maybe there's some better way to define the struct?

MightyPork:
Think I addressed most of the issues, but there's still work to be done - don't merge it just yet.

So far, I did it all in `user_init()`, without a separate task—not sure if that's a problem. Now I tried moving it to a task and it all blew up, but that may just some mistake. Needs further investigation.

I'm also noticing glitches now. However, it's the same breadboard setup with the boot-loops and mysterious errors, so I'll see how it performs when I get a better wiring in place.

MightyPork:
I've done some more testing, also with a task (didn't mess up this time, yay!).
You were absolutely right with the NMI, I see a glitches when it's running in a task.

Curious side effect (?) is that it also garbles UART output, I'm not sure if it's again an issue with my breadboard, or something weird going in with the interrupts. Either way, lights work mostly OK!

All the commits are now squashed for easier merging / review.

PR #107, opened 2016-03-01
Driver for PCF8574 (8-bit I2C GPIO expander)

UncleRus:
Simple GPIO expander driver.


foogod:
Sorry, but we can't accept any code under the GPL as it is incompatible with esp-open-rtos' BSD licensing.

If you'd be willing/able to relicense this under a BSD / MIT (or compatible) license instead, we would definitely be interested, though..

UncleRus:
I think MIT is good enough.

foogod:
Hmm, for some reason I didn't see this update until just now..  Sorry about the delay!

PR #110, opened 2016-03-04
Hardware SPI driver

UncleRus:
Supports both SPI buses, frequency up to 80MHz, etc.

### DIFF
--- a/core/esp_spi.c
+++ b/core/esp_spi.c
@@ -107,7 +107,7 @@ void spi_set_frequency_div(uint8_t bus, uint32_t divider)
 {
     uint32_t predivider = divider & 0xffff;
     uint32_t count = divider >> 16;
-    if (count > 1 && divider > 1)
+    if (count > 1 || divider > 1)
     {
         predivider = predivider > SPI_CLOCK_DIV_PRE_M + 1 ? SPI_CLOCK_DIV_PRE_M + 1 : predivider;
         count = count > SPI_CLOCK_COUNT_NUM_M + 1 ? SPI_CLOCK_COUNT_NUM_M + 1 : count;
--- a/core/esp_spi.c
+++ b/core/esp_spi.c
@@ -194,13 +194,13 @@ void spi_transfer(uint8_t bus, void *data, size_t len)
     spi_endianness_t e = spi_get_endianness(bus);
     spi_set_endianness(bus, SPI_LITTLE_ENDIAN);
 
-    size_t counts = len / _SPI_BUF_SIZE;
-    for (uint8_t i = 0; i < counts; i++)
+    size_t blocks = len / _SPI_BUF_SIZE;
+    for (uint8_t i = 0; i < blocks; i++)
         _spi_buf_transfer(bus, data + i * _SPI_BUF_SIZE, _SPI_BUF_SIZE);
 
     uint8_t tail = len % _SPI_BUF_SIZE;
     if (tail)
-        _spi_buf_transfer(bus, data + counts * _SPI_BUF_SIZE, tail);
+        _spi_buf_transfer(bus, data + blocks * _SPI_BUF_SIZE, tail);
 
     spi_set_endianness(bus, e);
 }
--- a/core/esp_spi.c
+++ b/core/esp_spi.c
@@ -72,17 +72,30 @@ bool spi_init(uint8_t bus, spi_mode_t mode, uint32_t freq_divider, bool msb, spi
 
 void spi_set_mode(uint8_t bus, spi_mode_t mode)
 {
+    bool cpha = (uint8_t)mode & 1;
+    bool cpol = (uint8_t)mode & 2;
+    if (cpol)
+        cpha = !cpha;  // CPHA must be inverted when CPOL = 1, I have no idea why
+
     // CPHA
-    if ((uint8_t)mode & 1)
+    if (cpha)
         SPI(bus).USER0 |= SPI_USER0_CLOCK_OUT_EDGE;
     else
         SPI(bus).USER0 &= ~SPI_USER0_CLOCK_OUT_EDGE;
 
     // CPOL - see http://bbs.espressif.com/viewtopic.php?t=342#p5384
-    if ((uint8_t)mode & 2)
+    if (cpol)
         SPI(bus).PIN |= SPI_PIN_IDLE_EDGE;
     else
         SPI(bus).PIN &= ~SPI_PIN_IDLE_EDGE;
+}
+
+spi_mode_t spi_get_mode(uint8_t bus)
+{
+    uint8_t cpha = SPI(bus).USER0 & SPI_USER0_CLOCK_OUT_EDGE ? 1 : 0;
+    uint8_t cpol = SPI(bus).PIN & SPI_PIN_IDLE_EDGE ? 2 : 0;
+
+    return (spi_mode_t)(cpol | (cpol ? 1 - cpha : cpha)); // see spi_set_mode
 }
 
 void spi_set_msb(uint8_t bus, bool msb)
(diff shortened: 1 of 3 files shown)


foogod:
This looks really cool.. I'll have to get a bit of time to read over it in detail, but this could be pretty handy, especially since I've been considering wanting to reimplement some of the sdk_spi_flash_* routines, but was not looking forward to all the low-level stuff that would require..

core/include/esp/spi.h
```diff
@@ -0,0 +1,195 @@
...
+ * \brief Transfer dword over SPI
+ * \param bus Bus ID: 0 - system, 1 - user
+ * \param data dword to send
+ * \return Received dword
+ */
+uint32_t spi_transfer_32(uint8_t bus, uint32_t data);
+/**
+ * \brief Transfer buffer over SPI
+ * \param bus Bus ID: 0 - system, 1 - user
+ * \param data Data to send. Buffer contents will be replaced with received data
```
projectgus: Most SPI APIs I've used split this into an output buffer pointer and an optional input buffer pointer.

This lets you do things like send repeating patterns without having to memcpy() each time, and constant initialisation patterns can be declared const (which doesn't save RAM on esp-open-rtos at the moment, but it will soon.)

Would you consider changing the API to something like
```
void spi_transfer(uint8_t bus, const void *out_data, void *in_data, size_t len);
```

Where either out_data or in_data can optionally be null?

projectgus:
+1 this looks really cool, thanks very much for contributing it.

I left one comment inline, but I'm happy for @foogod to merge this once he's had time to review it in detail.

UncleRus:
You're right, separate send/receive buffers is much better.

core/esp_spi.c
```diff
@@ -0,0 +1,212 @@
...
+        SPI(bus).USER0 |= (SPI_USER0_WR_BYTE_ORDER | SPI_USER0_RD_BYTE_ORDER);
+    else
+        SPI(bus).USER0 &= ~(SPI_USER0_WR_BYTE_ORDER | SPI_USER0_RD_BYTE_ORDER);
+}
+
+void spi_set_frequency_div(uint8_t bus, uint32_t divider)
+{
+    uint32_t predivider = divider & 0xffff;
+    uint32_t count = divider >> 16;
+    if (count > 1 || divider > 1)
```
foogod: (minor) The `count > 1 ||` seems redundant here..

core/esp_spi.c
```diff
@@ -0,0 +1,212 @@
...
+    {
+        predivider = predivider > SPI_CLOCK_DIV_PRE_M + 1 ? SPI_CLOCK_DIV_PRE_M + 1 : predivider;
+        count = count > SPI_CLOCK_COUNT_NUM_M + 1 ? SPI_CLOCK_COUNT_NUM_M + 1 : count;
+        IOMUX.CONF &= ~(bus == 0 ? IOMUX_CONF_SPI0_CLOCK_EQU_SYS_CLOCK : IOMUX_CONF_SPI1_CLOCK_EQU_SYS_CLOCK);
+        SPI(bus).CLOCK = (((predivider - 1) & SPI_CLOCK_DIV_PRE_M) << SPI_CLOCK_DIV_PRE_S) |
+            (((count - 1)     & SPI_CLOCK_COUNT_NUM_M)  << SPI_CLOCK_COUNT_NUM_S) |
+            (((count / 2 - 1) & SPI_CLOCK_COUNT_HIGH_M) << SPI_CLOCK_COUNT_HIGH_S) |
+            (((count - 1)     & SPI_CLOCK_COUNT_LOW_M)  << SPI_CLOCK_COUNT_LOW_S);
```
foogod: common_macros.h actually has several macros to make working with the *_M / *_S stuff easier (and more readable).  You might want to take a look at `VAL2FIELD` for this stuff..

(Also, not a big deal, but it occurs to me it might also be a little cleaner if you just decremented `count` and `predivider` first, so you don't need all the `+ 1` and `- 1`s everywhere..)

core/esp_spi.c
```diff
@@ -0,0 +1,212 @@
...
+    }
+}
+
+inline static void _set_size(uint8_t bus, uint8_t bytes)
+{
+    uint16_t bits = ((uint16_t)bytes << 3) - 1;
+    const uint32_t mask = ~((SPI_USER1_MOSI_BITLEN_M << SPI_USER1_MOSI_BITLEN_S) |
+        (SPI_USER1_MISO_BITLEN_M << SPI_USER1_MISO_BITLEN_S));
+    SPI(bus).USER1 = (SPI(bus).USER1 & mask) | (bits << SPI_USER1_MOSI_BITLEN_S) |
+        (bits << SPI_USER1_MISO_BITLEN_S);
```
foogod: Likewise, this looks like a perfect case for the `SET_FIELD` macro..

core/esp_spi.c
```diff
@@ -0,0 +1,212 @@
...
+    return _spi_single_transfer(bus, data, sizeof(data));
+}
+
+void spi_transfer(uint8_t bus, const void *out_data, void *in_data, size_t len)
+{
+    if (!out_data || !len) return;
+
+    _wait(bus);
+    spi_endianness_t e = spi_get_endianness(bus);
+    spi_set_endianness(bus, SPI_LITTLE_ENDIAN);
```
foogod: I would much prefer to see this error out (or, even better, fall back to a less-efficient approach) if it's set to `SPI_BIG_ENDIAN` rather than just silently sending data in/out in the wrong order (especially since this behavior isn't even documented!)

foogod:
Your comment says it works with either SPI bus, but I assume this hasn't actually been tested with SPI0 (i.e. to read/write to the internal flash chip), because I'm pretty sure that won't work without putting most of these routines in IRAM instead of IROM..

(not a blocker for merging, but we should probably note that even though it appears to support SPI0, it probably won't actually work on a typical ESP8266 module without some additional work on this driver..)

UncleRus:
I think switching endianness silently when we sending buffer is a good option because in this case big endianness does not make sense at all. For example it sends "32107654" instead of "01234567" when big endianness flag is set.

foogod:
Why doesn't that make sense?  It makes perfect sense to me (and would be perfectly correct) if I'm trying to send a buffer of multiple 32-bit words to an SPI device that speaks big-endian.

I could see that it might be useful to be able to specify the "word size" for buffer transfers, so that the user can specify 1 (no endian swapping), 2 (swap 16-bit values) or 4 (swap 32-bit values) when sending/receiving a buffer in big-endian mode..  At the very least, if sending a buffer is always going to be treated as "byte data only (endianness ignored)", then that needs to be documented.

foogod:
I also just realized I'm not really seeing any error checking anywhere in this code.  What happens if an SPI bus or device does not respond as expected?  How does an application detect IO errors, etc?

(I haven't checked to see what sort of facilities the SPI controller has for this, but I assume it must have something?)

UncleRus:
> I haven't checked to see what sort of facilities the SPI controller has for this, but I assume it must have something?

As far I know it doesn't.

UncleRus:
> it might be useful to be able to specify the "word size" for buffer transfers, so that the user can specify 1 (no endian swapping), 2 (swap 16-bit values) or 4 (swap 32-bit values) when sending/receiving a buffer in big-endian mode

Good idea, thank you!

UncleRus:
@foogod Please  review.

foogod:
Sorry for the delay.. been out sick for the past few days..

This looks pretty thorough!  I have to admit that some of it is beyond my understanding of the hardware (and I don't have a lot in the way of SPI peripherals to test with myself) to really know whether it's correct, so I'll just have to trust you that it all works, I guess.. :)  I hadn't thought of using the SPI controller's buffer window as a scratch area to do the byte-swapping/etc in..  that's rather clever.

I don't have any big issues with this..  @projectgus any comments before merging this in?

projectgus:
Looks great to me... merged! Thanks @UncleRus.

PR #113, opened 2016-03-09
howmyssl.com's root certificate has been updated

kzys:
Now the site is using Let's Encrypt's root certificate.


projectgus:
Thanks. Yay Let's Encrypt!

PR #114, opened 2016-03-09
Added sntp support and example

doragasu:
Added SNTP support, using [lwIP SNTP contribution](http://cvs.savannah.gnu.org/viewvc/contrib/apps/sntp/?root=lwip), plus the code needed to set and read the RTC.

I suppose I'll have to do a lot of code style changes (like using /* */ for comments), but I'm sending right now the PR to get feedback.


projectgus:
This looks good at a glance, thanks for submitting it.

I know you mentioned something about trying to integrate with newlib/libc timezone support. Have you also tried implementing `_time_r` and `_gettimeofday_r` syscalls?

If you add these functions to your sntp_fun.c file then it should be possible to use libc `time()` and `gettimeofday()`, etc. calls.

The prototypes for those syscalls can be found here:
https://github.com/projectgus/newlib-xtensa/blob/xtensa-2_2_0-lock/newlib/libc/include/reent.h#L155

foogod:
Note that if you're going to implement the *_r versions, you need to make sure your code is thread-safe.  It is alternately possible (and I suspect much easier in this case) to just implement the non-_r versions, which as I understand, FreeRTOS will then by default automatically call wrapped appropriately to make sure you don't have to deal with multithreading issues..

(I actually just noticed recently that I think some of the file-IO stuff we've implemented with *_r variants is not, in reality, actually threadsafe, and so should not have been implemented as *_r as well.. I was meaning to go back and double-check this and possibly open up a bug for it, but then didn't follow through)

projectgus:
> Note that if you're going to implement the *_r versions, you need to make sure your code is thread-safe. 

I'm fairly sure newlib's locking mechanism (which we employ) provides thread safety as required at the libc level, so we don't generally need to add locking primitives to the syscall implementations themselves. For example, malloc locking means that only one task can call into sbrk_r() at a time, and file descriptor locking means that only one task will call read()/write() with the same file descriptor at a given time.

The primary reason newlib requires you to implement *_r instead of the other style is to set errno suitably for the calling context.

I could be wrong though, please open another issue to discuss this if you think it needs fixing.

doragasu:
I suspected something like _time_r() and _gettimeofday_r() syscalls would exist, but I'm not a newlib expert and I couldn't find documentation searching on the net. I'll have a look and try using them.

Thanks for the feedback and tips!

doragasu:
Had a quick look to syscalls. I'll try implementing gettimeofday_r(), but I cannot find _time_r(). There's a _times_r() syscall, but it looks like it's for process times measurements.

It also looks like there isn't a settimeofday() syscall, but I might try renaming the function I coded to set the time to settimeofday() for clarity.

doragasu:
Who is supposed to call `_gettimeofday_r()`? I wrote a dirty test implementation, but I cannot find a newlib function using it, to test the interaction. I'm starting to suspect that maybe no newlib functions use `_gettimeofday_r()`...

doragasu:
Did some code cleanup. Also implemented `time()` and modified some routines to be a bit nearer to the POSIX standard implementation. The main difference now is that `tz_dsttime` from `struct timezone` doesn't work the same.

When set to nonzero on POSIX implementation, daylight savings will be applied according to `tzset`. When set to 1 on this patch, an hour offset will be directly applied, until set to 0. So basically you have to manually manage when daylight savings are applied. Doing it correctly would require implementing `tzset` and friends, since it doesn't look like newlib provides this functionality. And I think doing it might be overkill.

projectgus:
Hi,

Thanks for all this work!

You should be able to use the `time()` from newlib, as it's available and implemented in terms of the `gettimeofday_r()` syscall:
https://github.com/projectgus/newlib-xtensa/blob/xtensa-2_2_0-lock/newlib/libc/time/time.c#L40

... it should be as simple as `#include <time.h>` and then calling it. Let me know if that doesn't work though, maybe some pieces are missing still.

Timezone support on newlib seems... interesting. You can call `tzset()` which will use the `getenv_r` syscall (unimplemented for us so far) to read a TZ environment variable. There's a discussion on the mailing list about adding a key/value store to esp-open-rtos, so it'd be interesting to some day extend that to implementing "environment variables" and you can set the timezone that way. But I don't think it's useful right now.

`tzset()` sets 3 global variables as described here:
https://github.com/projectgus/newlib-xtensa/blob/xtensa-2_2_0-lock/newlib/libc/time/tzset.c#L21
... including `_timezone` and `_daylight`, but it also references a global tzinfo strcture `_gettzinfo()`.

A mixture of these seem to be used by other local time implementations, for example `localtime()` uses both. `strftime()` uses `_gettzinfo()` only.

I'm not suggesting you necessarily need to go down the rabbit hole of supporting all the newlib libc timezone stuff though, it seems pretty confusing! Although I think it'd be good use the libc time() rather than reimplementing it, if we can. Even better if we can use the same shared timezone info data structures that the libc uses, but only if you feel like it.

doragasu:
I'll have another look to it when I get some time. I already did some experiments manually setting `_tzset` and `_tz`, but couldn't get these settings to affect the output of `ctime()`. Other things I tried: calling `clock_gettime()` and `clock_getres()` after implementing `_gettimeofday_r()`. They both are implemented (I got unresolved externals).

projectgus:
Cool, let me know how you get on. I don't mind helping debug if you could use a second pair of eyes.

> I don't know the format if these files, and I doubt newlib will use it

newlib just expects TZ to be a string with a format specified by POSIX, there's format information at the link I posted above. In glibc tzset() supports either that string format or a filename pointing to a binary file.

projectgus:
PS I think you're correct about clock_gettime and friends, grepping the newlib source shows there's a Linux implementation but not a platform-agnostic one.

doragasu:
I've been pretty busy lately, but at last had some time to continue digging into this. I have removed my implementation of time() and I'm using the one from newlib. Couldn't test the code (I have not an ESP lying around right now), but it builds and it should work (I'll test it ASAP).

Now, what's the suggested route? Implementing `_getenv_r()` to return the configured timezone string when querying TZ env variable?

doragasu:
I have been pretty busy lately, and almost forgot about this. Is there still interest? How can we proceed?

projectgus:
I'm very interested. I haven't had time to test yet though, in your last comment you said you planned to - I'm guessing you haven't had time either?

doragasu:
I'm not 100% sure, because this was over a month ago, but I think I already tested it and worked. I'll try testing it again and report.

projectgus:
Gave this another review and a test, looks great! Thanks for submitting it, sorry about all the delays.

doragasu:
Great to read it has finally been merged! Sorry for not being able to pull some minutes to test it again.

PR #118, opened 2016-03-21
esp/gpio.h c++ linker compatibility

UncleRus:
Same as afd58bc1ffb1aa9cf2f030d05d7491906925c3e6


projectgus:
Good catch, thanks.

PR #121, opened 2016-04-05
DS18B20 API Improvements

foogod:
The original DS18B20 API was functional, but not ideal for working with multiple sensors on the same bus.  In addition to a bunch of misc cleanup (use of open-drain mode, removing unnecessary critical sections, etc), this update adds several new API functions for identifying and accessing DS18B20 devices in more flexible (and much faster!) ways.


projectgus:
This looks very good to me. I also like that it moves away from function wrapper macros and has more typedefs for known types.

@hetii, @darksv  - you both used the older DS18B20 API. Any comments regarding merging this?

hetii:
From my side all looks fine, especially when works faster ;)
BTW, thanks for info that new API is available.
Regards.

PR #122, opened 2016-04-05
Add WARNINGS_AS_ERRORS make flag

foogod:
Previous behavior was all warnings were treated as errors.  This is now
controllable via a make variable and defaults to off (but can be turned on in
local.mk for those who still want the old behavior)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,4 +58,4 @@ script:
   - cd ${TRAVIS_BUILD_DIR}
   # Remove ssid_config requirement for examples
   - sed -i "s%#warning%//#warning%" include/ssid_config.h
-  - make -C examples/ build-examples CROSS="ccache xtensa-lx106-elf-" V=1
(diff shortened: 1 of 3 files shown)


projectgus:
Looks good!

Only one thing, could you please add WARNINGS_AS_ERRORS=1 to the Travis make line as well (it's the last line in .travis.yml). You can check the Travis log here after you push the branch to verify it's building with -Werror)

foogod:
Ah, ok.. I wasn't actually sure where the Travis build config was located (since it was hidden in a dot-file)..  I've added that now.

I did also just realize that currently in ssid_config.h we have a `#warning` line to make folks update it with their own info, but under the new default config that won't stop the build anymore.  Do we want to change that to `#error` instead, or is it ok if we just emit a warning message and continue on (and assume people are going to actually read the output of the compilation if they don't get what they expect)?

kanflo:
Good point. I think it should be changed to ```#error``` to make it easy for newcomers to find where they need to put their wifi credentials to get the example programs to run. Pretty much in the same manner people are reminded to clone the submodules if LWIP is missing.

PR #124, opened 2016-04-07
Updated rboot to #75ca33b including the flash write bug.

kanflo:
I merged the latest rboot code (#75ca33b) with the "flash write bugfix" into EOR. Default rboot configuration is BOOT_BIG_FLASH and BOOT_RTC_ENABLED.

I have tested the ota_basic demo several times observing that the roms get overwritten and that the system works.

Side note. The rename of extras/rboot-ota/rboot-ota.c → extras/rboot-ota/rboot-integration.c was not my intention. Dunno where git got that idea from.


projectgus:
Hi @kanflo,

Sorry for the slow reply. This looks good to me, the build failure is unrelated and should be fixed on master. The OTA example builds for me locally, so I'm going to merge this.

Thanks!

Angus

kanflo:
Great, thanks!

PR #129, opened 2016-04-20
Added DHT library to extras and sample code to examples

jsuiker:
Just getting my feet wet with the ESP8266 and the FreeRTOS code you all have been building.  Thanks for all the work and please feel free to critique what I've done here.

~Jonathan

### DIFF
--- a/examples/dht_sensor/Makefile
+++ b/examples/dht_sensor/Makefile
@@ -1,0 +1,4 @@
+PROGRAM=dht_sensor
+EXTRA_COMPONENTS = extras/dht
+include ../../common.mk
+
--- a/examples/dht_sensor/dht_sensor.c
+++ b/examples/dht_sensor/dht_sensor.c
@@ -1,0 +1,46 @@
+/* 
+ *
+ * This sample code is in the public domain.
+ */
+#include "espressif/esp_common.h"
+#include "esp/uart.h"
+#include "FreeRTOS.h"
+#include "task.h"
+#include "dht.h"
+#include "esp8266.h"
+
+/* An example using the ubiquitous DHT** humidity sensors
+ * to read and print a new temperature and humidity measurement
+ * from a sensor attached to GPIO pin 4.
+ */
+int const dht_gpio = 4;
+
+void dhtMeasurementTask(void *pvParameters)
+{
+  int8_t temperature = 0;
+  int8_t humidity = 0;
+
+  // DHT sensors that come mounted on a PCB generally have
+  // pull-up resistors on the data pin.  It is recommended
+  // to provide an external pull-up resistor otherwise...
+  gpio_set_pullup(dht_gpio, false, false);
+
+  while(1) {
+
+    if (dht_fetch_data(dht_gpio, &humidity, &temperature)) {
+      printf("Humidity: %i%% Temp: %iC\n", humidity, temperature);
+    } else {
+      printf("Could not read data from sensor...");
+    }
+
+    // Three second delay...
+    vTaskDelay(3000 / portTICK_RATE_MS);
+  }
+}
+
+void user_init(void)
+{
+    uart_set_baud(0, 115200);
+    xTaskCreate(dhtMeasurementTask, (signed char *)"dhtMeasurementTask", 256, NULL, 2, NULL);
+}
+
--- a/extras/dht/dht.c
+++ b/extras/dht/dht.c
@@ -1,0 +1,144 @@
+/*
+ * Part of esp-open-rtos
+ * Copyright (C) 2016 Jonathan Hartsuiker (https://github.com/jsuiker)
+ * BSD Licensed as described in the file LICENSE
+ *
+ */
+
+#include "dht.h"
+#include "string.h"
+#include "task.h"
+#include "esp/gpio.h"
+
+#include <espressif/esp_misc.h> // sdk_os_delay_us
+
+#ifndef DEBUG_DHT
+#define DEBUG_DHT 0
+#endif
+
+#if DEBUG_DHT
+#define debug(fmt, ...) printf("%s" fmt "\n", "dht: ", ## __VA_ARGS__);
+#else
+#define debug(fmt, ...) /* (do nothing) */
+#endif
+
+/* 
+ *  Note:
+ *  A suitable pull-up resistor should be connected to the selected GPIO line
+ *
+ *  __           ______          _______                              ___________________________
+ *    \    A    /      \   C    /       \   DHT duration_data_low    /                           \
+ *     \_______/   B    \______/    D    \__________________________/   DHT duration_data_high    \__
+ *
+ *
+ *  Initializing communications with the DHT requires four 'phases' as follows:
+ * 
+ *  Phase A - MCU pulls signal low for at least 18000 us
+ *  Phase B - MCU allows signal to float back up and waits 20-40us for DHT to pull it low
+ *  Phase C - DHT pulls signal low for ~80us
+ *  Phase D - DHT lets signal float back up for ~80us
+ *  
+ *  After this, the DHT transmits its first bit by holding the signal low for 50us
+ *  and then letting it float back high for a period of time that depends on the data bit.
+ *  duration_data_high is shorter than 50us for a logic '0' and longer than 50us for logic '1'.
+ *
+ *  There are a total of 40 data bits trasnmitted sequentially. These bits are read into a byte array
+ *  of length 5.  The first and third bytes are humidity (%) and temperature (C), respectively.  Bytes 2 and 4
+ *  are zero-filled and the fifth is a checksum such that:
+ *
+ *  byte_5 == (byte_1 + byte_2 + byte_3 + btye_4) & 0xFF
+ *
+*/
+
+/*  
+ *  @pin                    the selected GPIO pin
+ *  @interval               how frequently the pin state is checked in microseconds
+ *  @timeout                maximum length of time to wait for the expected pin state
+ *  @expected_pin_state     high (true) or low (false) pin state
+ *  @counter                pointer to external uint8_t for tallying the duration waited for the pin state
+*/
+
+bool dht_await_pin_state(uint8_t pin, uint8_t interval, uint8_t timeout, bool expected_pin_state, uint8_t * counter) {
+
+    for (*counter = 0; *counter < timeout; *counter+=interval) {
+        if (gpio_read(pin) == expected_pin_state) return true;
+        sdk_os_delay_us(interval);
+    }
+
+    return false;
+}
+
+/*  
+ *  
+ *  
+ *  @pin                    the selected GPIO pin
+ *  @humidity               pointer to external int8_t to store resulting humidity value
+ *  @temperature            pointer to external int8_t to store resulting temperature value
+*/
+
+bool dht_fetch_data(int8_t pin, int8_t * humidity, int8_t * temperature) {
+    int8_t data[40] = {0};
+    int8_t result[5] = {0};
+    uint8_t i = 0;
+
+    uint8_t init_phase_duration = 0;
+    uint8_t duration_data_low = 0;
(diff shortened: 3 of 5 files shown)


examples/dht_sensor/FreeRTOSConfig.h
```diff
@@ -0,0 +1,15 @@
+/* Blink FreeRTOSConfig overrides.
```
projectgus: You can delete this config file in order to get the default FreeRTOSConfig for the example, you probably don't need to override anything here.

extras/dht/dht.c
```diff
@@ -0,0 +1,128 @@
+
```
projectgus: Please add a copyright/license notice here and on the header file.

Something like these 3 lines is fine:
https://github.com/SuperHouse/esp-open-rtos/blob/master/extras/dhcpserver/dhcpserver.c#L12

(With the Copyright line changed as suitable.)

projectgus:
Thanks @jsuiker ! This looks good to me apart fomr the couple of minor things I mentioned inline, although I'm going to ask @foogod for a second opinion as he's more involved with the sensor-library type stuff.

examples/dht_sensor/Makefile
```diff
@@ -0,0 +1,4 @@
+
+PROGRAM=dht_sensor
+COMPONENTS = FreeRTOS lwip core extras/dht
```
foogod: I've been meaning to go through and fix this in our existing examples (so it's completely understandable that you didn't know this) but instead of overriding `COMPONENTS` here, you should instead just set `EXTRA_COMPONENTS = extras/dht` (this will make it work properly even if we change the list of default core components later)

extras/dht/dht.c
```diff
@@ -0,0 +1,128 @@
+
+
+#include "dht.h"
+#include "string.h"
+#include "task.h"
+#include "esp/gpio.h"
+
+/* 
+ *  Note:
+ *  A suitable pull-up resistor should be connected to the selected GPOI line
```
foogod: (minor) typo here ("GPOI")

extras/dht/dht.h
```diff
@@ -0,0 +1,10 @@
+#ifndef __DTH_H__
+#define __DHT_H__
+
+#include <espressif/esp_misc.h> // sdk_os_delay_us
```
foogod: (nitpick) There's nothing in this header file that references `sdk_os_delay_us`, so there's no reason that `dht.h` should include `esp_misc.h`..  That should really go in `dht.c` instead.

extras/dht/dht.h
```diff
@@ -0,0 +1,10 @@
+#ifndef __DTH_H__
+#define __DHT_H__
+
+#include <espressif/esp_misc.h> // sdk_os_delay_us
+#include "FreeRTOS.h"
+
+bool waitForPinState(uint8_t pin, uint8_t interval, uint8_t timeout, bool expected_pin_sate, uint8_t * counter);
+bool readHumidityAndTemperature(int8_t pin, int8_t * humidity, int8_t * temperature, char ** status_message);
```
foogod: I would prefer these function names be prefixed with something like `dht_` to keep it clear what module they're part of and reduce potential namespace collision issues..  Also, we have generally been following the lowercase-with-underscores convention for function names, rather than camelCase..

extras/dht/dht.c
```diff
@@ -0,0 +1,128 @@
...
+    int8_t data[40] = {0};
+    int8_t result[5] = {0};
+    uint8_t i = 0;
+
+    uint8_t init_phase_duration = 0;
+    uint8_t duration_data_low = 0;
+    uint8_t duration_data_high = 0;
+
+    gpio_enable(pin, GPIO_OUT_OPEN_DRAIN);
+    gpio_set_pullup(pin, false, false);
```
foogod: You generally shouldn't call `gpio_set_pullup` from inside the driver.  The whole point of separating that function out from `gpio_enable` was to let the app-developer decide themselves whether the pullup should be set or not based on their hardware design, not the driver.

extras/dht/dht.c
```diff
@@ -0,0 +1,128 @@
...
+                }
+
+                if (result[4] == ((result[0] + result[1] + result[2] + result[3]) & 0xFF)) {
+                    // Data valid, checksum succeeded...
+                    *humidity = result[0];
+                    *temperature = result[2];
+                    *status_message = "Success...";
+                    return true;
+                } else {
+                    *status_message = "Checksum failed, invalid data received from sensor...";
```
foogod: Instead of returning a text string in `status_message`, it would be much better to return some sort of status code that can be easily interpreted by the caller to determine the cause of the issue.  Also, it may be useful to include debugging messages such as this if the app developer wants them, but since text messages can take up a fair bit of space (currently precious DRAM, too, though hopefully that will change at some point) they should be configurable at compile time so they can be omitted if code/data space is at a premium.

jsuiker: I changed over to a code rather than a string variable, but my C skills are not quite advanced enough to implement a compile-time toggle switch like what you're thinking.  Perhaps you could provide an example for me to emulate?

foogod: I wasn't really suggesting switching between returning a code vs strings, as that would get really ugly with C's static typing and make the API a pain to work with on the application side too.  I was mainly just suggesting you could add debugging `printf` statements in the driver code which could be enabled at compile time if the user wants to know more about what's going on.  What I usually do is something like the following (at the top of the code):

```c
#ifndef DEBUG_DHT
#define DEBUG_DHT 0
#endif

#if DEBUG_DHT
#define debug(fmt, ...) printf("%s" format "\n", "dht: ", ## __VA_ARGS__);
#else
#define debug(fmt, ...) /* (do nothing) */
#endif
```
..and then you can just use `debug("debug message!")`, etc, throughout your code..  if the project is built with `-DDEBUG_DHT=1`, the debug messages will be included and will be printed to stdout when it executes.  If not (the default), the debug statements and their string arguments, etc, won't even be included in the compiled output.

Actually, to be honest, in this case the more I think about it, the more I think you probably don't need to return even a status code, really.. you could just change all the `*status_message = "..."` lines to `debug("...")` instead.  I can't really come up with any scenarios where it would be useful/important for the caller to know whether the failure was in (for example) phase B vs phase C of the transaction..  that's probably only useful for diagnostic/debugging level stuff..

extras/dht/dht.c
```diff
@@ -0,0 +1,141 @@
...
+    int8_t data[40] = {0};
+    int8_t result[5] = {0};
+    uint8_t i = 0;
+
+    uint8_t init_phase_duration = 0;
+    uint8_t duration_data_low = 0;
+    uint8_t duration_data_high = 0;
+
+    gpio_enable(pin, GPIO_OUT_OPEN_DRAIN);
+    gpio_set_pullup(pin, internal_pullup, false);
```
foogod: This is better, but as I mentioned the general intention with these routines was that the _user_ should call `gpio_set_pullup` themselves, not the driver code.  Making it configurable does at least give the user the desired flexibility, but it unnecessarily complicates your API.  If you just remove this line, then the user can call `gpio_set_pullup` themselves for that pin during their app initialization and your driver code doesn't even have to care whether the internal pullup is being used or not..

extras/dht/dht.c
```diff
@@ -0,0 +1,141 @@
...
+ *  @pin                    the selected GPIO pin
+ *  @internal_pullup        boolean to indicate whether the internal pullup resistor should be used
+ *  @humidity               pointer to external int8_t to store resulting humidity value
+ *  @temperature            pointer to external int8_t to store resulting temperature value
+ *  @status_code            pointer to external uint8_t to store status code corresponding to the following messages:
+ *                              0: "Success!"
+ *                              1: "Checksum failed, invalid data received from sensor..."
+ *                              2: "DHT did not initialize correctly, problem in phase 'B'..."
+ *                              3: "DHT did not initialize correctly, problem in phase 'C'..."
+ *                              4: "DHT did not initialize correctly, problem in phase 'D'..."
```
foogod: If these constants are part of the API, it would be better to give them symbolic names in `dht.h` (using either `#define` or by using an `enum` datatype).  As I mentioned in another comment, though, the more I think about this, the more I think it's probably not necessary to return this info to the caller at all, and these could just be debugging `printf` calls, etc, instead (which would simplify the API too)..  Not a huge deal either way, though..

foogod:
The updated version looks much better!  I left a couple of inline suggestions, but really neither of those is a blocker for merging, in my opinion (mainly they would just make the API a bit cleaner/easier to use)..

extras/dht/dht.c
jsuiker: Nice :)  I like this solution

extras/dht/dht.c
jsuiker: Ok, so it doesn't matter that the pullup state is set before calling (and re-calling) gpio_enable?

foogod: Right..  `gpio_enable` is specifically coded not to mess with the pullup state, so `gpio_set_pullup` can be called once early on, and then no matter what's done later with `gpio_enable`, it will "do the right thing"..  We did it that way specifically because we've had the same issue in several other drivers as well, and it seemed silly for each driver to have to have its own code to deal with letting the user configure pullup state, so now they don't have to.

examples/dht_sensor/dht_sensor.c
```diff
@@ -0,0 +1,39 @@
...
+#include "FreeRTOS.h"
+#include "task.h"
+#include "dht.h"
+#include "esp8266.h"
+
+/* An example using the ubiquitous DHT** humidity sensors
+ * to print out a new temperature and humidity measurement
+ * every three seconds.
+ */
+int const gpio4 = 4;
```
foogod: It is probably a good idea to make the GPIO used be easily configurable, but if you're going to do this, it should probably be named something more general (like maybe `dht_gpio`)..  Currently if somebody decides to change this then they'll end up with code that says (for example) `gpio4 = 5`, which is just all kinds of confusing :)

jsuiker: ya ya ya

examples/dht_sensor/dht_sensor.c
```diff
@@ -0,0 +1,39 @@
...
+
+void dhtMeasurementTask(void *pvParameters)
+{
+  int8_t temperature = 0;
+  int8_t humidity = 0;
+
+  gpio_set_pullup(gpio4, false, false);
```
foogod: I'd be curious to know whether the internal pullups are sufficient for the DHT sensor (at least when connecting it over short wires directly to the pin, etc).. I suspect they are, in which case it might be more convenient to have the example default to having the internal pullup enabled instead..  Alternately, it might be good to put a comment in here pointing out that this assumes you have an external pullup already attached (and if you don't you'll want to change that setting)..  Not a big deal either way, though..

jsuiker: Both of the sensors I have came mounted on a PCB with a pull-up resistor integrated.  I know you can get them without, but haven't got any on hand.

foogod:
I'm pretty good with this at this point..  @projectgus, do you have any remaining issues?

examples/dht_sensor/dht_sensor.c
foogod: Ah, ok.. I don't actually have any, but was actually thinking about playing around with those, so maybe I'll get some and do some more experimentation of my own :)  This is definitely fine for now, though..

projectgus:
Looks great to me! A couple of really minor housekeeping things:

- There's no newlines at the ends of the files.

- Please [squash your commits as described at this link](http://eli.thegreenplace.net/2014/02/19/squashing-github-pull-requests-into-a-single-commit/), and re-push so they appear as a single commit.

jsuiker:
Should be ready, lemme know!

projectgus:
Terrific, thanks!

Issue #123, opened 2016-04-06
Fix lwip timebase - select(), maybe other things


projectgus:
(This is a good candidate for a "first bug" if someone is interested in contributing but unsure where to start.)

Bug was noted in #91 when using select() - the timeout argument to select appears to take RTOS ticks instead of seconds.

The implementation is in lwip/sys_arch.c and the bug may be as simple as the sys_now() function returning tick count instead of tick count.

However, before applying a fix someone should verify this and the other time-related functions in sys_arch.c.

After apply the fix, check the codebase for uses of select() or other time-sensitive LWIP functions.

RLuckom:
Happy to start working on this. Sorry for the long note--I'm not super familiar with this stuff and I'm being extra careful to explain myself to make sure I understand correctly.

I noticed that http://git.savannah.gnu.org/cgit/lwip.git/tree/doc/sys_arch.txt#n202 explicitly states that sys_now() is supposed to return miliseconds.

Looking at the rest of that file, there are the following time-related functions:

`sys_arch_sem_wait` is also supposed to take a timeout param in milliseconds. sys_arch's implementation seems to expect that `timeout` will be in ticks, not ms, and divides `timeout` by `portTICK_RATE_MS` before passing it to `xSemaphoreTake`. This is probably the source of the `select` bug.

`sys_arch_mbox_fetch` is the same, supposed to take timeout in ms, but divides timeout by `portTICK_RATE_MS` before passing to `xQueueReceive`.

I didn't see any other functions in sys_arch that dealt with timeouts.

Looking closer at uses of `sys_now`, I found:

1. lwip/lwip/config/lwipopts.h defines `static inline uint32_t sys_now(void)` as `NOW()/(TIMER_CLK_FREQ/1000);`. I don't see where the `NOW()` macro is defined, but assuming it is the same as tick count, that is returning ms. This function has a different type than the one in sys.arch.c, and I believe that in most places in lwip, the function from sys_arch.c, not this one, is being used.
2. lwip/lwip/src/core/timers.c uses `sys_now` for most or all of its time and timediff needs. It uses `sys_now` to initialise `timeouts_last_time` (and reset it in `restart_timeouts`), and in `sys_check_timeouts` it is used to calculate the diff when determining if timeouts have expired.
[...]

PR #134, opened 2016-04-28
correct timebase in sys_arch.c

RLuckom:
sys_now() now returns ms.

### DIFF
--- a/lwip/sys_arch.c
+++ b/lwip/sys_arch.c
@@ -210,7 +210,7 @@ unsigned long ulReturn;
 	{
 		configASSERT( xInsideISR == ( portBASE_TYPE ) 0 );
 
-		if( pdTRUE == xQueueReceive( *pxMailBox, &( *ppvBuffer ), ulTimeOut/ portTICK_RATE_MS ) )
+		if( pdTRUE == xQueueReceive( *pxMailBox, &( *ppvBuffer ), ulTimeOut / portTICK_RATE_MS ) )
 		{
 			xEndTime = xTaskGetTickCount();
 			xElapsed = ( xEndTime - xStartTime ) * portTICK_RATE_MS;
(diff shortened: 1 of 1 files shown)


lwip/sys_arch.c
```diff
@@ -210,7 +210,7 @@ unsigned long ulReturn;
 	{
 		configASSERT( xInsideISR == ( portBASE_TYPE ) 0 );
 
-		if( pdTRUE == xQueueReceive( *pxMailBox, &( *ppvBuffer ), ulTimeOut/ portTICK_RATE_MS ) )
+		if( pdTRUE == xQueueReceive( *pxMailBox, &( *ppvBuffer ), ulTimeOut ) )
```
RLuckom: ulTimeout is supposed to be in ms. If so, it should not be divided by portTICK_RATE_MS

RLuckom:
See https://github.com/SuperHouse/esp-open-rtos/issues/123 for the long-winded rationale.

lwip/sys_arch.c
RLuckom: Sorry, I think I misunderstood portTICK_RATE_MS. This was correct before.

projectgus:
Welcome to the project, thanks for looking into this so thoroughly!

Looks good! And you don't think we've accidentally depended on the existing behaviour in any other places?

The one place I know about is here:
https://github.com/SuperHouse/esp-open-rtos/blob/master/extras/paho_mqtt_c/MQTTESP8266.c#L73

... but that might be the only one?

RLuckom:
Thanks for pointing that out.

I haven't actually found a reason for the select() bug referenced in issue 134. As far as I can tell, select calls sys_arch_sem_wait (which handles ms correctly), and sys_arch_sem_wait converts ms to ticks and then calls into xSemaphoreTake, which is defined to be an invocation of xQueueGenericReceive (and is supposed to take a timeout in ticks).

Because the mqtt extra is calling select, and I haven't found a connection between this change and select, I'm not sure I should add that change. However I'll keep looking into it.

projectgus:
Thanks for the detailed investigation and explanation @RLuckom . Sorry I didn't get back to you earlier.

As I commented on the parent issue, I agree with you that there isn't actually a bug with select(). However I think this is also a legitimate bug fix, so I've cherry-picked it into master as 0734fa4166.

Thanks again!

PR #135, opened 2016-05-03
PHY management features

projectgus:
I've started digging into the PHY management, both in the SDK and a bit at the hardware register.

These two commits have what I've figured out so far - including an esp-open-rtos implementation for configuration the default SDK PHY settings.

### DIFF
--- a/core/include/esp/phy.h
+++ b/core/include/esp/phy.h
@@ -1,0 +1,58 @@
+/** esp/phy.h
+ *
+ * PHY hardware management functions.
+ *
+ * These are not enough to configure the ESP8266 PHY by themselves
+ * (yet), but they provide utilities to modify the configuration set
+ * up via the SDK.
+ *
+ * Functions implemented here deal directly with the hardware, not the
+ * SDK software layers.
+ *
+ * Part of esp-open-rtos
+ * Copyright (C) 2016 ChefSteps, Inc
+ * BSD Licensed as described in the file LICENSE
+ */
+#ifndef _ESP_PHY_H
+#define _ESP_PHY_H
+
+#include <esp/types.h>
+#include <common_macros.h>
+
+#include <stdint.h>
+
+typedef enum {
+    BT_COEXIST_NONE,
+    BT_COEXIST_USE_BT_ACTIVE,
+    BT_COEXIST_USE_BT_ACTIVE_PRIORITY,
+} bt_coexist_mode_t;
+
+/** Override the Bluetooth Coexistence BT_ACTIVE pin setting
+    taken from the phy_info structure.
+
+    This enables other pins to be used for Bluetooth Coexistence
+    signals (rather than just the two provided for by phy_info). (Note
+    that not that not all pins are confirmed to work, GPIO 0 is
+    confirmed not usable as the SDK configures it as the WLAN_ACTIVE
+    output - even if you change the pin mode the SDK will change it
+    back.)
+
+    To change pins and have coexistence work successfully the BT
+    coexistence feature must be enabled in the phy_info configuration
+    (get_default_phy_info()). You can then modify the initial
+    configuration by calling this function from your own user_init
+    function.
+
+    (Eventually we should be able to support enough PHY registers
+    to enable coexistence without SDK support at all, but not yet.)
+
+    This function will enable bt_active_pin & bt_priority_as GPIO
+    inputs, according to the mode parameter.
+
+    Remember that the default Bluetooth Coexistence pins will be
+    configured as GPIOs by the SDK already, so you may want to
+    reconfigure/re-iomux them after calling this function.
+*/
+void bt_coexist_configure(bt_coexist_mode_t mode, uint8_t bt_active_pin, uint8_t bt_priority_pin);
+
+#endif
--- a/include/espressif/phy_info.h
+++ b/include/espressif/phy_info.h
@@ -1,0 +1,482 @@
+/** Internal Espressif SDK "PHY info" data structure
+
+   The data structure (sdk_phy_info_t) is used to configure the
+   ESP8266 PHY layer via the SDK. The fields here are not written
+   directly to hardware, the SDK code (mostly in libphy) parses this
+   structure and configures the hardware.
+
+   The structure loaded at reset time from a flash configuration
+   sector (see read_saved_phy_info()) (Espressif's SDK sources this
+   from a file "esp_init_data_default.bin"). If no valid structure is
+   found in the flash config sector then the SDK loads default values
+   (see get_default_phy_info()). It is possible to implement a custom
+   get_default_phy_info() to change the PHY default settings (see the
+   'version' field below).
+
+   @note It is possible that the SDK will quietly write a new
+   configuration sector to flash itself following internal
+   calibration, etc. However this does not seem to happen, you need to
+   flash it explicitly if you want it stored there.
+
+   @note Most of what is below is unconfirmed, except where a @note
+   says that it has been confirmed to work as expected. Please
+   consider submitting notes if you find behaviour here that works or
+   doesn't work as expected.
+
+   Information on the meaning/offset of fields comes from Espressif's
+   flash download tool, which uses an Excel spreadsheet (in the
+   init_data directory of the ZIP file) to configure and a Python
+   script to convert an esp_init_data_custom.bin file to flash:
+   http://bbs.espressif.com/viewtopic.php?f=5&t=433
+
+   Many fields remain undocumented (& disassembly of libphy suggests
+   that some documented fields supported undocumented values.)
+
+   A few additional notes about the phy_info fields can be found
+   in the ESP Arduino ESP8266 phy_init_data structure (however most of
+   that content is verbatim from Espressif's spreadsheet):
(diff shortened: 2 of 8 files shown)


include/espressif/phy_info.h
```diff
@@ -0,0 +1,456 @@
...
+#define BT_COEXIST_PROTOCOL_BT_ONLY 1
+
+/* Coexistence is enabled, Bluetooth has its own antenna */
+#define BT_COEXIST_PROTOCOL_FLAG_SEPARATE_ANT 2
+/* Coexistence is enabled, Bluetooth shares WiFi antenna */
+#define BT_COEXIST_PROTOCOL_FLAG_SHARE_ANT 4
+
+/* Coexistence is enabled, use only BT_ACTIVE signal */
+#define BT_COEXIST_PROTOCOL_FLAG_BT_ACTIVE_ONLY 0
+/* Coexistence is enabled, use both BT_ACTIVE and BT_PROTOCOL signals */
```
pfalcon: BT_PRIORITY perhaps?

include/espressif/phy_info.h
```diff
@@ -0,0 +1,456 @@
...
+#define CRYSTAL_SLEEP_OFF 0
+#define CRYSTAL_SLEEP_ON 1
+#define CRYSTAL_SLEEP_GPIO16 2
+#define CRYSTAL_SLEEP_GPIO2 3
+
+/* RF Stage 0 & 1 attenuation constants. Use for sdk_phy_info_t.lp_atten_stage01
+
+   @note These values have been tested and are confirmed to work as
+   expected - provided low_power_en is set then the signal is
+   attenuated as per this setting.
```
pfalcon: Cool, would be nice to include which vector network analyzer was used to measure attenuation.

include/espressif/phy_info.h
```diff
@@ -0,0 +1,456 @@
...
+
+/* Set this flag to enable frequency offset correction */
+#define FREQ_CORRECT_ENABLE BIT(0)
+
+/* Set = Baseband PLL frequency is 160MHz (can only apply +ve offset)
+ * Unset = Baseband PLL frequency is 168MHz (can apply +ve/-ve offset */
+#define FREQ_CORRECT_BB_160M BIT(1)
+
+/* Set = use force_freq_offset field to correct, Unset = automatically
+   measure &correct offset
```
pfalcon: Minor whitespace issue.

include/espressif/phy_info.h
```diff
@@ -0,0 +1,456 @@
...
+#define RF_CAL_MODE_SAVED_2 2
+/* Run full RF CAL routine. RF init takes approx 200ms. */
+#define RF_CAL_MODE_FULL 3
+
+/* Data structure that maps to the phy_info configuration block */
+typedef struct __attribute__((packed)) {
+    uint8_t _reserved00[0x05]; /* 0x00 - 0x04 */
+
+    /* This "version" field was set to 5 in the SDK phy_info,
+       and the original SDK startup code checks it is 5 and then loads
```
pfalcon: Someone might think that the structure then should be maned sdk_phy_info_v5_t, but I don't think we'll see other versions (hmm, other structures do change between versions of non-rtos sdk).

include/espressif/phy_info.h
```diff
@@ -0,0 +1,456 @@
...
+    */
+    uint8_t target_power[6]; /* 0x22 - 0x27 */
+
+    /* Maps the 8 MCS (modulation & coding schemes) for 802.11g
+       to a target_power level index (0-5).
+
+       MCS data rates are 6, 9, 12, 18, 24, 36, 48, 54Mbps
+       and I believe this is the same order used here.
+
+	   SDK default mapping (from MCS ie index in this array to
```
pfalcon: What's "ie"?

include/espressif/phy_info.h
```diff
@@ -0,0 +1,456 @@
...
+	   target_power ie index in the target_power array) is:
+
+	   MCS 0 -  6Mbps - target_power 0 - 20.5dBm
+	   MCS 1 -  9Mbps - target_power 0 - 20.5dBm
+	   MCS 2 - 12Mbps - target_power 1 - 19.5dBm
+	   MCS 3 - 18Mbps - target_power 1 - 19.5dBm
+	   MCS 4 - 24Mbps - target_power 3 - 18.5dBm
+	   MCS 5 - 36Mbps - target_power 4 - 17dBm
+	   MCS 6 - 48Mbps - target_power 5 - 16dBm
```
pfalcon: What about MCS 7 aka 54Mbps

include/espressif/phy_info.h
```diff
@@ -0,0 +1,456 @@
...
+    */
+    uint8_t target_power_index_mcs[8]; /* 0x28 - 0x2f */
+
+    /* One of CRYSTAL_FREQ_40M / CRYSTAL_FREQ_26M / CRYSTAL_FREQ_24M
+
+	   The crystal configured here is the input to the PLL setting
+	   calulcations which are used to derive the CPU & APB peripheral
```
pfalcon: Typo (calul...)

include/espressif/phy_info.h
```diff
@@ -0,0 +1,456 @@
...
+
+    /* Set the voltage of PA_VDD, which appears to be an internal analog
+       reference voltage(?)
+
+       This field is called vdd33_const in the Arduino phy fields,
+       and relates to usage of the TOUT pin (ADC pin).
+
+       Set to PA_VDD_MEASURE_VCC (0xFF) and leave TOUT (ADC) pin floating in order to use the ADC to measure the 3.3V input voltage.
+
+       Set to value in the range 18-36 (1.8V to 3.6V) to set a reference voltage(?) when using TOUT pin as an ADC input. I think this is the reference voltage used to scale the 0-1V which is allowed on the pin, in order to get an accurate reading. So it should match the system VCC voltage... I think!
```
pfalcon: long line

pfalcon:
Nice work. As a random gossip, I wanted to post announcement of https://github.com/pfalcon/esp-open-headers/ to RE list, but figured there's little to post about, as I don't want to keep doing it manually. My biggest interest is in esf_bug module https://github.com/pfalcon/esp-open-headers/blob/master/pp/esf_buf.h , which is the biggest source of esp8266 instability (as the exhaustion of send/receive buffers leads to deadlocks).

include/espressif/phy_info.h
projectgus: No VNA was used, these are relative power measurements so I confirmed with a mixture of monitor mode on a USB dongle and an old Spectrum Analyzer (that's no longer in absolute calibration, but that doesn't really apply here).

include/espressif/phy_info.h
projectgus: From the versions I looked at, it seems like Espressif added some new (undocumented) fields towards the end but didn't change a lot else. I couldn't see any examples of them changing this number though. :/ It may not even be a version, it's just the field that the RTOS SDK 0.9.9 happens to check for on startup.

projectgus:
Thanks for the review comments @pfalcon. Pushed new version just now.

I also realised the MCS index table was based on old information (first version of the structure was just based on the Arduino header file), so I updated it.

include/espressif/phy_info.h
pfalcon: Maybe it's just me, but I'd say it would be useful to know these details... I'm not sure where they would belong, but definitely could fit into commit message. By "usb dongle" you mean normal wifi dongle, and looking at RSSI it provides?

Well, I guess it's common comment about someone's doing RE - it would be nice to know as much as possible in case someone else will continue that work or will need to re-verify some moments. I of course know how hard is to record all those details... (So the point of comments is ... people looking, it's not useless ;-) ).

projectgus: OK, makes sense. I added a bit more technical detail. I think more detail than what I just added is probably not meaningful unless I'd been able to test in an EMC test environment with a full calibrated signal path.

projectgus:
>  wanted to post announcement of https://github.com/pfalcon/esp-open-headers/ to RE list, but figured there's little to post about

Interesting work, thanks for sharing that. It'll be interesting to see how the data structures in rtos sdk libpp compare to those ones.

core/include/esp/phy.h
```diff
@@ -0,0 +1,58 @@
...
+#include <esp/types.h>
+#include <common_macros.h>
+
+#include <stdint.h>
+
+typedef enum {
+    BT_COEX_NONE,
+    BT_COEX_USE_BT_ACTIVE,
+    BT_COEX_USE_BT_ACTIVE_PRIORITY,
+} bt_coex_mode_t;
```
foogod: "BT_COEX" smells like unnecessary contraction/abbreviation to me.  I can accept "BT" for "bluetooth" because it is a fairly standard abbreviation, but "coexist" is only three more characters than "coex" and is a lot more obvious what it means, particularly since that's what you've called it elsewhere already anyway (at the very least, it should be consistently named everywhere)..

foogod:
This looks really good.. aside from the "coex" vs "coexist" bit, everything else looks good to me!

projectgus:
Thanks @foogod. Changed coex to coexist everywhere.

PR #136, opened 2016-05-03
OTA improvements, always enable OTA mode

projectgus:
This is a grouping of OTA improvements to do with making the OTA support more usable and more secure.

These are some of the steps towards the discussed goal of making esp-open-rtos use OTA mode for all images. At the moment it just integrates rboot directly from @raburton, we may find in time that we diverge enough that we need to fork our own bootloader but for now it works OK after the recent excellent upstream rebasing job by @kanflo .

Comments from any interested parties, particularly those involved in the previous discussion about moving to OTA mode (@kanflo, @foogod) would be very welcome. :)


raburton:
I guess I qualify as an interested party, in some sense, as the author of rBoot. From my point of view it would be a shame to fork rBoot and if there are features you want I'm willing to take suggestions for consideration upstream. I know you're rewritten much of the API side, which seems entirely reasonable, to fit in with your framework code but if you change the bootloader it's another thing for you to maintain in your project and breaks compatibility with other users. At the moment rBoot allows you to boot roms created with the plain original SDK, plain RTOS SDK, your RTOS SDK, Sming, nodemcu, etc. because it is just one bootloader. Each framework can use it's own customised api to interact with it, but at the core they remain compatible allowing dual booting two or more completely different roms from different frameworks on the same device.

BTW the rBoot defines can be set via variables in environment/Makefiles the so you don't need to create extra header files for them (extras/rboot-ota/rboot.h) or you can just use this to enable the rboot integration option and then add your overrides to rboot.h options in there.

bootloader/Makefile
```diff
@@ -0,0 +1,71 @@
...
+export RBOOT_BIG_FLASH    = 1
+
+export RBOOT_BUILD_BASE=$(abspath $(BUILD_DIR))
+
+# Default ESPTOOL params, all the same as when using normal esp-open-rtos makefiles
+ESPTOOL ?= esptool.py
+ESPPORT ?= /dev/ttyUSB0
+ESPBAUD ?= 115200
+
+FLASH_SIZE ?= 16
```
kanflo: Could we bump FLASH_SIZE to 32?

kanflo:
Good work! I agree with @raburton that we should not fork rboot. I have been thinking about a boot and OTA strategy for my EOR projects with the industry knowledge I have regarding these things.

The current scheme relies on functional OTA functionality built into our application and if a buggy application gets flashed we are bricked. My plan is to create a secondary boot loader that can handle remote upgrades and _that is never overwritten_, freeing the main applicaton from needing remote upgrade support. It only needs to be able to detect that a remote upgrade is available, record this fact in an manner recoverable by rboot and restart. Rboot will then start the extended boot loader ("xboot"?) that will handle the upgrade. The current rboot/OTA scheme leaves the application running while it is downloading a new version, something that will not happen using this new scheme. Unless the new version is downloaded into a scratch area that is. The plan also includes having "xboot" calculate the application hash on each boot and only launching if found intact.

projectgus:
Thanks for weighing in @raburton. I wasn't sure whether to ask you outright, as it's obviously a big ask for you to support yet another random framework. :). So I'm very happy to hear from you, I really appreciate it.

By way of context to the discussion, there's a couple of fairly sprawling discussions on the mailing list here:
https://groups.google.com/forum/#!searchin/esp-open-rtos/OTA/esp-open-rtos/_j7MffxhqsA/j0_Hw5ziBQAJ
and here:
https://groups.google.com/forum/#!searchin/esp-open-rtos/bootloader/esp-open-rtos/SnJeAkjBK8E/yiDx5z6gAQAJ

The core of which is:
* Moving esp-open-rtos to always build images in "OTA" mode
* Possibly augmenting/replacing Espressif's boot configuration modes (the 16-byte header) with something a bit more flexible regarding parameters

Depending on how far we get down that path, I can foresee we might have to discuss exactly which options make sense for upstreaming into rboot and which don't. But I'm happy to leave that discussion for later.

> BTW the rBoot defines can be set via variables in environment/Makefiles the so you don't need to create extra header files for them (extras/rboot-ota/rboot.h) or you can just use this to enable the rboot integration option and then add your overrides to rboot.h options in there.

OK, thanks for the tip. I had a look, and I think I can unify everything in a fairly clean way with a small tweak to the Makefile in rboot. I'll make the change and send you a PR to consider.

> Could we bump FLASH_SIZE to 32?

I don't really like any of this part of the Makefile, it'd be a lot better if we could inherit the esptool params from the top-level common.mk/local.mk - but I couldn't find a clean way to do that (maybe by splitting common.mk into parts, which might not be a bad refactor to do anyhow).

> The current scheme relies on functional OTA functionality built into our application and if a buggy application gets flashed we are bricked.

Yes. This PR's changes came out of some work I'm doing for a client's commercial project as well. For it we've allocated a "factory" slot 0 firmware that we never overwrite via OTA, and then 2 "user" slot firmwares that we do overwrite via OTA. All of the SHA256/image verification stuff is used to make sure that we never switch to a new firmware slot without first confirming that it's SHA256 matches.

And there's also a check that we never erase sectors in the "factory" firmware region (AFAIK there's no way to enable sector erase protection via ESP8266).

As an anti-brick emergency measure against buggy application code, I added support to rboot to use any GPIO as the "GPIO boot" mode trigger in order to pull a pin low and force the factory slot 0 to boot. (I'll send that as a PR to rboot in the next few days, as well.)

Using a fully-featured factory firmware as the "failsafe" for OTA lets us leverage all of the trust/security features of the main application (because we're using the same trusted data source for application stuff as well.)

All that said, the xboot idea seems neat, I look forward to seeing more about it. :)

kanflo:
> The core of which is:
> Moving esp-open-rtos to always build images in "OTA" mode

I guess a makefile selectable OTA slot for ```make flash``` will come in handy.

> And there's also a check that we never erase sectors in the "factory" firmware region

That is a nice idea.

> As an anti-brick emergency measure against buggy application code, I added support to rboot to use any GPIO as the "GPIO boot" mode trigger in order to pull a pin low and force the factory slot 0 to boot.

I thought about this too and will put it in "xboot" (leaving rboot untouched). Which gpio pin will be read from @foogod's parameter framework (as well as wifi credentials and so on). I have begun adding "master reset buttons" to my new designs for this purpose.

Another idea is to never touch the rboot config area. Slot #0 (xboot) will be the permanent rboot target and when it has finished it will set slot #1 as a temporary rboot target and reboot. The overhead in time is minimal and rboot can be use as is.

raburton:
@projectgus I'll do what I can to support what you need, but PRs are always appreciated! Especially as I have another set of professional exams coming up.

@kanflo this isn't a bad idea (and I know other people have done this, or at least said they were investigating/ doing this), if you need the extra functionality that can only be achieved in a full sdk-linked rom. The downside is the space used, and given that most of that space is taken up by the sdk code I think many people I've discussed this with previously decided it was easier to just include the functionality they wanted into the regular app rom, where it might only had a couple of kb (wouldn't be an option for some of your suggested uses though, so I think you would have to go the full rom route).

projectgus:
Reworked this quite substantially using latest upstream rboot from @raburton (with some new changes), and removed the option to not use OTA mode.

The "bootloader" directory also has a prebuilt firmware committed inside it, and "make flash" will write that firmware if no other bootloader was compiled. Each time the firmware is replaced esptool.py also flashes the firmware and erases the config (so you always get firmware slot 0.)

> I guess a makefile selectable OTA slot for make flash will come in handy.

This could be a good addition. For now I'm happy with serial flashing always clearing the config (so it goes to slot 0). It's less flexible, but for development it should work OK. What do you think?

How does this all sound?

PR #137, opened 2016-05-03
Better crash dumps

projectgus:
Last PR for today, these are some changes to make crash dumps easier to debug (and two bugs that I found using the tools.)

New tool filteroutput.py will monitor the serial log and automatically call addr2line on anything that looks like a code address. This works surprisingly well, ie here's a sample from deliberately crashing the blink example:

```
Fatal exception (9):
epc1=0x40225575

0x40225575: crashy_mccrash at /home/gus/dev/esp/rtos/open-rtos/examples/blink/blink.c:16
 (inlined by) user_init at /home/gus/dev/esp/rtos/open-rtos/examples/blink/blink.c:65

epc2=0x00000000
epc3=0x00000000
excvaddr=0x5a5abdbd
depc=0x00000000
excsave1=0x40220127

0x40220127: sdk_user_init_task at /home/gus/dev/esp/rtos/open-rtos/core/app_main.c:341

Registers:
a0 40220127 a1 3fff4a70 a2  5a5abdbd a3  abad1d3a

0x40220127: sdk_user_init_task at /home/gus/dev/esp/rtos/open-rtos/core/app_main.c:341

a4  00000001 a5  00000001 a6  3fff4d14 a7  00000008
a8  60000000 a9  000000a0 a10 3fff235c a11 0000000a
a12 00000000 a13 00000000 SAR 0000001f

Stack: SP=0x3fff4a70
0x3fff4a70: 00000000 00000000 00000000 00000000
0x3fff4a80: 3fff56d8 00000000 00000000 40220127

0x40220127: sdk_user_init_task at /home/gus/dev/esp/rtos/open-rtos/core/app_main.c:341

0x3fff4a90: 00000000 00000000 00000000 00000000
0x3fff4aa0: 00000000 00000000 00000000 a5a5a5a5
0x3fff4ab0: 00000000 00000050 3fff4cb0 00000000
0x3fff4ac0: 3fff1b00 3fff1b00 3fff4ab8 3fff1af8
0x3fff4ad0: 0000000f a5a5a5a5 a5a5a5a5 3fff4ab8
0x3fff4ae0: 00000000 00000000 3fff4b08 454c4449
```

* Line numbers directly in the serial output, no need to run an external tool (you can also pipe output from a log file into filteroutput.py if you need to.

* Enough stack information to kind-of-infer the stack trace (we could probably stop at 0xa5a5a5a5 each time, but I go a bit further just in case that value comes up organically...

* The register dump, combined with addrsource on the crashing address and the stack dump, mans you can reconstruct a pretty solid picture of what values were input to the function before it crashed.


projectgus:
(Last commit 6767891 was needed because this PR contains "the straw that broke the camels back", our build output in verbose mode hit Travis' 4MB log limit. Ouch!)

projectgus:
Added an additional commit that provides an abort() implementation that prints the calling address and also dumps stack in the same way the fatal exception handler does, before resetting.

kanflo:
Nice work!

projectgus:
Thanks @kanflo. I'm keen to merge this if noone else has any concerns. @foogod, do you have any time to look at this one? Cheers.

utils/filteroutput.py
```diff
@@ -0,0 +1,105 @@
...
+        help='Serial port to monitor (will monitor stdin if None)',
+        default=None)
+    parser.add_argument(
+        '--baud', '-b',
+        help='Baud rate for serial port',
+        type=int,
+        default=74880)
+    parser.add_argument(
+        '--reset-on-connect', '-r',
+        help='Reset ESP8266 (via DTR) on serial connect',
```
foogod: It might be good to add a note that reset-on-connect will always happen on Linux anyway, because of the way the serial drivers are written (this option may still be useful on non-Linux platforms, though)..

core/exception_vectors.S
```diff
@@ -81,7 +83,9 @@ KernelExceptionVector:
         .type   KernelExceptionVector, @function
 
         break   1, 0
-        call0   sdk_user_fatal_exception_handler
+        mov     a2, a1
+        movi    a3, 0
```
foogod: I haven't had a chance to check this, but doesn't this clobber the values of a2 and a3 before we have a chance to print them to the user?

foogod:
Looks like a really good improvement.  I'm all for adding this functionality.

One other suggestion might be while we're changing all this to have the "Fatal exception" line be preceeded by some easily-identifiable and unlikely-to-be-accidentally-generated text which can be used to reliably detect/identify the beginning of a crash dump, but that's kinda minor..

core/exception_vectors.S
projectgus: The register state is only dumped if the second parameter (a3) is set to 1, and it dumps the register as they were saved onto the stack by _xt_int_enter/xtensa_context_save.

I think we could modify the exception handlers to always have access to registers in a fatal exception, but this is less intrusive (and most exceptions happen via the user exception path.) Could be expanded in the future, though.

utils/filteroutput.py
projectgus: Ah, will do. I've been using NodeMCU boards for so long I'd forgotten they're "special" in that DTR & RTS together don't trigger a reset.

projectgus:
Thanks @foogod. I added two more commits, 981c878 dumps some heap statistics and 1e9296f prevents a corner case where crashing inside the fatal exception handler (ie due to heap corruption) could have caused a crash loop.

I have some more changes in another branch, to dump state and stack information for all tasks. But it needs a bit more cleanup.

I'm going to merge this now, but if you see anything in those other two commits then please let me know and I'll send another PR.

I''ll update the [Crash Dumps wiki page](https://github.com/SuperHouse/esp-open-rtos/wiki/Crash-Dumps) soon to explain the new information provided in the dumps.

PR #142, opened 2016-05-23
Store .rodata in flash by default

projectgus:
This may not be ready for merging yet, but I wanted to put it up for discussion/review anyhow.

Moves most .rodata to flash by default. http_get_mbedtls example's RAM footprint shrinks by 5317 bytes (7284 byte .rodata -> 1967 byte.) blink example shrinks by 3981 bytes (5948 to 1967).

Most usefully, you can do things like compile in debug message strings without taking a RAM hit for them.

Fundamentally there's not a lot in this change, it draws on the code we've had in e-o-r for a while to do unaligned reads from flash. The main difference was testing various types of operation and watching for unwanted/risky accesses that might happen when the flash caching layer was disabled.

I'm not keen to merge yet because of potential performance implications. The GNU linker doesn't seem to place .rodata constants anywhere near the code which accesses them, and based on my experiments understanding the flash caching layer this could be a significant performance hit. I'm going to keep looking into that and I'll provide some more information soon.

(That said, I haven't yet benchmarked any actual applications before and after this change, so it may be that the difference is mostly academic.)

### DIFF
--- a/ld/common.ld
+++ b/ld/common.ld
@@ -132,43 +132,6 @@ SECTIONS
     _etext = .;
   } >iram1_0_seg :iram1_0_phdr
 
-  .irom0.text : ALIGN(4)
-  {
-    _irom0_text_start = ABSOLUTE(.);
-    /* esp-open-rtos compiled code goes into IROM by default
-       (except for libgcc which is matched above.)
-    */
-    *(.literal .text .literal.* .text.*)
-    /* Anything explicitly marked as "irom" or "irom0" should go here */
-    *(.irom.* .irom.*.* .irom0.*)
-    _irom0_text_end = ABSOLUTE(.);
-
-    /* Temporary .rodata hacks start here, eventually all rodata will
-    be in irom by default */
-    /* mbedtls rodata */
-    *mbedtls.a:*.o(.rodata.* .rodata)
-    /* actual certificate in example (TEMPORARY HACK) */
-    *:cert.o(.rodata.* .rodata)
-    /*  C++ constructor and destructor tables, properly ordered:  */
-    __init_array_start = ABSOLUTE(.);
-    KEEP (*crtbegin.o(.ctors))
-    KEEP (*(EXCLUDE_FILE (*crtend.o) .ctors))
-    KEEP (*(SORT(.ctors.*)))
-    KEEP (*(.ctors))
-    __init_array_end = ABSOLUTE(.);
-    KEEP (*crtbegin.o(.dtors))
-    KEEP (*(EXCLUDE_FILE (*crtend.o) .dtors))
-    KEEP (*(SORT(.dtors.*)))
-    KEEP (*(.dtors))
-    /*  C++ exception handlers table:  */
-    __XT_EXCEPTION_DESCS__ = ABSOLUTE(.);
-    *(.xt_except_desc)
-    *(.gnu.linkonce.h.*)
-    __XT_EXCEPTION_DESCS_END__ = ABSOLUTE(.);
-    *(.xt_except_desc_end)
-
-  } >irom0_0_seg :irom0_0_phdr
-
   .data : ALIGN(4)
   {
     _data_start = ABSOLUTE(.);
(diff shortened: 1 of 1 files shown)


projectgus:
I've tested this more thoroughly and done a few benchmarks and I think this is fine to merge.

First test was measuring how many RTOS ticks the "handshake" step in the http_get_mbedtls function took. From about 20 samples before/after I couldn't see any correlation, the error caused by the network was more than any performance difference.

Then this code to measure AES performance:

```
void user_init(void)
{
    uart_set_baud(0, 115200);
    static uint8_t data[1024];
    static uint8_t output[1024];
    static uint8_t iv[16];

    static uint8_t key[256 / 8];

    memset(data, 0, sizeof(data));
    memset(iv, 0, sizeof(iv));

    mbedtls_aes_context ctx;
    uint32_t before, after;
    RSR(before, CCOUNT)
    mbedtls_aes_init(&ctx);
    mbedtls_aes_setkey_enc(&ctx, key, 256);

    for(int r = 0; r < 10; r++) {
        mbedtls_aes_crypt_cbc(&ctx,
                              MBEDTLS_AES_ENCRYPT,
                              sizeof(data),
                              iv,
                              data,
                              output);
        memcpy(data, output, 1024);
    }
    RSR(after, CCOUNT);
    printf("cycle count %d\n", after - before);
    vPortExitCritical();
    while(1) {}
}
```

... AES is a good test because all the constant tables are stored in .rodata.

I re-ran the tests measuring the FRC2 timer as well, to make sure CCOUNT wasn't stalling during flash cache misses, The timer tick counts were similar.

| Flash Mode | branch | CCOUNT delta | FRC2 delta |
| ---------------- | --------- | --------------------- | ---------------- |
| qio              | master  | 2800494          | 175041       |
| qio              | rodata_in_flash | 2800808  | 175063  |
| dio              | master  | 2828848          | 176822       |
| dio              | rodata_in_flash | 2828861 | 176812   |

(WDEV.SYS_TIME also gave similar results.)

From what I've measured of SPI flash cache behaviour, I think cache misses happen enough due to normal .irom text section which is why this doesn't make a large difference. Have to do a bit more analysis to be sure that's why, but I think this is ready to merge.

projectgus:
Remembered that the master branch already puts rodata in flash for mbedtls (otherwise TLS doesn't work), which explained why nothing significant changed. I've amended the table above with new numbers.

Seems that using RAM for AES data tables is 2.35x faster than using SPI flash, regardless of QIO or DIO flash modes. I think that's probably close to a worst-case slowdown for real world code (the AES functions use a _lot_ of rodata constants.) Most code will see a much lower impact.

My view is that this is worth merging given the RAM it frees up, given RAM is such a scarce resource. I've recently been working on a project where we needed this change in order to have TLS as well as other features enabled concurrently.

If there was some way to modify/configure the linker so that the rodata sections stayed close in flash to the functions that used them, I'm pretty confident that the 2.35x slowdown could be brought down a _lot_, as well.

foogod:
> If there was some way to modify/configure the linker so that the rodata sections stayed close in flash to the functions that used them, I'm pretty confident that the 2.35x slowdown could be brought down a *lot*, as well.

I think that's kinda a big assumption.. it would really depend a lot on the cacheline size and associativity in the flash cache, which as far as I know is still a complete black box (a while back I put a couple of ideas for ways to test/probe it on the RE wiki, but haven't had any time to try them out)..

If the cache is made up of a few large cachelines, then proximity could make a big difference.  If there's smaller cachelines with decent associativity, it would probably make little or no difference.  If the cachelines are smaller, but with low associativity, it could be a lot harder to predict (and in particularly bad cases, could completely change from one firmware to the next regardless of where we put things, just due to how the rest of the application code flow moves).

In any case, I have no problem with slowing things down a bit to make this change.. I think in general RAM availability is more critical to most people than raw execution speed.

I haven't had time to really look over the PR, but wanted to check.. do we have some way to tell the compiler/linker that a particular piece of static data _should_ go in RAM instead?  (last I checked, the `IRAM` qualifier only worked for code, not data)

projectgus:
> I think that's kinda a big assumption.. it would really depend a lot on the cacheline size and associativity in the flash cache, which as far as I know is still a complete black box

I didn't make it clear enough, but I've actually done some probing and I think I have a fairly good idea how the cache works now. I'm going to make a couple of checks to verify and then I'll post what I have somewhere.

> In any case, I have no problem with slowing things down a bit to make this change.. I think in general RAM availability is more critical to most people than raw execution speed.

I agree.

> 

I haven't had time to really look over the PR, but wanted to check.. do we have some way to tell the compiler/linker that a particular piece of static data should go in RAM instead? (last I checked, the IRAM qualifier only worked for code, not data)

I will check and add that to the PR. I'm not 100% sure we can do it for string literals (storage modifier attributes don't apply directly to them), but I think ESP8266 Arduino does it somehow so I'll have a peek at how they manage it.

projectgus:
Oops, accidentally merged without adding the macro - but it's there now.

PR #144, opened 2016-05-28
MQTT refactor, use submodule, fix select() bug

projectgus:
Refactor MQTT client, replace select() with lower-impact LWIP features.

Changes the MQTT API a small amount, to bring it into line with upstream Paho MQTT C client.

@urx, do you have any thoughts about these changes?


urx:
@projectgus all this changes seems good for me. Code has become better than it was :)

yesco:
Should one really include changes in the file: include/ssid_config.h ?

I'm curious about using lwip instead of normal sockets, in my case I'd prefer normal sockets as the code would be the same (?) under linux. One of my reasons of using esp-open-rtos is that I prefer to have portable code that works cross platform. Helps for debugging too, in my experience.

kanflo:
Nice catch! Nope, ssid_config.h should not be included in the change.

projectgus:
Sorry for the delays, I've been on holidays for a while.

> Nice catch! Nope, ssid_config.h should not be included in the change. 

Yes, bit of a head/desk moment there. Fixed now.

> I'm curious about using lwip instead of normal sockets, in my case I'd prefer normal sockets as the code would be the same (?) under linux. One of my reasons of using esp-open-rtos is that I prefer to have portable code that works cross platform. Helps for debugging too, in my experience.

I see where you're coming from here, it is good having the option of using the same BSD socket interface eveywhere. In the case of the Paho MQTT client, though, there is already a separate Linux implementation included in the main MQTTClient distribution - so you still get to use the same MQTTClient interface for both Linux & esp-open-rtos if you'd like, just the internals that are different. The existing esp-open-rtos BSD socket implementation wouldn't work on Linux either I don't think, without some modifications. And this LWIP-only implementation uses less RAM and yields a smaller compiled code side.

projectgus:
Since #158 this is out of date, because we needed to patch a bug in the upstream MQTTClient.c.

I'm going to close this for now, at least until that fix gets merged into the upstream repo.

PR #145, opened 2016-05-31
Set correct base address for register HOST_INF_SEL

svenschwermer:


foogod:
I suppose this is probably a reasonable change..  Note, however, that `eagle_soc.h` is essentially an old "compatibility" header which is not really intended to be used directly by esp-open-rtos apps anyway.  The preferred way to access this register in esp-open-rtos is actually to include `esp/dport_regs.h` and access it as `DPORT.PERI_IO` (with the `DPORT_PERI_IO_*` bit constants, etc)

PR #149, opened 2016-06-21
sntp: free the pbuf after sending the request

iosen:
Fixes a memory leak.


kanflo:
Nice catch, thanks!

projectgus:
This looks like a good fix, thanks for contributing it. I think we could bypass the check of bpuf->ref - the pbuf is allocated inside the function so it should always have ref count 1. What do you think?

iosen:
Thanks for the feedback. Removed the 'bpuf->ref'.

projectgus:
Thanks!

PR #150, opened 2016-06-21
Call DisconnectNetwork(...) before MQTT reconnect

kanflo:
I noticed I could not reconnect to my RasPi MQTT server following a disconnect. Adding DisconnectNetwork(...) did the trick.


projectgus:
Looks good to me, thanks.

Issue #147, opened 2016-06-06
No expected effect when changing configCPU_CLOCK_HZ or configTICK_RATE_HZ


Zaltora:
Hi, I tried to change some settings of FreeRTOSConfig.h for time control and see what happen.
This is my code "tick.cpp" : 

```
#include "FreeRTOSConfig.h"
#include "espressif/esp_common.h"
#include "esp/uart.h"
#include "FreeRTOS.h"
#include "task.h"
#include "esp8266.h"


void handler_task_loop (void *pvParameters)
{


    uint32_t time = sdk_system_get_time() ;
    uint32_t time_rtos = xTaskGetTickCount()*portTICK_RATE_MS ;

	while(1) {

	printf("Uptime_rtos:%u ms\t\t",xTaskGetTickCount()*portTICK_RATE_MS);
	printf("Uptime_sdk:%u ms\n",sdk_system_get_time()/1000);
	printf("loop_rtos:%u ms\t\t",(xTaskGetTickCount()*portTICK_RATE_MS)-time_rtos);
	printf("loop_sdk:%u ms\n",(sdk_system_get_time()-time)/1000);
	printf("------------------------------\n");
	time_rtos = xTaskGetTickCount()*portTICK_RATE_MS ;
	time = sdk_system_get_time() ;
	vTaskDelay(100/portTICK_RATE_MS);
	}
}


extern "C" void  user_init(void)
{
	uart_set_baud(0, 115200);
	printf("SDK version:%s\n", sdk_system_get_sdk_version());

	sdk_system_update_cpu_freq(SYS_CPU_160MHZ) ;

	printf("CPU freq:%u\n", sdk_system_get_cpu_freq());
	printf("CPU clock:%u\n", configCPU_CLOCK_HZ);
	printf("TICK RATE:%u\n", configTICK_RATE_HZ);
	printf("------------------------------\n");

    xTaskCreate(handler_task_loop, (signed char *)"loop", 256, NULL, 3, NULL);
}
```

when i set these settings  in FreeRTOSConfig.h file : 

```
#define configTICK_RATE_HZ			( ( portTickType ) 1000 )
#define configCPU_CLOCK_HZ			( ( unsigned long ) 160000000 )
```
this is my output:

```
SDK version:0.9.9
CPU freq:160
CPU clock:160000000
TICK RATE:1000
------------------------------
mode : softAP(1a:fe:34:de:a0:cf)
add if1
bcn 100
[...]

Issue #80, opened 2016-01-05
Update to FreeRTOS 8


DagAgren:
Has anyone looked at how difficult it would be to update to FreeRTOS 8? Personally I much prefer it, just because the various type names are not such an ugly mess any more.

I might have a look at it later if nobody has tried it yet, if I can find the time.

projectgus:
I've thought about it, but I have not looked into it yet. I'd be very interested to see how you go, and I'd be keen to merge any working update (slash help with debugging).

The main thing that was holding me back is that I think a tiny number (one or two) FreeRTOS symbols  are still in the binary libraries. But it's the kind of thing that should be easy to pull out.

DagAgren:
Remember which, or how to find them?

Right now I am trying to decide between implementing a project on ESP8266, or EMW3165. The latter has much nicer hardware than the mess on the ESP8266, but it has no equivalent to the esp-open-rtos project so you are stuck with terrible vendor libraries... Been trying to see if I can sort it out a bit but it doesn't look too promising.

Then again, I need both WiFi and some simnple audio output, and while you can abuse the I2S on the ESP to do it, it seems the pins are quite noisy and it doesn't work too well. PWM audio on the STM32F4 core in the EMW3165 is much easier and more reliable... Sigh.

projectgus:
I couldn't remember but I took a look through a build/program.map file to jog my memory. It's not as bad as I remembered, `vPortYield` in libmain.a is I can see and it's probably safe to leave alone.

The other pain point that I can see with porting is anywhere the SDK libraries call into FreeRTOS functions, if the signatures/names have changed. This is where it'll be good once we have all of the SDK libraries available at least as plain assembly files, so you can just edit the function calls. The `feature/assemble_sdk` branch has a start I made on this but it doesn't work yet.

Regarding microcontroller choice, that's a hard one. The other option, I guess, would be to use esp8266 plus a second micro/chip to do the audio over SPI or I2S. Probably still cheaper than EMW3165, but more fiddly to assemble.

DagAgren:
Supposedly FreeRTOS 8 is backwards compatible, so in theory it shouldn't be a problem. Main big visible changes are renaming types and using stdint.h, but that should leave the actual signatures the same.

Wish somebody sold an I2S codec on a convenient little module, that would be fairly easy to use from the ESP8266, since I got I2S working. Not feeling like making circuit boards though.

foogod:
[...]

PR #154, opened 2016-06-30
First open_esplibs batch, libmain (most of libmain open sourced)

projectgus:
@foogod has done great work with this and I think now is a good time to merge it. I originally wanted to get automated tests running firsts, but that's been too long coming (my fault) and there are issues (#147, possibly #80) that it covers.

I've tested many of the more complex examples (http_get_mbedtls, ota_basic, mqtt-client) with this and they all seem fine. There may be more subtle bugs introduced, but I think the only way to find them will be if people start using these changes.

@foogod has made an excellent set of compile-time selection switches, so if there are bugs they should be easy to isolate.


projectgus:
Hmm, I've done something wrong with the merge here - it's pulled in a bunch of double commits ie 36886412 vs b0844b01.

@foogod, if I rebase this whole branch against current master then will that make things hard for you in any way?

core/include/esp/sar_regs.h
```diff
@@ -0,0 +1,39 @@
...
+
+/* Note: This memory region is not currently well understood.  Pretty much all
+ * of the definitions here are from reverse-engineering the Espressif SDK code,
+ * many are just educated guesses, and almost certainly some are misleading or
+ * wrong.  If you can improve on any of this, please contribute!
+ */
+
+struct SAR_REGS {
+    uint32_t volatile _unknown0[18];   // 0x00 - 0x44
+    uint32_t volatile UNKNOWN_48;      // 0x48 : used by sdk_system_restart_in_nmi()
```
pfalcon: These two lines show how inconsistent naming is, leading to stuff like "unknownc" above. Can there be convention like "unkown_0c" (underscore, 2 digits)?

foogod: These are not inconsistant.  "_unknown" is explicitly intended to be a temporary placeholder for spacing purposes and should not be referenced by any code (there may be one or two places that do reference them, but if there is that's an error and needs to be cleaned up).

The `UNKNOWN_48` is an example of the (rare) situation where a field needed to be referenced somewhere but I still could not figure out what it should be named from the context.  The uppercase indicates it's not just a spacer/placeholder but is actually a referenced field.

If you want to go through and clean up the "_unknown" names with underscores/leading digits/etc, you certainly can, but frankly it seems pretty minor to me, considering that they are explicitly not supposed to be being referenced by anything (but if somebody wants to do a PR for that that's fine by me)..

foogod:
Feel free to rebase the branch (I don't think it'll screw up anything for me at this point, but if it does, I can deal with it :) )

projectgus:
Thanks, rebased. This looks ready to merge to me.

open_esplibs/libmain/timers.c
```diff
@@ -0,0 +1,100 @@
...
+    new_entry->next = 0;
+    tailptr = &entry->next;
+    if (!timer_list) {
+        tailptr = &timer_list;
+    }
+    *tailptr = new_entry;
+}
+
+void sdk_os_timer_arm(ETSTimer *ptimer, uint32_t milliseconds, bool repeat_flag) {
+    if (!ptimer->timer_handle) {
```
UncleRus: Bug in condition

PR #159, opened 2016-07-06
DHT11/DHT22 sensor library fixed.

sheinz:
Fixed DHT library.
Tested with DHT11 and DHT22 modules.


projectgus:
Thanks @sheinz, this looks really good.

@jsuiker sent the original DHT implementation - do you have any comments on this?

jsuiker:
If it works, it's good in my book :)  I'm glad someone got a chance to test both units and make some improvements...

PR #160, opened 2016-07-08
BMP280 pressure sensor driver and example.

sheinz:
BMP280 is a successor of BMP180 pressure sensor. But it has different communication protocol. It has different registers and different operation modes.
At first I thought about extending existing BMP180 driver but there would have been two many preprocessor conditions or run-time branching.

So, here's my implementation of BMP280 driver.
I tested it with my BMP280 module. The output data is the same as with ArduinoUNO + BMP280.

### DIFF
--- a/examples/bmp280/Makefile
+++ b/examples/bmp280/Makefile
@@ -1,0 +1,3 @@
+PROGRAM=BMP280_example
+EXTRA_COMPONENTS = extras/i2c extras/bmp280
+include ../../common.mk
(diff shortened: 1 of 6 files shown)


projectgus:
Thanks for this, looks very good to me.

Does anyone have any other review comments?

kanflo:
I think this one is good for merging Thanks @sheinz!

projectgus:
Merged. @sheinz if you have time could you please confirm that the BMP280 example works as expected with the current master? (there's been changes to the i2c driver & the esp-open-rtos core since the PR went up. The merge is clean and the example builds, but I don't have a sensor to test with.)

sheinz:
I've just verified. My bmp280 module works with esp-open-rtos master head.

PR #163, opened 2016-07-15
File system support. SPIFFS integration.

sheinz:
This PR includes:
 * **SPIFFS** as a submodule in `extras/spiffs`
 * SPIFFS configuration and flash read/write access in `extras/spiffs`
 * **mkspiffs** utility to create SPIFFS image for flashing
 * example in `examples/spiffs`
 * POSIX file system test **fs-test** as a submodule in `example/posix_fs`
 * file system test example in `example/posix_fs`
 * build-process integration for image creation and flashing

POSIX file operations support:
 * open/close
 * read/write
 * lseek
 * stat/fstat (only size field is available)

File system test includes two tests:
 * Load test. Performs various file system operations in random order and verifies data validity.
 * Speed test. Simple speed benchmark.

The main motivation for writing speed test was the need for some baseline for future SPI flash access optimization. Currently flash is accessed via SDK/bootrom functions which require buffers/addresses/size alignment. The wrapper functions handle unaligned access from SPIFFS but they introduce overhead.
I intend to rewrite access to flash to eliminate unaligned access overhead or to make it as minimal as possible. Or prove that it is not possible :)
I already reverse engineered flash write functions from bootrom and results are promising.
So, tests will help with messing around with flash access.

Comments/suggestions are welcome!


kanflo:
I am glad we're finally getting this "must have" into EOR. My merge finger is itching on the trigger, do you have any comments @projectgus?

One thing that could be discussed later but I still want to bring up now is how the location of the FS is stored. I think we might find we need something like a partition table or storing start and size of SPIFFS, @foogod's param area and so on.

sheinz:
Just a quick update.

I've refactored SPI flash access. It doesn't use SDK/open_esplib access to flash. It uses SPI registers directly to perform read/write/erase operations. This eliminated unnecessary addresses/buffers/sizes alignment. The only alignment that was really needed is for writing into 32 bits registers SPI(0).W

There're currently two issues.

1. SPI(x).W[16]  are data registers for SPI operation. Each data registers is 32 bit long. So, the total 16x4 = 64 bytes memory blocks can be read/written from/to SPI at a time. The bootrom code uses only 32 bytes blocks for read/write SPI flash memory. Why?
I tried to use larger blocks for SPI flash read/write with the following results. The write operations with blocks up to 64 bytes works perfectly. The read operations works for blocks up to 63 bytes. For 64 bytes blocks read operation hangs.

I created an isolated test to check this issue [spi-flash-test](https://github.com/sheinz/esp-spi-flash-test).
Also asked on [Espressif forum](http://bbs.espressif.com/viewtopic.php?f=6&t=2439).

2. No file system speed improvement after switching to low lever SPI flash access. The slowest part of file system operation is the actual flash read/write so the time on processing data is negligible comparing to flash read/write.
But what I don't understand it seems like file system speed has degraded after refactoring :(
No idea why... Still figuring out...
SPI flash access is not completely clear to me.

sheinz:
I finally managed to reach the desired speed performance.
The problem was with accessing memory with 8 bits load/store operations. They seem to slow everything quite a bit.
I had to write a couple of functions in assembler to perform optimal read/write from/to unaligned memory buffer.

The speed increase is not that great. But in my opinion the clear open source flash access is much more valuable.

## Speed test results
Flash access via SDK + bootrom:

    Read speed: 718229 bytes/s
    Write speed: 12293 bytes/s

    Read speed: 707842 bytes/s
    Write speed: 11875 bytes/s

    Read speed: 720278 bytes/s 
    Write speed: 12356 bytes/s

Flash access using own implementation:

    Read speed: 749858 bytes/s
    Write speed: 12759 bytes/s

    Read speed: 747860 bytes/s
    Write speed: 12650 bytes/s

    Read speed: 742006 bytes/s
    Write speed: 12490 bytes/s

The problem with 64 bytes chunks read remains. Current implementation uses 64 bytes blocks for write and 60 bytes blocks for read.

My implementation here is finished. Please review/comment.

extras/spiffs/esp_spiffs.c
```diff
@@ -0,0 +1,203 @@
...
+int _open_r(struct _reent *r, const char *pathname, int flags, int mode)
+{
+    uint32_t spiffs_flags = SPIFFS_RDONLY;
+
+    if (flags & O_CREAT)    spiffs_flags |= SPIFFS_CREAT;
+    if (flags & O_APPEND)   spiffs_flags |= SPIFFS_APPEND;
+    if (flags & O_TRUNC)    spiffs_flags |= SPIFFS_TRUNC;
+    if (flags & O_RDONLY)   spiffs_flags |= SPIFFS_RDONLY;
+    if (flags & O_WRONLY)   spiffs_flags |= SPIFFS_WRONLY;
+
```
pellepl: you may want to include O_DIRECT and O_EXCL also, as it is supported in spiffs

sheinz: Thanks for the review! I will include O_EXCL. Unfortunately **newlib** is configured the way that O_DIRECT is not available in esp-open-rtos.

extras/spiffs/spiffs_config.h
```diff
@@ -0,0 +1,263 @@
...
+#ifndef SPIFFS_COPY_BUFFER_STACK
+#define SPIFFS_COPY_BUFFER_STACK        (64)
+#endif
+
+// Enable this to have an identifiable spiffs filesystem. This will look for
+// a magic in all sectors to determine if this is a valid spiffs system or
+// not on mount point. If not, SPIFFS_format must be called prior to mounting
+// again.
+#ifndef SPIFFS_USE_MAGIC
+#define SPIFFS_USE_MAGIC                (0)
```
pellepl: I recommend enabling this...

extras/spiffs/spiffs_config.h
```diff
@@ -0,0 +1,263 @@
...
+#endif
+
+#if SPIFFS_USE_MAGIC
+// Only valid when SPIFFS_USE_MAGIC is enabled. If SPIFFS_USE_MAGIC_LENGTH is
+// enabled, the magic will also be dependent on the length of the filesystem.
+// For example, a filesystem configured and formatted for 4 megabytes will not
+// be accepted for mounting with a configuration defining the filesystem as 2
+// megabytes.
+#ifndef SPIFFS_USE_MAGIC_LENGTH
+#define SPIFFS_USE_MAGIC_LENGTH         (0)
```
pellepl: ... and enabling this also.

extras/spiffs/spiffs_config.h
```diff
@@ -0,0 +1,263 @@
...
+#define SPIFFS_CFG_PHYS_SZ(ignore)        (SPIFFS_SIZE)
+#endif
+#ifndef SPIFFS_CFG_PHYS_ERASE_SZ
+#define SPIFFS_CFG_PHYS_ERASE_SZ(ignore)  (4*1024)
+#endif
+#ifndef SPIFFS_CFG_PHYS_ADDR
+#define SPIFFS_CFG_PHYS_ADDR(ignore)      (SPIFFS_BASE_ADDR)
+#endif
+#ifndef SPIFFS_CFG_LOG_PAGE_SZ
+#define SPIFFS_CFG_LOG_PAGE_SZ(ignore)    (256)
```
pellepl: I'd like to run some tests to optimize spiffs on the esp regarding page and blocksizes. I will try doing that this week, if you want to want for it. otherwise I can make a pull request of it.

extras/spiffs/spiffs_config.h
sheinz: Enabled. Will test and update PR. Thanks.

extras/spiffs/esp_spiffs.c
```diff
@@ -0,0 +1,203 @@
...
+            cache_buf.buf, cache_buf.size, 0);
+
+    if (err != SPIFFS_OK) {
+        printf("Error spiffs mount: %d\n", err);
+    }
+
+    return err;
+}
+
+#define FD_OFFSET 3
```
pellepl: Minor: you handle file handle offsets nicely here already. FYI, there is also support in spiffs for that: https://github.com/pellepl/spiffs/blob/master/src/default/spiffs_config.h#L192 . This _might_ be preferred if we want to support more media apart from stdio and files in the future. Then agan, we can do that when it is necessary (if ever).

extras/spiffs/spiffs_config.h
sheinz: I tried only those values. I thought that logical page size should be the same as physical page size for optimal performance.
The flash chip also supports block erase of 32k and 64k. But it is not implemented currently. I'm not sure if we need it.

extras/spiffs/esp_spiffs.c
sheinz: I missed that in configuration file. For sure it's better to use the support in spiffs. Thanks again. Will fix that.

extras/spiffs/spiffs_config.h
pellepl: On the top of my head I think a page size of 128 would be better, perhaps
in combination of a logical block size of 2*4*1024. Spiffs will always use
the ```PHYS_ERASE_SZ``` for erasing, so having a 8kB logical block means
spiffs will call erase hal func two times when it needs to erase a block.

Yes, the runtime performance will suffer a bit from having smaller page
sizes, but the flash footprint will be better (i.e. the spiffs structure
will be more filled). I need to verify this to be sure. Perhaps a 256 page
with a 8kB block would be optimal..

2016-07-22 10:33 GMT+02:00 sheinz <notifications@github.com>:

extras/spiffs/esp_spiffs.c
pellepl: Just don't forget the extra config for the actual offset in the config struct, ```fh_ix_offset``` (https://github.com/pellepl/spiffs/blob/master/src/spiffs.h#L220). And if I haven't said it before, great work :)

sheinz:
Thanks @pellepl for the review. Updated the PR. Nobody knows spiffs better than the author, so it's supper awesome to have such a reviewer :)

extras/spiffs/mkspiffs/mkspiffs.c
```diff
@@ -0,0 +1,236 @@
...
+
+static s32_t _read_data(u32_t addr, u32_t size, u8_t *dst)
+{
+    memcpy(dst, (uint8_t*)image + addr, size);
+    return SPIFFS_OK;
+}
+
+static s32_t _write_data(u32_t addr, u32_t size, u8_t *src)
+{
+    memcpy((uint8_t*)image + addr, src, size);
```
pellepl: i do not recommend using a memset here. Instead, all data should be ANDed to destination to mimic a NOR flash. Otherwise, some flags may become corrupt in the fs.

sheinz: Oh, I see. You mean memcpy. I agree. In my case the previous value is completely overwritten. But with the real flash only 1 flipped to 0, not vice versa.

pellepl: Yes, sorry, meant memcpy.

projectgus:
Hi @sheinz ,

Thanks so much for this, it looks really impressive! Very nicely integrated into the build system, and being able to do unaligned flash access is a huge boon.

Thanks also @pellepl for reviewing! And @kanflo for some good suggestions to look at down the line.

I had a quick look through the code, and a quick test on my nodemcu board here, and everything looks good to me. Once you're done with any remaining discussions/considerations then I'm happy to merge.

One minor thing I noticed is that my go-to target 'make test' didn't build the spiffs. Neither does 'make flash', because these don't depend on 'all'. This would be an easy thing to fix in common.mk (looks like 'all' has too many dependencies as well!), but maybe it's better this way if it avoids unnecessarily regenerating the file?

> I had to write a couple of functions in assembler to perform optimal read/write from/to unaligned memory buffer.

Nice. I'm impressed these are so much faster than newlib's memcpy. I even went to look if flash caching might be the reason, but your implementation is in IROM and newlib's is in IRAM! So not that. Perhaps we should swap your function (or an extended version) in place of the current memcpy?

> Unfortunately newlib is configured the way that O_DIRECT is not available in esp-open-rtos.

FWIW, if you'd like to reconfigure/recompile newlib then this is possible. There's a wiki page that explains how it's set up: https://github.com/SuperHouse/esp-open-rtos/wiki/libc-configuration

Depends on what you think about the possible benefits, I guess.

Cheers,

Angus

sheinz:
Hi Angus,

> One minor thing I noticed is that my go-to target 'make test' didn't build the spiffs. Neither does 'make flash', because these don't depend on 'all'. This would be an easy thing to fix in common.mk (looks like 'all' has too many dependencies as well!), but maybe it's better this way if it avoids unnecessarily regenerating the file?

Fixed spiffs image build in 'make test' and 'make flash'.

> Nice. I'm impressed these are so much faster than newlib's memcpy. I even went to look if flash caching might be the reason, but your implementation is in IROM and newlib's is in IRAM! So not that. Perhaps we should swap your function (or an extended version) in place of the current memcpy?

My implementation of unaligned memcpy is in IRAM `.section .iram1.text, "x"`
newlib's memcpy might be slow if it uses byte length load/store operations. I don't know why but they seem slow. My experiments also show that memcpy from the bootrom is pretty fast. There's also **xthal_memcpy** version from [lx106-hal](https://github.com/tommie/lx106-hal) which handles unaligned source and destination and pretty fast as well. It is also available in the bootrom.

Adding support of O_DIRECT might be a topic for a new PR. 

Thanks for the review.

extras/spiffs/mkspiffs/mkspiffs.c
sheinz: Fixed

projectgus:
Nice!

> My implementation of unaligned memcpy is in IRAM .section .iram1.text, "x"

Derp. I must have been pretty tired when I was looking at this last night!

> newlib's memcpy might be slow if it uses byte length load/store operations. I don't know why but they seem slow.

It does, at least up until the first 4 byte aligned address: https://github.com/projectgus/newlib-xtensa/blob/xtensa-2_2_0-lock/newlib/libc/machine/xtensa/memcpy.S

This will be very slow reading from instruction space because ESP8266 doesn't support unaligned loads there, so there's a routine in core/exception_vectors.S which catches what would otherwise be a fatal CPU exception and completes it via an aligned load (with the associated performance hit).

If it's slow when reading from data RAM then that's less expected, but I guess it's not totally surprising.

I'll open another issue to track the idea of swapping in a faster memcpy implementation.

Rest of this looks good to me. Time to merge?

pellepl:
@sheinz, sorry for being late about this, but it seems performance would be a wee bit better having 8kB logical blocksize and 256 byte page size.
In fact, instead of hard coding these settings they could perhaps be defaulted and modifiable from local.mk if wanted?
These are all suggestions, from my point of view it could very well be merged. Great work, @sheinz!

kanflo:
Lovely work @sheinz, please go ahead and merge @projectgus.

sheinz:
> it seems performance would be a wee bit better having 8kB logical blocksize and 256 byte page size.

Indeed the performance slightly increases with 8kB logical block size. I have changed it to 8kB.

> In fact, instead of hard coding these settings they could perhaps be defaulted and modifiable from local.mk if wanted?

It is a good idea. It might be a separate PR as the build process is getting a little bit tricky. Both firmware and mkspiffs use those configurations. And mkspiffs must be rebuilt if any configuration is changed. 

I've found a standalone utility with the same name (what a coincidence) [mkspiffs](https://github.com/igrr/mkspiffs) . I might have not written my own if I found it earlier. But the problem is that igrr's mkspiffs uses older version of spiffs and uses dynamic configuration whereas   firmware is configured to use spiffs as a singleton. I'm not sure if the image produced with spiffs dynamically configured would be compatible with spiffs as a singleton.

pellepl:
Ah, I thought you've got inspired from @igrr when you wrote it. :) If you want to use @igrr's version you can go ahead - the singleton config is only a runtime thing and do not affect the structure of the actual file system. On the other hand, what you have now works fine - the mkspiffs part (any of them) are neither big or very complex, so I'd say go for a merge now.

alainmaes:
@sheinz, As rboot is integrated in esp-open-rtos it would be nice to have 2 spiffs partitions next to the 2 rom partitions. That would allow to safely install a new filesystem together with the new rom. Basically it would come down to adding an optional parameter to the mount function to select the base address based on the rom that is loaded. Would that be possible?

pellepl:
>>Unfortunately newlib is configured the way that O_DIRECT is not available in esp-open-rtos.
>FWIW, if you'd like to reconfigure/recompile newlib then this is possible. There's a wiki page that explains how it's set up: https://github.com/SuperHouse/esp-open-rtos/wiki/libc-configuration
>
>Depends on what you think about the possible benefits, I guess.

From spiffs point of view using O_DIRECT would be a very special case. Not something one would normally do on a spi flash as may it mess with wear. I'd gather the work is a lot deal greater than the benefits of it. Also, this feature can be reached by using the spiffs api directly. As the fd offsets now are handled by spiffs itself, a posix fd equals a spiffs fd, meaning it is pretty seamless.

sheinz:
> As rboot is integrated in esp-open-rtos it would be nice to have 2 spiffs partitions next to the 2 rom partitions. That would allow to safely install a new filesystem together with the new rom. Basically it would come down to adding an optional parameter to the mount function to select the base address based on the rom that is loaded. Would that be possible?

Currently spiffs is configured as a singleton and spiffs base address with size are compile time defined. I didn't think more than one instance of spiffs might be needed at that time. As it turned out I was wrong :) Nevertheless I would leave this PR as it is and suggest dynamic spiffs configuration for another PR.

You can still use single spiff partition with both roms unless your applications require different set of spiffs files.

@pellepl What do you think, should we switch to dynamic spiffs configuration by default? Or, maybe, provide compile time option to select how spiffs should be configured?

kanflo:
I would prefer dynamic base address and size but with a compile time option we could cater for both flavors.

pellepl:
> @pellepl What do you think, should we switch to dynamic spiffs
configuration by default? Or, maybe, provide compile time option to select
how spiffs should be configured?

Well, based on how I seen people use spiffs I'd say it is hard to do it
right for every need. Some people use only one plain vanilla fs instance,
others want two instances simultaneously, yet others want one per boot
image etc etc. I have even come across a quadruple case.

Then again, the majority is one vanilla system of course.

I think skipping the singleton config is a good idea. When populating the
spiffs config struct, take values from the local makefile as you do now for
location and size. Page- and blocksizes can be defaulted to the values you
have today, unless overridden in makefile.

This way you have a working vanilla as of today, but also a skeleton of how
to tinker with it if needed.

Or, just push current and we can fix above in a future PR.😉

kanflo:
@sheinz, could you consider adding the submodule ```posix_fs/fs-test``` to ```feature/spiffs```? I think it would be cleaner to have everything EOR related directly in the EOR git and only use submodules for 3rd party libraries such as LWIP, SPIFFS etc.

sheinz:
Hi @kanflo, 

`fs-test` is not esp8266 or eor specific. It is general POSIX file operation test. As it lives in its own repository it has its own Travis CI. It might be useful for testing not only spiffs but any other file system that comply with POSIX specification. I can copy `fs-test` to eor but new tests that might be added into `fs-test` would be harder to integrate into `eor` later.
What disadvantages do you see having `fs-test` as a submodule?

Another point is that `examples/posix_fs` is not really an example. It's more like a test. I think I've seen someone suggested having set of tests that might run automatically on the real hardware to ensure regression free development.

kanflo:
Hi @sheinz. Oh, I had missed that. In that case it should be a submodule. As it is a test it could go into a separate directory but ```examples/``` is good for now. I will go ahead and merge. Your effort is much appreciated.

projectgus:
Nice work everyone!

PR #164, opened 2016-07-16
i2c: use open drain outputs.

ourairquality:
The code appeared to be forcing the outputs high rather than using an open drain to let them float high and this attempts to correct this.

Initialize the pins when the driver is initialized.

Enable the internal pull-ups to prevent them floating too high if not connected and damaging the circuit.

### DIFF
--- a/extras/i2c/i2c.c
+++ b/extras/i2c/i2c.c
@@ -44,6 +44,19 @@ void i2c_init(uint8_t scl_pin, uint8_t sda_pin)
     started = false;
     g_scl_pin = scl_pin;
     g_sda_pin = sda_pin;
+
+    // Just to prevent these pins floating too much if not connected.
+    gpio_set_pullup(g_scl_pin, 1, 1);
+    gpio_set_pullup(g_sda_pin, 1, 1);
+
+    // I2C bus idle state.
+    gpio_enable(g_scl_pin, GPIO_INPUT);
+    gpio_enable(g_scl_pin, GPIO_INPUT);
+
+    // Set the pins to a low output state for when they are configured
+    // as outputs.
+    gpio_write(g_scl_pin, 0);
+    gpio_write(g_sda_pin, 0);
 }
 
 static void i2c_delay(void)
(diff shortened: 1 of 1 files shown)


projectgus:
Hi, 

Thanks for getting involved in this project and sending the change! Looks good to me. 

I think the reason for the behaviour you were seeing is that GPIO_OUT_OPEN_DRAIN was only added after the i2c driver. A push/pull output driven low is functionally the same as an open drain output driven low, hence the swapping between OUTPUT/LOW and INPUT. But I think this is a good change - it makes the intention clearer, and removes the possibility of accidentally driving the output high.

Have you tested this with any i2c devices?

Fwiw,  I think it may be possible to never set the pin mode to GPIO_INPUT and have the code still work. Reading gpio_input() on an open drain pin should read the state of the input buffer, so it will register if the pin is pulled low from the other end. I think. So if you feel like tweaking the driver some more, you could explore that. :) 

@kanflo, as the original i2c driver author do you have any comments on this?

ourairquality:
@projectgus Thank you for explaining this, and I had misunderstood how it was working and the patch has been updated adding some comments. It still might be a useful to initialize the pins in `i2c_init` and it seems unnecessary to be writing a `0` to the output each time the `clear_s*` functions are called because the output state should always be `0`. The revised patch proposes some changes for these points.

Also tried a different approach of leaving the pins as open-drain outputs and just reading the inputs without re-configuring them as inputs and that worked too, but shall leave it as-is switching the pin configuration.

Tested on the bmp180 and also sht21.

Have some reservations about the approach used in the bmp180 driver, of creating a separate task, as the i2c driver does not appear the be task-safe so it might not support multiple drivers using this approach. The bmp180 driver also seems to support only one device as it keeps state in global variables. The i2c driver also appears to only supports one set of pins. Issues for another PR!

Some code for the sht2x is available at https://github.com/ourairquality/oaq-esp-open-rtos which might be possible to refactor out into the esp-open-rtos extras and a simple example.

kanflo:
Thanks for the PR @ourairquality and welcome to the project. I think this PR is a fine improvement of the initial i2c driver so I will go ahead and merge it. Adding the possibility to use several instances of the driver on different pins is a great idea.

PR #165, opened 2016-07-18
bmp180: make a lower level interface available and support oversampling.

ourairquality:
An application using multiple I2C devices will need it's own
loop. This reworks the code to make the detection, calibration
constant loading, and measurment functions available too without
having to use the bmp810 task which is still retained.

Adds support for oversampling.

Fixes a bug in the calculation of the temperature.

Better error handling. Checks for I2C errors and errors in the loading
of the calibration constants and propagates these up.

### DIFF
--- a/extras/bmp180/bmp180.h
+++ b/extras/bmp180/bmp180.h
@@ -52,4 +52,31 @@ void bmp180_trigger_pressure_measurement(const xQueueHandle* resultQueue);
 // Give the user the chance to create it's own handler
 extern bool (*bmp180_informUser)(const xQueueHandle* resultQueue, uint8_t cmd, bmp180_temp_t temperature, bmp180_press_t pressure);
 
+// Calibration constants
+typedef struct
+{
+    int16_t  AC1;
+    int16_t  AC2;
+    int16_t  AC3;
+    uint16_t AC4;
+    uint16_t AC5;
+    uint16_t AC6;
+
+    int16_t  B1;
+    int16_t  B2;
+
+    int16_t  MB;
+    int16_t  MC;
+    int16_t  MD;
+} bmp180_constants_t;
+
+// Returns true if the bmp180 is detected.
+bool bmp180_is_available();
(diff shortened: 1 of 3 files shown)


sheinz:
I think it is good to have a simple interface for accessing pressure/temperature. Just a function call.
Having a task and a queue is good but it's more like an application layer. An application might have a different approach and not need a task for fetching measurements. I believe that was @ourairquality motivation for this PR.

I would suggest separating driver and application layers. If necessary have a separate component that fetches measurements in a task and feed them to an application. But its more like an application logic. In my opinion example is a nice place for such logic.

projectgus:
I think this refactor looks generally good, but I agree with @sheinz that task structure is often better as an application-layer thing, or at least a higher layer than the device driver.

We might want to think about adding a really basic device driver interface that somehow allows you to structure a driver task? I don't know. For now this looks like a good chance to merge anyhow, as it moves what we already have in that general direction.

extras/bmp180/bmp180.h
```diff
@@ -52,4 +52,31 @@ void bmp180_trigger_pressure_measurement(const xQueueHandle* resultQueue);
...
+    int16_t  B1;
+    int16_t  B2;
+
+    int16_t  MB;
+    int16_t  MC;
+    int16_t  MD;
+} bmp180_constants_t;
+
+// Returns true if the bmp180 is detected.
+extern bool bmp180_is_available();
```
sheinz: You don't need extra **extern** for function declaration.

ourairquality:
@sheinz Fixed thanks. Also separated the task-and-queue api from the lower level api but still within the same file, leaving it as future work to consider removing the higher level api or moving it to an example.

sheinz:
I've verified low level API from this PR and it works fine for me.
I thinks it's good for merge.

kanflo:
Great @sheinz, then please go ahead with your new merge super powers :D

PR #166, opened 2016-07-18
DS3231 support

bhuvanchandra:

PR #167, opened 2016-07-19
DS3231 v2

bhuvanchandra:
Changes since #166  :

- Use enums instead of defines.
- Add license string in the ds3231 driver.

### DIFF
--- a/extras/ds3231/component.mk
+++ b/extras/ds3231/component.mk
@@ -1,0 +1,9 @@
+# Component makefile for extras/ds3231
+
+# expected anyone using bmp driver includes it as 'ds3231/ds3231.h'
+INC_DIRS += $(ds3231_ROOT)..
+
+# args for passing into compile rule generation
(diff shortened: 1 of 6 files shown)


extras/ds3231/component.mk
```diff
@@ -0,0 +1,9 @@
+# Component makefile for extras/ds3231
+
+# expected anyone using bmp driver includes it as 'ds3231/ds3231.h'
```
sheinz: **bmp** driver

sheinz:
I'm not sure if we follow any coding style here in EOR,  but using tabs for indentation is not recommended, at least for C programming.

UncleRus:
Coding style requirements are described in readme:
https://github.com/SuperHouse/esp-open-rtos#coding-style

extras/ds3231/component.mk
bhuvanchandra: I took the bmp driver dir as skeleton, missed updating the comment. Will fix this in next version.

PR #168, opened 2016-07-19
DS3231 v3

bhuvanchandra:
Changes since #167  :

    Fix typo in comment.

Changes since #166  :

    Use enums instead of defines.
    Add license string in the ds3231 driver.

### DIFF
--- a/extras/ds3231/ds3231.h
+++ b/extras/ds3231/ds3231.h
@@ -1,0 +1,187 @@
+/* Driver for DS3231 high precision RTC module
+ *
+ * Part of esp-open-rtos
+ * Copyright (C) 2015 Richard A Burton <richardaburton@gmail.com>
+ * Copyright (C) 2016 Bhuvanchandra DV <bhuvanchandra.dv@gmail.com>
+ * MIT Licensed as described in the file LICENSE
+*/
+
+#ifndef __DS3231_H__
+#define __DS3231_H__
+
+#include <stdint.h>
+#include <stdbool.h>
+#include <time.h>
+
+#define DS3231_ADDR            0x68
+
+#define DS3231_STAT_OSCILLATOR        0x80
+#define DS3231_STAT_32KHZ        0x08
+#define DS3231_STAT_BUSY        0x04
+#define DS3231_STAT_ALARM_2        0x02
+#define DS3231_STAT_ALARM_1        0x01
+
+#define DS3231_CTRL_OSCILLATOR        0x80
+#define DS3231_CTRL_SQUAREWAVE_BB    0x40
+#define DS3231_CTRL_TEMPCONV        0x20
+#define DS3231_CTRL_SQWAVE_4096HZ    0x10
+#define DS3231_CTRL_SQWAVE_1024HZ    0x08
+#define DS3231_CTRL_SQWAVE_8192HZ    0x18
+#define DS3231_CTRL_SQWAVE_1HZ        0x00
+#define DS3231_CTRL_ALARM_INTS        0x04
+#define DS3231_CTRL_ALARM2_INT        0x02
+#define DS3231_CTRL_ALARM1_INT        0x01
+
+#define DS3231_ALARM_WDAY        0x40
+#define DS3231_ALARM_NOTSET        0x80
+
+#define DS3231_ADDR_TIME        0x00
+#define DS3231_ADDR_ALARM1        0x07
+#define DS3231_ADDR_ALARM2        0x0b
+#define DS3231_ADDR_CONTROL        0x0e
+#define DS3231_ADDR_STATUS        0x0f
+#define DS3231_ADDR_AGING        0x10
+#define DS3231_ADDR_TEMP        0x11
+
+#define DS3231_12HOUR_FLAG        0x40
+#define DS3231_12HOUR_MASK        0x1f
+#define DS3231_PM_FLAG            0x20
+#define DS3231_MONTH_MASK        0x1f
+
+enum {
+    DS3231_SET = 0,
+    DS3231_CLEAR,
+    DS3231_REPLACE
+};
+
+enum {
+    DS3231_ALARM_NONE = 0,
+    DS3231_ALARM_1,
+    DS3231_ALARM_2,
+    DS3231_ALARM_BOTH
+};
+
+enum {
+    DS3231_ALARM1_EVERY_SECOND = 0,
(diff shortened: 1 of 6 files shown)


projectgus:
Hi @bhuvanchandra,

Thanks for these changes, looks good. The only other change, as mentioned in the other PR, is to change tabs to spaces (I missed this when looking at the original PR.) There are some coding guidelines here: https://github.com/SuperHouse/esp-open-rtos/#coding-style

Also, thanks for re-squashing your commits to keep the history clean & logical. You don't need to start a new PR each time, though. You can force push (git push -f) to the old PR branch and the PR will automatically update. I know it's considered a bad habit to force push, but for "your" PR branches it's OK. :)

Angus

bhuvanchandra:
Hi @projectgus,

Fixed the tab indents and force pushed the changes.

--
Bhuvan

extras/ds3231/ds3231.h
```diff
@@ -0,0 +1,99 @@
...
+
+enum {
+    DS3231_ALARM2_EVERY_MIN = 0,
+    DS3231_ALARM2_MATCH_MIN,
+    DS3231_ALARM2_MATCH_MINHOUR,
+    DS3231_ALARM2_MATCH_MINHOURDAY,
+    DS3231_ALARM2_MATCH_MINHOURDATE
+};
+
+bool ds3231_setTime(struct tm *time);
```
sheinz: You have all those nice function comments in C file. Don't you think header is a better place for them? As a header is the first place people look when figuring out how to use a library.

bhuvanchandra: Agreed, will update it.

bhuvanchandra:
Hi @projectgus,

Also pushed the changes as per @sheinz comments.

--
Bhuvan

ourairquality:
Started testing this and it looks ok so far. It seems fine to merge it.

One request is to change `ds3231_getTempInteger` to return a `uint16_t` and include all the 12 bits available rather than truncating to 8 bits. Then `ds3231_getTempFloat` could reuse the same code, or perhaps even be removed leaving it to the caller to convert to a float if they need to.

ourairquality:
The casting '(signed)data[0]' in `ds3231_getTempFloat` might not be correct. Here's a suggestion that also returns all 12 bits for the integer version which could be used for the float version too:

```
bool ds3231_getTempInteger(int16_t *temp)
{
    uint8_t data[2];

    data[0] = DS3231_ADDR_TEMP;
    if (ds3231_send(data, 1) && ds3231_recv(data, 2)) {
        *temp = (int16_t)(int8_t)data[0] << 2 | data[1] >> 6;
        return true;
    }
    return false;
}
```

sheinz:
I agree with @ourairquality that `(signed)data[0]` is not correct. But it will work nevertheless as `*temp` is `int8`.

With suggested by @ourairquality code I would suggest multiply the result by 25 and specify in the comment that function returns temperature in hundredth of a °C, for example 2525 is 25.25°C.

From the datasheet:
>Temperature is represented as a 10-bit code with a
resolution of 0.25°C and is accessible at location 11h and
12h. The temperature is encoded in two’s complement
format. The upper 8 bits, the integer portion, are at location
11h and the lower 2 bits, the fractional portion, are in
the upper nibble at location 12h. For example, 00011001
01b = +25.25°C. 

@bhuvanchandra Have you tested the negative temperature values? I'm not sure if the temperature in a register is a signed integer. It might just flip the sign bit without inverting the data bits. I haven't found information in the datasheet.

bhuvanchandra:
@ourairquality and @sheinz, thanks! for your feedback and suggestions.

@ourairquality, agreed, will update it accordingly.

@sheinz i didn't test at negative temperature, i guess it will flip the sign bit for negative values. The only clue we have is that the temperature value(10-bit) is encoded in 2's complement.

How about this snippet:
```
bool ds3231_getTemperature(int16_t *temp)
{
    uint8_t data[2];

    data[0] = DS3231_ADDR_TEMP;

    if (ds3231_send(data, 1) && ds3231_recv(data, 2)) {
        /* Check the sign bit */
        if (data[0] >> 7) {
            *temp = (int16_t) ((((data[0] & 0x7F) << 2) | ((data[1] & 0XC0) >> 6)) - 1024) * 25;
            return true;
        }

        *temp = (int16_t) ((data[0] << 2) | ((data[1] & 0XC0) >> 6)) * 25;
        return true;
    }
    return false;
}
```

ourairquality:
@sheinz Could an API that returns the binary fixed point representation generated by the device please be included as a minimum. Multiplying by 25 just frustrates most callers, causes premature overflow and loss of precision in subsequent integer computation. Personally I would not even include the `*25` and `float` API, but just request the binary fixed point API from which the other two variants can easily reuse.

@bhuvanchandra The encoding of the sign appears to be twos-complement (but I shall try to test) and the suggested code did work on some examples and is much simple. It's not necessary to mask out the low bits before the right shifts etc, lots of unnecessary code. I think the C standard guarantees that `int16_t` has a two's complement representation, but not certain, but it seems fine to me to assume all target devices use two's complement.

ourairquality:
Just to be sure on the negative temperatures, put it in the freezer and with the above suggested code it made a clean transition below zero and back above zero which is consistent with it being two's complement and supports my code suggestion above.

sheinz:
> Could an API that returns the binary fixed point representation generated by the device please be included as a minimum. Multiplying by 25 just frustrates most callers, causes premature overflow and loss of precision in subsequent integer computation. Personally I would not even include the *25 and float API, but just request the binary fixed point API from which the other two variants can easily reuse.

Then provide additional function, that might be named something like `ds3231_getRawTemperature` which would return pure data from the registers. 

In my opinion having those three function would cover all the users needs for the temperature:
   
    ds3231_getRawTemperature - will return pure register data(shifted)
    ds3231_getIntegerTemperature  -  will return only integer part of the temperature
    ds3231_getFloatTemperature
 
* names are just for reference, feel free to name it as you see appropriate

projectgus:
Supplying those three options sounds ideal to me. I can see different use cases for each.

bhuvanchandra:
@ourairquality yes, your code suggestion works nicely at negative temperatures also. I also tested it at my end today.
@sheinz, @projectgus, will add the three options.

--
Bhuvan

bhuvanchandra:
Pushed the changes.

kanflo:
Thanks @bhuvanchandra, this looks good so I'll go ahead and merge it.

PR #169, opened 2016-07-20
Upgrade to FreeRTOS v9.0.0

sasdf:
After we have `open_esplibs`, we are able to upgrade to latest FreeRTOS.  
Some function in libmain.a and libpp.a will call `xTaskCreateGeneric`, which has been removed in v9.0.0, so I wrote a proxy function to `xTaskCreate` in port.c. It can be remove after all reverse engineering works are done.   
Another difference is the type of parameter `pcName` in `xTaskCreate` and `xTimerCreate` is `const char * const` instead of `signed char *`.


projectgus:
\o/ nice work! Ping #80 where an upgrade was previously discussed.

Interesting that so few places in the code actually needed changing. I'm guessing the compatibility headers take care of a lot of it at compile time? There's probably clean up we can eventually do in port.c, etc., to update to the latest APIs.

Can you confirm if this line definitely didn't need changing: https://github.com/SuperHouse/esp-open-rtos/blob/master/open_esplibs/libmain/timers.c#L74 Though we should probably fix that regardless, given it's such a hack.

How much testing have you done? I'm interesting in merging in this, but as it's a big change it'd be ideal if a few people can put it through its paces before we pull it in.

rodrigopex:
Angus and sasdf what do you suggest? I can help on testing this.

Best regards,
Rodrigo Peixoto.

projectgus:
@rodrigopex , great! @sasdf  may have some other suggestions, but for now I'd suggest testing this as much as you can and letting us know what you find.

sasdf:
@projectgus , I have no idea about the line in timer.c, sorry.
I haven't done any test, but it works for my project.
Basic things like tasks, semaphore, queue, notification, uart ,and socket works fine.
@rodrigopex , Thanks for your help!

ourairquality:
Gave this a test with the wificfg and an app I am working on and it seems to be running ok. Not extensive testing but it seems to run ok, and tested the AP and Station modes. A few changes were needed:

* some more uses of `xTaskCreate()` in the examples needed updating to cast the task name to `(const char * const)`.

* In FreeRTOS.h need to change to `#include <FreeRTOSConfig.h>` to get the per-project FreeRTOSConfig.h override file included first.

Is that enough confidence to merge this now.

Did anyone else have any freeback?

sheinz:
Thanks @sasdf, @ourairquality and everybody who helped testing this. The PR is merged with https://github.com/SuperHouse/esp-open-rtos/pull/245

PR #171, opened 2016-07-22
libmain: add sdk_system_uart_swap and sdk_system_uart_de_swap

ourairquality:
Could see system_uart_de_swap define in the header file but not in libmain, but added it anyway.

### DIFF
--- a/open_esplibs/libmain/user_interface.c
+++ b/open_esplibs/libmain/user_interface.c
@@ -558,4 +558,38 @@ enum sdk_dhcp_status sdk_wifi_station_dhcpc_status(void) {
     return sdk_dhcpc_flag;
 }
 
+void sdk_system_uart_swap()
+{
+    while (FIELD2VAL(UART_STATUS_TXFIFO_COUNT, UART(0).STATUS)) {};
+    while (FIELD2VAL(UART_STATUS_TXFIFO_COUNT, UART(1).STATUS)) {};
+
+    /* Disable pullup IO_MUX_MTDO, Alt TX. GPIO15. */
+    iomux_set_pullup_flags(3, 0);
+    /* IO_MUX_MTDO to function UART0_RTS. */
+    iomux_set_function(3, IOMUX_GPIO15_FUNC_UART0_RTS);
+    /* Enable pullup MUX_MTCK. Alt RX. GPIO13. */
+    iomux_set_pullup_flags(1, IOMUX_PIN_PULLUP);
+    /* IO_MUX_MTCK to function UART0_CTS. */
+    iomux_set_function(1, IOMUX_GPIO13_FUNC_UART0_CTS);
+
+    DPORT.PERI_IO |= DPORT_PERI_IO_SWAP_UART0_PINS;
+}
+
+void sdk_system_uart_de_swap()
+{
+    while (FIELD2VAL(UART_STATUS_TXFIFO_COUNT, UART(0).STATUS)) {};
+    while (FIELD2VAL(UART_STATUS_TXFIFO_COUNT, UART(1).STATUS)) {};
+
+    /* Disable pullup IO_MUX_U0TXD, TX. GPIO 1. */
+    iomux_set_pullup_flags(5, 0);
+    /* IO_MUX_U0TXD to function UART0_TXD. */
+    iomux_set_function(5, IOMUX_GPIO1_FUNC_UART0_TXD);
+    /* Enable pullup IO_MUX_U0RXD. RX. GPIO 3. */
+    iomux_set_pullup_flags(4, IOMUX_PIN_PULLUP);
+    /* IO_MUX_U0RXD to function UART0_RXD. */
+    iomux_set_function(4, IOMUX_GPIO3_FUNC_UART0_RXD);
+
+    DPORT.PERI_IO &= ~DPORT_PERI_IO_SWAP_UART0_PINS;
+}
+
 #endif /* OPEN_LIBMAIN_USER_INTERFACE */


foogod:
`system_uart_swap` is actually one of a large hunk of functions in user_interface.c which I've been working on finishing up decompiling but haven't yet made it into the mainline..  I haven't had a lot of time lately but I'm gonna see if I can get some of that finished up so we can work on merging it in (but I'm not opposed to adding this implementation in the meantime)

Interestingly, `system_uart_de_swap` exists in the header file, but apparently does not exist at all in the 0.9.9 version of the libraries (it was apparently added to later versions, though).  Given that this is apparently an oversight on Espressif's part, I'll agree it does seem reasonable to add it here even though it's not in the binary version of the library..

projectgus:
@ourairquality, this looks like a good addition. Thanks.

The only request I have is, could you please use the constants defined in iomux_regs.h for the "function" numbering instead of plain numbers? If it's also possible to also use the GPIO constants in some way to avoid the plain numbers for the IOMUX numbering then that would also be great, although we might need to make some tweaks to iomux.h & iomux_regs.h for that (when I've worked with it, it seems a little clunky at the moment.) Feel free to take a run at that if you want, or not if you don't. :)

ourairquality:
@projectgus I would like to have used some defines, but I can't see any that are usable. For example, `IOMUX_GPIO1_FUNC_UART0_TXD` is not a function number that can be passed to `iomux_set_function` and `IOMUX_GPIO1` is not a pin number that can be passed to `iomux_set_function`.  What could be done is to use `gpio_to_iomux` to at least make it clearer which GPIO pin numbers were being modified? Other than that it looks like a new set of defines are needed here?

projectgus:
@ourairquality , yeah you're right. I remembered it was a bit awkward, but not quite that awkward!

The philosophy between the xxx_regs.h/xxx.h headers for peripherals is to have a registers-only interface and then a compatible wrapper around that interface. However in this case they're not really cross-compatible. I'd be open to changes that did make them cross-compatible, though.

You can use the register interface entirely, but you lose the fact that iomux_set_function() masks out the non-function-related pins in the iomux so you can keep those functional.

Hmm, yeah. I think it must be possible to refactor these functions to be more harmonious. Just probably fiddly as well! Any suggestions welcome, otherwise I'm OK with merging this as-is and I'll open a separate Issue to track the problem of the iomux/iomux_reg incompatibility.

ourairquality:
@projectgus The iomux pin functions have been split out in the header file in an attempt to move this along. Would appreciate it if someone else with a good command of what they want could rework the header file in a follow up PR.

sheinz:
I would prefer to have a single definition for GPIO_FUNC*. Having both GPIO0_FUNC_SPI0_CS2 and IOMUX_GPIO0_FUNC_SPI0_CS2 might be confusing. User might use one in place of another. 

As I suggested in [Issue#179](https://github.com/SuperHouse/esp-open-rtos/issues/179) the solution might be to just remove *IOMUX_FUNC* in `iomux_set_function`, so the current definitions of *IOMUX_GPIOn_FUNC** would fit. (*iomux_set_gpio_func* should be adjusted as well)

But another point that I don't like is the first parameter *iomux_num*:
`inline static void iomux_set_function(uint8_t iomux_num, uint32_t func)`

When I first used function **iomux_set_function** I wrote:
`iomux_set_function(IOMUX_GPIO3, IOMUX_GPIO3_FUNC_I2SO_DATA);`
Which is obviously wrong but at the first glance of header files it seemed right to me.

May be we should rename the function (and change its implementation) to something like:
inline static void **gpio**_set_function(uint8_t **gpio_num**, uint32_t **iomux_func**)
And use it like:
`gpio_set_function(3, IOMUX_GPIO3_FUNC_I2SO_DATA);`

projectgus:
This seems like a good change, although I'd suggest naming such a function gpio_set_iomux_function() to be more explicit how we're changing iomux not just the gpio peripheral.

I think the reason I shied away from this originally was performance (either you do ugly inlining tricks or you have to have a look up table in code), but that's probably a premature optimisation - noone's going to be calling this in a tight loop.

sheinz:
The functions might look like:
```
/**
 * Set IOMUX function.
 *
 * @param iomux_num Index of IOMUX register. Can be converted from GPIO number
 * with gpio_to_iomux.
 * @param iomux_func GPIO function definition IOMUX_GPIOn_FUNC_* 
 */
inline static void iomux_set_function(uint8_t iomux_num, uint32_t iomux_func)
{
    uint32_t prev = IOMUX.PIN[iomux_num] & ~IOMUX_PIN_FUNC_MASK;
    IOMUX.PIN[iomux_num] = func | prev;
}
/**
 * Set GPIO pin function by changing corresponding IOMUX register.
 *
 * @param gpio_num GPIO pin number: 0-15
 * @param iomux_func GPIO function definition IOMUX_GPIOn_FUNC_* 
 */
inline static void gpio_set_iomux_function(uint8_t gpio_num, uint32_t iomux_func)
{
    iomux_set_function(gpio_to_iomux(gpio_num, iomux_func));
}
```
And in *iomux_set_gpio_function*:
`    iomux_set_function(iomux_num, IOMUX_FUNC(func));`

But I still don't like to expose functions with **iomux_num** parameter to a user as it is not obvious what value to pass there. It would be better to have a set of functions visible to a user that accept only GPIO number.
Naming is still a little bit awkward as we have two different functions with almost identical names: **iomux_set_gpio_function** and **gpio_set_iomux_function**.

sheinz:
Another suggestion is to move `gpio_set_iomux_function` to *gpio.h* file. As it is more high level GPIO manipulation.

A user might look there first when he needs to figure out how to work with GPIO and never needs to look into *iomux.h* for low level IOMUX register manipulation. The user still needs to look into *iomux_regs.h* though for a parameter to pass into `gpio_set_iomux_function`.

ourairquality:
Updated to use the definitions made usable in https://github.com/SuperHouse/esp-open-rtos/pull/187 PTAL

sheinz:
I checked the functions on my device. They work as expected. In my opinion the PR is good for merging.

kanflo:
Thanks all, I'll merge this one.

PR #178, opened 2016-08-08
SPIFFS: Selectable configuration of SPIFFS

sheinz:
In [PR#163](https://github.com/SuperHouse/esp-open-rtos/pull/163) was suggested to have ability to select how SPIFFS is configured, as a singleton during compile-time or during run-time. This PR adds this feature.

By default SPIFFS is configured as before at compile-time. If run-time configuration is needed `SPIFFS_SINGLETON=0` should be specified in program's Makefile. In that case `esp_spiffs_init` accepts two arguments: address and size of spiffs flash region.

Logical page size, block size and erase block size is still a compile time defined parameters. I don't see a use case when those parameters needs to be changed in run-time. If needed they can be redefined in program's Makefile as well.

### DIFF
--- a/examples/spiffs/spiffs_example.c
+++ b/examples/spiffs/spiffs_example.c
@@ -10,6 +10,14 @@
 #include "spiffs.h"
 #include "esp_spiffs.h"
 
+/**
+ * This example shows the default SPIFFS configuration when SPIFFS is
+ * configured in compile-time (SPIFFS_SINGLETON = 1).
+ *
+ * To configure SPIFFS in run-time uncomment SPIFFS_SINGLETON in the Makefile
+ * and replace the commented esp_spiffs_init in the code below.
+ *
+ */
 
 static void example_read_file_posix()
 {
@@ -76,7 +84,13 @@ static void example_fs_info()
 
 void test_task(void *pvParameters)
 {
+#if SPIFFS_SINGLETON == 1
     esp_spiffs_init();
+#else
+    // for run-time configuration when SPIFFS_SINGLETON = 0
+    esp_spiffs_init(0x200000, 0x10000);
+#endif
+
     if (esp_spiffs_mount() != SPIFFS_OK) {
         printf("Error mount SPIFFS\n");
     }
--- a/extras/spiffs/component.mk
+++ b/extras/spiffs/component.mk
@@ -1,7 +1,28 @@
 # Component makefile for extras/spiffs
 
+# If spiffs is configured as SINGLETON it must be configured in compile time.
+SPIFFS_SINGLETON ?= 1
+
 SPIFFS_BASE_ADDR ?= 0x300000
 SPIFFS_SIZE ?= 0x100000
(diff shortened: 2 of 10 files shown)


examples/spiffs/spiffs_example.c
```diff
@@ -77,6 +85,8 @@ static void example_fs_info()
 void test_task(void *pvParameters)
 {
     esp_spiffs_init();
+    // esp_spiffs_init(0x200000, 0x1000);   // for run-time configuration
```
kanflo: Could this not be surrounded by ```#if SPIFFS_SINGLETON == 1``` or perhaps the comment could refer to the singleton configuration.

extras/spiffs/mkspiffs/README.md
```diff
@@ -14,21 +14,23 @@ $(eval $(call make_spiffs_image,files))
...
-```
-make SPIFFS_SIZE=0x100000
-```
+To manually generate SPIFFS image from a directory SPIFFS configuration must be
+provided as command line arguments.
 
-mkspiffs cannot be built without specifying SPIFFS size because it uses the
-same SPIFFS sources as the firmware. And for the firmware SPIFFS size is
-compile time defined.
+Arguments:
```
kanflo: Nice!

kanflo:
Good addition. I only have a small comment, otherwise this one is good for merging.

examples/spiffs/spiffs_example.c
sheinz: I thought about doing it at first but for some reason rejected it. 
I will surround it with the preprocessor condition.

extras/spiffs/component.mk
```diff
@@ -1,7 +1,30 @@
 # Component makefile for extras/spiffs
 
+# If spiffs is configured as SINGLETON it must be configured in compile time.
+SPIFFS_SINGLETON ?= 1
+
 SPIFFS_BASE_ADDR ?= 0x300000
 SPIFFS_SIZE ?= 0x100000
+SPIFFS_LOG_PAGE_SIZE ?= 256
+SPIFFS_LOG_BLOCK_SIZE ?= 8192
+SPIFFS_ERASE_BLOCK_SIZE ?= 4096
```
pellepl: I would actually prefer not having ERASE_BLOCK_SIZE configurable.  As far as i know the esp hal only supports 4k erases.

sheinz: Yes, you are correct, only 4k erase is supported.

pellepl: Then i would recommend hard coding SPIFFS_ERASE_BLOCK_SIZE. If any user
messes with that setting, the file system will inherently fail.

Den 8 aug. 2016 21:10 skrev "sheinz" <notifications@github.com>:

sheinz:
@pellepl I've removed SPIFFS_BLOCK_ERASE_SIZE option from Makefile and from mkspiffs utility.
Pushed the changes.
Thanks for the suggestion.

PR #180, opened 2016-08-09
Sysparam by @foogod

kanflo:
As @foogod is restricted on time I got his permission to create a PR for his excellent sysparam framework. I have rebased his branch, squashed his commits and added licensing information as well as fixed a few compiler warnings.

### DIFF
--- a/core/app_main.c
+++ b/core/app_main.c
@@ -27,7 +27,7 @@
 #include "espressif/esp_common.h"
 #include "espressif/phy_info.h"
 #include "sdk_internal.h"
-#include "esplibs/libmain.h"
+#include "sysparam.h"
 
 /* This is not declared in any header file (but arguably should be) */
 
@@ -141,6 +141,8 @@ void IRAM sdk_user_start(void) {
     uint32_t cksum_len;
     uint32_t cksum_value;
     uint32_t ic_flash_addr;
+    uint32_t sysparam_addr;
+    sysparam_status_t status;
 
     SPI(0).USER0 |= SPI_USER0_CS_SETUP;
     sdk_SPIRead(0, buf32, 4);
@@ -205,6 +207,20 @@ void IRAM sdk_user_start(void) {
         //FIXME: should we halt here? (original SDK code doesn't)
     }
     memcpy(&sdk_g_ic.s, buf32, sizeof(struct sdk_g_ic_saved_st));
+
+    // By default, put the sysparam region just below the config sectors at the
+    // top of the flash space
+    sysparam_addr = flash_size - (4 + DEFAULT_SYSPARAM_SECTORS) * sdk_flashchip.sector_size;
+    status = sysparam_init(sysparam_addr, flash_size);
+    if (status == SYSPARAM_NOTFOUND) {
+        status = sysparam_create_area(sysparam_addr, DEFAULT_SYSPARAM_SECTORS, false);
+        if (status == SYSPARAM_OK) {
+            status = sysparam_init(sysparam_addr, 0);
+        }
+    }
+    if (status != SYSPARAM_OK) {
+        printf("WARNING: Could not initialize sysparams (%d)!\n", status);
+    }
 
     user_start_phase2();
 }
--- a/core/app_main.c
+++ b/core/app_main.c
@@ -27,6 +27,7 @@
 #include "espressif/esp_common.h"
 #include "espressif/phy_info.h"
 #include "sdk_internal.h"
+#include "esplibs/libmain.h"
 #include "sysparam.h"
 
 /* This is not declared in any header file (but arguably should be) */
--- a/core/include/esp/types.h
+++ b/core/include/esp/types.h
@@ -3,6 +3,7 @@
 
 #include <stdint.h>
 #include <stdbool.h>
+#include <stddef.h>
 
 typedef volatile uint32_t *esp_reg_t;
 
--- a/core/include/sysparam.h
+++ b/core/include/sysparam.h
@@ -1,0 +1,413 @@
+#ifndef _SYSPARAM_H_
+#define _SYSPARAM_H_
+
+#include <esp/types.h>
+
+#ifndef DEFAULT_SYSPARAM_SECTORS
+#define DEFAULT_SYSPARAM_SECTORS 4
+#endif
+
+/** @file sysparam.h
+ *
+ *  Read/write "system parameters" to persistent flash.
+ *
+ *  System parameters are stored as key/value pairs.  Keys are string values
+ *  between 1 and 255 characters long.  Values can be any data up to 255 bytes
+ *  in length (but are most commonly also text strings).  Up to 126 key/value
+ *  pairs can be stored at a time.
+ *
+ *  Keys and values are stored in flash using a progressive list structure
+ *  which allows space-efficient storage and minimizes flash erase cycles,
+ *  improving write speed and increasing the lifespan of the flash memory.
+ */
+
+/** Status codes returned by all sysparam functions
+ *
+ *  Error codes (`SYSPARAM_ERR_*`) all have values less than zero, and can be
+ *  returned by any function.  Values greater than zero are non-error status
+ *  codes which may be returned by some functions to indicate various results.
+ */
+typedef enum {
+    SYSPARAM_OK           = 0,  ///< Success
+    SYSPARAM_NOTFOUND     = 1,  ///< Entry not found matching criteria
+    SYSPARAM_PARSEFAILED  = 2,  ///< Unable to parse retrieved value
+    SYSPARAM_ERR_NOINIT   = -1, ///< sysparam_init() must be called first
+    SYSPARAM_ERR_BADVALUE = -2, ///< One or more arguments were invalid
+    SYSPARAM_ERR_FULL     = -3, ///< No space left in sysparam area (or too many keys in use)
+    SYSPARAM_ERR_IO       = -4, ///< I/O error reading/writing flash
+    SYSPARAM_ERR_CORRUPT  = -5, ///< Sysparam region has bad/corrupted data
+    SYSPARAM_ERR_NOMEM    = -6, ///< Unable to allocate memory
+} sysparam_status_t;
+
+/** Structure used by sysparam_iter_next() to keep track of its current state
+ * and return its results.  This should be initialized by calling
+ * sysparam_iter_start() and cleaned up afterward by calling
+ * sysparam_iter_end().
+ */
+typedef struct {
+    char *key;
+    uint8_t *value;
+    size_t key_len;
+    size_t value_len;
+    bool binary;
+    size_t bufsize;
+    struct sysparam_context *ctx;
+} sysparam_iter_t;
+
+/** Initialize sysparam and set up the current area of flash to use.
+ *
+ *  This must be called (and return successfully) before any other sysparam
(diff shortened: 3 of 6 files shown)


sheinz:
I haven't looked it through but at the first glance it seems like a really quality piece of software.
I will try to review it more thoroughly later.

For now I only have one comment. Do we really need it inside *core* and be available unconditionally? The user might not need it but it will still be compiled, linked and checking sysparam region at system start.
Would it be possible to include it as the rest of the add-ons in *extras* directory? So a user might turn it on as needed.

kanflo:
You have a point. For my part though, I think sysparam is an essential part of EOR and therefore should be included in core.

foogod:
A large part of the original reason I developed this was to have a consistent mechanism to configure system-related parameters (hence the name), such as boot options, WiFi parameters, timezone settings, etc.  While those bits of integration haven't been done yet, most of them will require it to be part of core, which is why it was originally written that way.

ourairquality:
This is interesting but it looks too heavy weight for a required core module.

* Thread safety does not appear to be addressed. If used in the core then it might be entered in unexpected paths, be a deadlock hazard etc. Core sysparam needs to be light weight.

* It support use cases such as frequently updating parameters that are not needed in many applications, and perhaps not even in the core. These use cases might make it interesting for 'extras' as a parameter database.

* It's not clear that it handles failing sectors. It tries to zero the index, but what if the a length is wrong then it can not skip an entry. For infrequently updated parameters, a sector level 'clean' flag might be more robust.

Thus I suggest this be an 'extras' component and that core sysparams use something very simple such as fixed sized fields and compile-time defined offsets, and sector level redundancy. For thread safety perhaps having just one old sector is adequate safety, assuming updates do not occur in quick succession, so that readers use the current sector while a new sector is written and then flags are changed to switch new readers to the latest sector while existing reads can complete reading from the old section, with no locking except for writers.

foogod:
Sigh..  So I'm realizing that folks here may not have been around when I was discussing a lot of this on the maillist a couple of months ago, so for context, here's the contents of the last update message I sent out there (back in May):

> A new version of the sysparam stuff is up in the 'sysparam' branch of my repo (https://github.com/foogod/esp-open-rtos/tree/sysparam).  I think this one is actually pretty close to being ready for prime time, but would love to hear any input folks have before I put together a PR for it.
>
> Changes from the previous version:
> - Atomic updates:  Even if the system loses power or otherwise fails in the middle of a sysparam write, no corruption of the flash data should occur.
> - Fully validated writes (data is written, then read back and checked before being committed)
> - Keys and values can now be up to 65535 bytes (previously 255) in length (provided you create a sysparam area big enough to put them in)
> - Up to 4095 different keys can be stored at the same time (previously 126)
> - Key/value overhead is still the same as before (4 bytes each)
> - Values can be stored as "text" (default) or "binary" to indicate non-printable data (useful to know when printing/dumping arbitrary key/values)
> - It is now possible for sysparam_init() to identify and use a sysparam region even if it was created with a different size/position than are expected (as long as it's not too far away from where it's expecting to find it).  For example, if the code is expecting a sysparam region that's two sectors in size, but there was one created (by a previous firmware, etc) that's four sectors in size in roughly the same location, it will find and continue to use the existing region with no problem.
> - sysparam initialization has been added to system startup (no need to call sysparam_init() from the app anymore)
>
> Outstanding issues:
> - Sysparam reads should be fully MT-safe (even writing and reading at the same time should be safe, due to the atomic-update nature of the system), but multiple concurrent writes are not MT-safe and can potentially cause corruption.  It would be good to put some synchronization around this to make it fully threadsafe.
> - There is also a potential race condition such that if a read call is issued from one task, while another task's write triggers a compaction, then a deletion occurs, then enough additional data is written such that another write causes yet *another* compaction, all before the other task's read completes (that is two successive compactions happening within the span of executing a single read call), the read may erroneously fail or return bad data.  Unfortunately, I think efficiently protecting against that scenario will actually require inventing some locking constructs that do not currently exist in FreeRTOS (luckily, it seems fairly unlikely to happen in practice)..

Regarding thread-safety, as I mentioned, reads are already completely MT-safe (and realistically I think that's all that should be required to address your concerns about concurrent use within core system routines).  Writes currently are not, but it probably would not be too hard to make it so, I just haven't had time to work on any of this stuff for quite a while (so if somebody else wants to tackle it, feel free).

Note also that the read routines and writing code are fairly independent and you can use the read portion without needing to include any of the writing code in the linked result.  If the core only reads values and does not write them, then any app which does not use the writing capability will not pull in the writing code at all, so it ultimately is not as heavy as it looks for most apps.

Regarding handling failing sectors, currently absolutely nothing in esp-open-rtos (or Espressif's SDK) really handles failing sectors very well either, so I think that's kinda an unreasonable requirement.  Please note that using sysparam has a much much lower wear footprint than the sort of simplistic mechanisms currently used (and which you advocate continuing to use), so sector failure is actually far less likely to occur in the first place.  Considering that with the current design of the OS in pretty much all cases if sectors start failing your module is pretty much hosed anyway, and considering that sysparam *can* actually detect and recover from a large variety of flash write failures already, I'd say sysparam is ultimately a substantial improvement in this regard.

Frankly, if this isn't a core component, I think it loses a substantial amount of its potential value, and I've basically wasted my time writing it.  Is it absolutely perfect?  No.  Is it a hell of a lot better in almost every way than what we have now, and does it have the potential to add a huge benefit to both core functionality and most user apps?  I believe so.  That, in my opinion should be plenty of reason to put it in the core.  (Honestly, given the code quality of most of the other stuff that's in there currently, I feel the level of perfection people seem to be demanding here is a bit unreasonable, too.)

I joined this project because I wanted to help make something that was substantially better than Espressif's offerings, not just a not-quite-identical clone of an old version of some incomplete stuff somebody else actually wrote most of.  I am beginning to wonder a bit if I'm the only one...

ourairquality:
Took another good look over the code and understand now that it supports multiple readers and a single writer, thank you.

I recognize that your code is designed to reduce erase cycles but it does not really seem to handle bad sectors, and perhaps a note on this could be added just so code does not depend on it for frequent updates.

Perhaps it could also be noted that writes need to be synchronised and that this is a burden for the application code. But if this is necessary then is there any problem adding a mutual exclusion semaphore for the sysparam code now and using it in the write paths? 

Can the double-compaction-while-reading failure be triggered by a single write, or does it need two separate writes? Can an application protect against this somewhat with a yield-delay after each write, and could this be built into the code now?

Guess we need a multi-reader-single-writer lock here, and if freertos does not have this then can a simpler reader count be used protected by the same semaphore as used by the write paths. The readers could check this on entry and if negative just spin waiting for the writer to finish or otherwise increase it, and decreases it on exit. The writers would spin waiting for it to be zero and decrease it until done or just hold the lock.

kanflo:
As I mentioned before I agree with @foogod that this needs to be part of core. Right now about the only "setting" in a typical EOR application is the wifi credentials and that needs to be moved to sysparam if one wants to do anything more with EOR that one off demo applications (I sure do :). Among other the things that need to be stored "out of bounds" of the application we have wifi mode (AP/client), AP name, node id and name, node private key, IP/port of FOTA server and optional core dump server. I nailed it down to about 30 parameters needing to be stored in sysparam.

MT needs to be fixed (with one big semaphore to start with if nothing else) and IMHO we are very much better off with this framework than without it. It marks an important step towards building what I consider "real devices" using EOR.

ourairquality:
Ok, I suggest merging it and patch in thread safety later. I'd like this sorted too.

ourairquality:
Gave this a quick try. Used a constant string for the key, and the writes failed, writing all zeros. Added a quick hack buffer in _do_write and wrote to the flash from this buffer and it worked. Could it be that the key string is stored in the flash and that is the problem?

Also tried copying the key string to a buffer in the caller, but this failed too, perhaps gcc being too smart. Guessing this could be a problem for the data too. Do all writes need to be buffered.

ourairquality:
Here is a suggestion. There was already a buffer allocated for verifying writes which can be used as a write buffer too. If this PR could be landed it would be easier to submit patches.

```
static inline IRAM sysparam_status_t _do_write(uint32_t addr, const void *data, void *buffer, size_t data_size) {
    memcpy(buffer, data, data_size);
    CHECK_FLASH_OP(sdk_spi_flash_write(addr, buffer, data_size));
    return SYSPARAM_OK;
}

static IRAM sysparam_status_t _write_and_verify(uint32_t addr, const void *data, size_t data_size) {
    int i;
    size_t count;
    sysparam_status_t status = SYSPARAM_OK;
    uint8_t *buf = malloc(VERIFY_BUF_SIZE);

    if (!buf) return SYSPARAM_ERR_NOMEM;
    for (i = 0; i < data_size; i += VERIFY_BUF_SIZE) {
        count = min(data_size - i, VERIFY_BUF_SIZE);
        status = _do_write(addr + i, data + i, buf, count);
        if (status != SYSPARAM_OK) break;
        status = _do_verify(addr + i, data + i, buf, count);
        if (status != SYSPARAM_OK) {
            debug(1, "Flash write (@ 0x%08x) verify failed!", addr);
            break;
        }
    }
    free(buf);
    return status;
}
```

kanflo:
Thanks @ourairquality. I am merging this one so we can start improving it.

foogod:
Sorry I didn't have a chance to respond earlier, but regarding a couple of the above questions:

> Can the double-compaction-while-reading failure be triggered by a single write, or does it need two separate writes?

It actually requires a series of several very specific things to occur:

- A write must occur which triggers a compaction
- The compaction must actually free enough space for the write to be successful
- There must subsequently be at least one delete which frees additional potential space
- There must be another write which triggers another compaction
- The second compaction must actually free enough space for the second write to be successful.

So it actually requires (at the very least) a write-delete-write combination, all within the span of a single read in another thread.  What's more, these write and delete operations must be of just the right sizes to trigger two successful compactions in that period.  (Frankly, this all seems unlikely enough in practice that I think it's not worth spending a huge amount of effort on fixing it.. there are far worse issues with some other existing ESP8266 functionality which would probably be a better use of people's time to fix).

> Can an application protect against this somewhat with a yield-delay after each write, and could this be built into the code now?

I suppose that's possible, but frankly the only way I see this as likely to happen in the first place is if the write task has higher priority than the reader, and in that case I'm pretty sure taskYIELD will pick the highest-priority available task, which would be the write task again, so I doubt it would actually achieve anything.

> Guess we need a multi-reader-single-writer lock here, and if freertos does not have this

Yup, and it doesn't (I checked)

> then can a simpler reader count be used protected by the same semaphore as used by the write paths. The readers could check this on entry and if negative just spin waiting for the writer to finish or otherwise increase it, and decreases it on exit. The writers would spin waiting for it to be zero and decrease it until done or just hold the lock.

Maybe, though frankly what you're describing there only needs a little more logic to be a full-fledged (proper) reader-writer lock anyway, so if we're gonna spend the effort to do this I'd recommend just implementing a correct reader-writer lock, which could then potentially be used by apps, etc, as well.

> Gave this a quick try. Used a constant string for the key, and the writes failed, writing all zeros. Added a quick hack buffer in _do_write and wrote to the flash from this buffer and it worked. Could it be that the key string is stored in the flash and that is the problem?

Hmm, you know what, when I was first developing this, it was before we had actually implemented the "move constant strings to flash" change, which is probably why I never encountered that issue.. it's quite possible now that I think about it that trying to use strings stored in flash would fail in this way.  Yeah, that will definitely need to be fixed.

PR #185, opened 2016-08-14
spiffs: enable temporal cache for spiffs v 0.3.6

pellepl:
In spiffs 0.3.6, it is recommended to enable the temporal fd cache feature. This optimizes the opening of frequently opened files. The cost is 6 extra bytes per file descriptor.


sheinz:
Thanks @pellepl for the PR.
I looked into spiffs wiki for this feature details. Really impressive numbers for the esp8266 web server example. And as a web server is the most common use case for spiffs in esp8266 it is good to have the temporal cache enabled.

On my hardware everything works fine, so I will merge it right away.

PR #186, opened 2016-08-15
BME280 support.

ourairquality:
The BME280 is close to compatible with the existing BMP280 and extends it with support for measuring humidity, so support has been bundled into the bmp280 driver.

The example now auto-detects the device and displays the humidity for the BME280.

The I2C bus initialization has been moved out of the bmp280 driver to support multiple devices.

The check-id and reset logic has been bundled into the driver initialization. It needs to be re-initialized after reset anyway and the chip-id is need to initialize it, just re-initialize to reset.

Support has been added for multiple devices. The calibration data storage needs to be managed by the caller rather than static data. The caller can choose the I2C address to allow two BMx280 devices to be used on the same I2C bus.

An interface has been added to return the measurement values in an integer fixed float format. The float format interface is still there.

All the values are read in one I2C transaction to ensure they are a consistent set.

### DIFF
--- a/examples/bmp280/bmp280_example.c
+++ b/examples/bmp280/bmp280_example.c
@@ -6,43 +6,55 @@
 #include "FreeRTOS.h"
 #include "task.h"
 
+#include "i2c/i2c.h"
 #include "bmp280/bmp280.h"
 
 // In forced mode user initiate measurement each time.
 // In normal mode measurement is done continuously with specified standby time. 
 // #define MODE_FORCED
 
-const uint8_t scl_pin = 5;
-const uint8_t sda_pin = 4;
+const uint8_t scl_pin = 0;
+const uint8_t sda_pin = 2;
 
 #ifdef MODE_FORCED
 static void bmp280_task_forced(void *pvParameters)
 {
     bmp280_params_t  params;
-    float pressure, temperature;
+    float pressure, temperature, humidity;
 
     bmp280_init_default_params(&params);
     params.mode = BMP280_MODE_FORCED;
 
+    bmp280_t bmp280_dev;
+    bmp280_dev.i2c_addr = BMP280_I2C_ADDRESS_0;
+
     while (1) {
-        while (!bmp280_init(&params, scl_pin, sda_pin)) {
+        while (!bmp280_init(&bmp280_dev, &params)) {
             printf("BMP280 initialization failed\n");
             vTaskDelay(1000 / portTICK_RATE_MS);
         }
 
+        bool bme280p = bmp280_dev.id == BME280_CHIP_ID;
+        printf("BMP280: found %s\n", bme280p ? "BME280" : "BMP280");
+
         while(1) {
             vTaskDelay(1000 / portTICK_RATE_MS);
-            if (!bmp280_force_measurement()) {
+            if (!bmp280_force_measurement(&bmp280_dev)) {
                 printf("Failed initiating measurement\n");
                 break;
             }
-            while (bmp280_is_measuring()) {}; // wait for measurement to complete
+            // wait for measurement to complete
+            while (bmp280_is_measuring(&bmp280_dev)) {};
 
-            if (!bmp280_read(&temperature, &pressure)) {
+            if (!bmp280_read_float(&bmp280_dev, &temperature, &pressure, &humidity)) {
                 printf("Temperature/pressure reading failed\n");
                 break;
             }
-            printf("Pressure: %.2f Pa, Temperature: %.2f C\n", pressure, temperature);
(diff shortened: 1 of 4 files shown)


sheinz:
Hi @ourairquality,

Thanks for the PR. I actually ordered a BME280 module and thought about extending the driver with the support myself :) but haven't received the module yet.

I like the ability to have multiple devices on the same i2c bus!

examples/bmp280/bmp280_example.c
```diff
@@ -6,66 +6,89 @@
...
 {
     bmp280_params_t  params;
-    float pressure, temperature;
+    float pressure, temperature, humidity;
 
     bmp280_init_default_params(&params);
     params.mode = BMP280_MODE_FORCED;
 
+    bmp280_calib_t bmp280_calib;
+    bmp280_calib.i2c_addr = BMP280_I2C_ADDRESS_0;
```
sheinz: `bmp280_calib_t` now becomes something like a class instance in OOP terms. It's no longer just a calibration data. It might be better called something like `bmp280_t` and have all the instance variables inside including calibration data. And I don't really like to expose calibration data to a user as the user doesn't care about it, so it's better to make it "private". But it's a cosmetic comments and might be a separate PR.

extras/bmp280/bmp280.c
```diff
@@ -46,116 +46,155 @@
...
     return false;
 }
 
-static bool check_chip_id()
+static bool read_calibration_data(bmp280_calib_t *calib_data)
 {
-    return (read_register8(BMP280_REG_ID)==BMP280_CHIP_ID);
+    uint8_t i2c_addr = calib_data->i2c_addr;
+
+    if (read_register16(i2c_addr, 0x88, &calib_data->dig_T1) &&
```
sheinz: I would still prefer to read calibration data as a one block of data into "packed" struct. At least pressure calibration data as those registers goes one after another. Unfortunately it can't be done for humidity calibration data.

ourairquality: Yes, it's not possible, some of the constants pull bits from different places and they are not contiguous. There is also trying to be endian neutral.

sheinz:
Any ideas how to rename the driver to reflect that it is no longer just *bmp280* but *bme280* as well? May be `bmx280` ?

examples/bmp280/bmp280_example.c
ourairquality: Can rename this, was thinking the same. Perhaps some of the other drivers APIs could be harmonized too, such as the bmp180. Can we do that as a follow up as I want to get things working for a start.

Our data logger may well use the constants, log them at startup and then just log the ADC values to allow post calibration if necessary, but not certain yet. Could add an API to encapsulate access to them, but it just does not seem like the light weight style of this code base.

ourairquality:
I suggest renaming it to the bme280 driver, or just leaving it named the bmp280 driver. Our group has chosen the bme280 for our first trial run and it includes the humidity so I am guessing it will be much more popular than the bmp280. The http://purpleair.org/ group appear to have also chosen the bme280 for they stations, and I have seen other reports that it has relatively good accuracy.

Would you like me to renaming it for this PR?

extras/bmp280/bmp280.c
```diff
@@ -46,116 +46,155 @@
...
 {
     params->mode = BMP280_MODE_NORMAL;
     params->filter = BMP280_FILTER_OFF;
     params->oversampling = BMP280_STANDARD;
+    params->oversampling_humidity = BMP280_STANDARD;
     params->standby = BMP280_STANDBY_250;
 }
 
-static uint8_t read_register8(uint8_t addr)
+static bool read_register8(uint8_t i2c_addr, uint8_t addr, uint8_t *value)
```
sheinz: After the changes this function seems unnecessary... or could be marked **inline**.

ourairquality:
Renamed bmp280_calib_t to bmp280_t, and removed read_register8.

Do you have any preference for renaming this to the bme280 driver? Using bmx280 might be a problem if a different device with that name is ever released, and I think people will look for 'bme280' support and it might be just a little easier to find if it is all renamed?

sheinz:
> Do you have any preference for renaming this to the bme280 driver? Using bmx280 might be a problem if a different device with that name is ever released, and I think people will look for 'bme280' support and it might be just a little easier to find if it is all renamed?

I also though that a new device with name 'bmx...' might be released and we will have to rename it again. I think it's ok to keep it as it is.

There's a README.md with outdated information. Would it be possible to update it to reflect BME280 support?

ourairquality:
Documentation updated. ptal

sheinz:
I checked the changes with my bmp280 module and it works ok. It's good for merging in my opinion.

sheinz:
Thanks @ourairquality!

PR #187, opened 2016-08-15
iomux_set_function: remove the IOMUX_FUNC transform of the func argument

ourairquality:
The allows the IOMUX_GPIO<n>_FUNC_<function> definitions to be used here which seems to be the result of discussions in https://github.com/SuperHouse/esp-open-rtos/pull/171

### DIFF
--- a/core/esp_spi.c
+++ b/core/esp_spi.c
@@ -30,32 +30,27 @@
 
 static bool _minimal_pins[2] = {false, false};
 
-inline static void _set_pin_function(uint8_t pin, uint32_t function)
-{
-    iomux_set_function(gpio_to_iomux(pin), function);
-}
-
 bool spi_init(uint8_t bus, spi_mode_t mode, uint32_t freq_divider, bool msb, spi_endianness_t endianness, bool minimal_pins)
 {
     switch (bus)
     {
         case 0:
-            _set_pin_function(_SPI0_MISO_GPIO, _SPI0_FUNC);
-            _set_pin_function(_SPI0_MOSI_GPIO, _SPI0_FUNC);
-            _set_pin_function(_SPI0_SCK_GPIO, _SPI0_FUNC);
+            gpio_set_iomux_function(_SPI0_MISO_GPIO, _SPI0_FUNC);
+            gpio_set_iomux_function(_SPI0_MOSI_GPIO, _SPI0_FUNC);
+            gpio_set_iomux_function(_SPI0_SCK_GPIO, _SPI0_FUNC);
             if (!minimal_pins)
             {
-                _set_pin_function(_SPI0_HD_GPIO, _SPI0_FUNC);
-                _set_pin_function(_SPI0_WP_GPIO, _SPI0_FUNC);
-                _set_pin_function(_SPI0_CS0_GPIO, _SPI0_FUNC);
+                gpio_set_iomux_function(_SPI0_HD_GPIO, _SPI0_FUNC);
+                gpio_set_iomux_function(_SPI0_WP_GPIO, _SPI0_FUNC);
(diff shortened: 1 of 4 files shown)


core/include/esp/iomux.h
```diff
@@ -43,10 +43,17 @@ inline static esp_reg_t gpio_iomux_reg(const uint8_t gpio_number)
...
+ * gpio_iomux_reg above to obtain the IOMUX register number of a GPIO number.
+ * The func is an IOMUX_GPIO<n>_FUNC_<function> value, see iomux_regs.h
+ *
+ * ie *gpio_iomux_reg(3) is equivalent to IOMUX_GPIO3
+ */
 inline static void iomux_set_function(uint8_t iomux_num, uint32_t func)
 {
     uint32_t prev = IOMUX.PIN[iomux_num] & ~IOMUX_PIN_FUNC_MASK;
-    IOMUX.PIN[iomux_num] = IOMUX_FUNC(func) | prev;
+    IOMUX.PIN[iomux_num] = func | prev;
```
sheinz: If you remove `IOMUX_FUNC` here it should be added in function `iomux_set_gpio_function`, otherwise `iomux_set_gpio_function` will be broken.

core/esp_spi.c
```diff
@@ -23,8 +23,8 @@
 #define _SPI1_SCK_GPIO  14
 #define _SPI1_CS0_GPIO  15
 
-#define _SPI0_FUNC 1
-#define _SPI1_FUNC 2
+#define _SPI0_FUNC IOMUX_FUNC(1)
+#define _SPI1_FUNC IOMUX_FUNC(2)
```
sheinz: As I suggested in https://github.com/SuperHouse/esp-open-rtos/pull/171 the function `gpio_set_iomux_function` might be added in *gpio.h*. In that case `_set_pin_function` will not be needed here. Also I'm not sure if _SPIx_FUNC defines are necessary here as they already defined in iomux_regs.h

core/include/esp/iomux.h
```diff
@@ -43,10 +43,17 @@ inline static esp_reg_t gpio_iomux_reg(const uint8_t gpio_number)
     return &(IOMUX.PIN[gpio_to_iomux(gpio_number)]);
 }
 
+/**
+ * Set the I/O Mux function. The iomux_num is an IOMUX register number, see
+ * gpio_iomux_reg above to obtain the IOMUX register number of a GPIO number.
```
sheinz: Not `gpio_iomux_reg` but `gpio_to_iomux`

core/include/esp/iomux.h
```diff
@@ -43,10 +43,17 @@ inline static esp_reg_t gpio_iomux_reg(const uint8_t gpio_number)
...
+ * The func is an IOMUX_GPIO<n>_FUNC_<function> value, see iomux_regs.h
+ *
+ * ie *gpio_iomux_reg(3) is equivalent to IOMUX_GPIO3
```
sheinz: You can't pass IOMUX_GPIO3 into this function as *iomux_num* parameter. *iomux_num* is an index of a register. IOMUX_GPIO3 is a register itself.
It is still a confusing function with confusing parameters.

core/include/esp/iomux.h
ourairquality: Fixed, thanks.

core/include/esp/iomux.h
ourairquality: Sorry, cut-and-paste error. ptal

core/esp_spi.c
ourairquality: Done, thank. ptal.

Issue #176, opened 2016-08-03
Function sdk_system_deep_sleep ignores argument.


sheinz:
From the function signature it is clear that system should go to deep sleep for the specified amount of microseconds. But regardless of specified value the following output is shown:
```deep sleep 1073s```

I verified that in function **sdk_system_deep_sleep** the value is correct.
And correct value is passed into 
``` sdk_os_timer_setfn(&sdk_sta_con_timer, _deep_sleep_phase2, (void *)time_in_us);```
But the timer callback function is called with incorrect value.

I checked with disabled **open_esplibs** the result is the same.

**Comment by sheinz** [comment]
I've located the problem:
```
static void _deep_sleep_phase2(void *timer_arg) {
    uint32_t time_in_us = (uint32_t)timer_arg;

    printf("deep sleep %ds\n\n", time_in_us / 1000000);
```
The argument is not timer_arg but timer_handler. The actual argument should be extracted with function **pvTimerGetTimerID**:
```
    uint32_t time_in_us = (uint32_t)pvTimerGetTimerID(timer_handle);

    printf("deep sleep %ds\n\n", time_in_us / 1000000);
```

**Comment by kanflo** [comment]
Nice catch, thanks. Could you please provide a PR with the fix?

**Comment by sheinz** [comment]
I'm not sure if my fix is the best solution.
I looked into the new release of ESP8266_RTOS_SDK and they removed timers.o and use timers from ets_timer.o which does not use FreeRTOS timers.
If the timer from timers.o is used only once in function **sdk_system_deep_sleep** we might remove it as well and switch to the timer from ets_timer.o.
Less code is better.

**Comment by sheinz** [comment]
As it turned out timers *sdk_os_timer_* are used by binary libraries as well as by *lwip* code.

If not properly extract a timer argument from the timer handle in the timer callback function the argument is invalid. 
For example currently in [lwip](https://github.com/SuperHouse/esp-lwip/blob/esp-1.4.1-rtos/espressif/espconn_tcp.c#L227) :
```
static void ICACHE_FLASH_ATTR
espconn_cclose_cb(void *arg)
{
	espconn_msg *pclose_cb = arg;
```
I'm not sure if timer argument is properly extracted in binary libraries... If not, passing argument to a timer callback function doesn't work at all.

[...]

PR #190, opened 2016-08-17
Fix for system_deep_sleep ignoring argument

sheinz:
Fix for https://github.com/SuperHouse/esp-open-rtos/issues/176.

The problem was  that *sdk_os_timer_* timers use FreeRTOS timers underneath which do not pass a timer argument to the callback function. They pass FreeRTOS timer handle which should be used in function **pvTimerGetTimerID** to extract the actual argument.

*sdk_ets_timer_* timers do not use FreeRTOS timers and pass a timer argument directly into a callback function.

### DIFF
--- a/include/etstimer.h
+++ b/include/etstimer.h
@@ -36,4 +36,8 @@ typedef struct ETSTimer_st {
     void *timer_arg;
 } ETSTimer;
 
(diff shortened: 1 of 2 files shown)


projectgus:
This looks like a good fix to me, at least until some later time when more of this can be rewritten.

PR #191, opened 2016-08-18
iomux_set_function: fix merge, code comments.

ourairquality:
Fixes the merge from https://github.com/SuperHouse/esp-open-rtos/pull/187 Just a code comment correction.


sheinz:
I don't see much difference between those comment blocks. Also the names of the parameter `func` and `iomux_func` are pretty much the same. So. I would prefer to leave the function as it is now.

ourairquality:
Ok, thought it was a merger error, perhaps you improved it.

PR #192, opened 2016-08-18
sysparam: buffer writes.

ourairquality:
If the source is in the flash then the write to the flash seems to just write zeros, and since it is common for the key to be a constant string, buffer all writes.

Perhaps someone could check if this makes sense?

### DIFF
--- a/core/sysparam.c
+++ b/core/sysparam.c
@@ -116,16 +116,18 @@ static struct {
 
 /***************************** Internal routines *****************************/
 
-static inline IRAM sysparam_status_t _do_write(uint32_t addr, const void *data, size_t data_size) {
-    CHECK_FLASH_OP(sdk_spi_flash_write(addr, (void*) data, data_size));
+/* The buffer is needed in case the source is in the flash. */
+static inline IRAM sysparam_status_t _do_write(uint32_t addr, const void *data, void *buffer, size_t data_size) {
+    memcpy(buffer, data, data_size);
+    CHECK_FLASH_OP(sdk_spi_flash_write(addr, buffer, data_size));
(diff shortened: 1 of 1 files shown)


core/sysparam.c
```diff
@@ -116,16 +116,18 @@ static struct {
...
     return SYSPARAM_OK;
 }
 
 static inline IRAM sysparam_status_t _do_verify(uint32_t addr, const void *data, void *buffer, size_t len) {
     CHECK_FLASH_OP(sdk_spi_flash_read(addr, buffer, len));
-    if (memcmp(data, buffer, len)) {
+    if (memcmp(data, buffer, len))
         return SYSPARAM_ERR_IO;
-    }
+
```
foogod: Why are you removing these braces?  Not having braces on multi-line `if` constructs is (IMHO) very bad coding practice and should be discouraged.

foogod:
Aside from my comment about unnecessarily removing useful braces, this looks like a reasonable approach to solving the problem..

core/sysparam.c
ourairquality: Guess it all depends on your 'good coding style', sorry. Many coding styles omit braces if there is a single statement for the `if` and no `else`. Shall put it back!

kanflo: IMHO (regardless of coding style) I prefer to have light weight patches only containing changes relevant for the PR. General code cleanup, fixing indentation, renaming variables for clarity I like to see in separate PRs.

ourairquality:
Moved this along a lot further in https://github.com/SuperHouse/esp-open-rtos/pull/195

PR #193, opened 2016-08-18
sysparam: remove IRAM declarations.

ourairquality:
Not sure why some of the functions had IRAM declarations, it appears unnecessary. These were in the flash write path but well above the critical section, and writes are expected to be a cold path, so perhaps it would be better to save some RAM.

### DIFF
--- a/core/sysparam.c
+++ b/core/sysparam.c
@@ -116,12 +116,12 @@ static struct {
 
 /***************************** Internal routines *****************************/
 
-static inline IRAM sysparam_status_t _do_write(uint32_t addr, const void *data, size_t data_size) {
+static inline sysparam_status_t _do_write(uint32_t addr, const void *data, size_t data_size) {
     CHECK_FLASH_OP(sdk_spi_flash_write(addr, (void*) data, data_size));
     return SYSPARAM_OK;
 }
 
-static inline IRAM sysparam_status_t _do_verify(uint32_t addr, const void *data, void *buffer, size_t len) {
+static inline sysparam_status_t _do_verify(uint32_t addr, const void *data, void *buffer, size_t len) {
     CHECK_FLASH_OP(sdk_spi_flash_read(addr, buffer, len));
     if (memcmp(data, buffer, len)) {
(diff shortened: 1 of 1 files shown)


foogod:
These were set IRAM as a performance optimization.  Since (I believe) flash writes completely blow away the IROM cache, switching back and forth between IROM and flash IO routines repeatedly can potentially be quite a performance hit.  The `_write_and_verify` process can potentially involve quite a number of these calls depending on the size of the data being written.

These routines are not very large.  I would expect the savings to be on the order of a couple of tens of bytes..  At the very least, I would recommend we do some performance tests to see whether the speed impact is significant or not first.

ourairquality:
But these are slow paths, called a few times in the life of the device, so it's not worth bothering about. If there were any significant performance impact then these would be updating too quickly and the flash would wear out in a short period.

foogod:
That's not necessarily true, and it's making some big assumptions about usage patterns.

Keep in mind that the Winbond flash chips on these devices are rated to a minimum of 100,000 erase cycles (and I've heard reports that most actually last many times longer than the minimum).  Even using a modest 8-sector sysparam region (16KB of usable parameter space), with the default 64 byte read/write buffer we're using, that comes out to over 50 million calls to the flash read/write functions before we're in danger of wearing anything out (and more realistically may be closer to 500 million in practice).  If even a small portion of those 50 million happen to be part of a single write-and-verify, performance could be an issue.

It's quite possible that somebody would develop an application that might write a lot of data to a larger sysparam region intermittently (i.e. "burstily") and not wear out the flash for years, but still have its performance during peak (likely most critical) times substantially impeded if the sysparam write routines are unnecessarily slow.  Before we try to save a measly couple of bytes of IRAM, I'd like to at least know how much of an impact that might incur (and how much IRAM space it actually saves).

It's possible that it's all irrelevant and the speed difference is not that significant, but I haven't tested it so I don't think we can just assume that.

ourairquality:
Even if it added 1 second and was updated each hour it would still be insignificant. Anyway it's not a big issue, nothing broken.

PR #194, opened 2016-08-18
sysparam: add a semaphore to synchronize writers.

ourairquality:
This synchronizes writers, but not yet readers.

It appears the code supports multiple readers and a single writer. There seems to be an issue with multiple compaction still to be explored, but this PR is start to at least synchronize the writers.

### DIFF
--- a/core/sysparam.c
+++ b/core/sysparam.c
@@ -10,6 +10,8 @@
 #include <sysparam.h>
 #include <espressif/spi_flash.h>
 #include <common_macros.h>
+#include "FreeRTOS.h"
+#include "semphr.h"
 
 //TODO: make this properly threadsafe
 //TODO: reduce stack usage
@@ -112,6 +114,7 @@ static struct {
     uint32_t end_addr;
     size_t region_size;
     bool force_compact;
+    xSemaphoreHandle sem;
 } _sysparam_info;
 
 /***************************** Internal routines *****************************/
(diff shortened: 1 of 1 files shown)


foogod:
This seems reasonable to me.  Note that there shouldn't be any need to synchronize readers, since with the atomic update process they are entirely threadsafe, even when there is a writer making updates at the same time, so we should only need to protect against multiple simultaneous writes..

ourairquality:
Looking over the read paths there seem to be more thread safety issues, and I think readers need locking too. For example consider `_find_entry` accessing `_sysparam_info.cur_base` which a writer could change.

I don't buy the argument that other code has more significant issues so this code need not be safe, and I doubt other people do either.

Locking readers might mean that the `sysparam_iter_*` functions need to go, or be some separate reflective debug code and not for production use. Perhaps better to change the API to return a list of the key names, rather than an iterator, or to return a key matching a given index, anything atomic.

kanflo:
Thanks for this fix @ourairquality. I guess you considered one single semaphore for the entire sysparam? I guess it boils down to what usage one sees in sysparam. Personally, I do not regard it as a high traffic data store. Thus, having a global lock would be fine by me. Others may of course have a different view of sysparam and therefore want a more fine grain locking, for performance.

ourairquality:
@kanflo The changes in https://github.com/SuperHouse/esp-open-rtos/pull/195 include better thread safety support also for reads. Perhaps I should have closed this.

kanflo:
Bummer, I was coming to #195. Perhaps those changes could be broken out of that PR?

PR #195, opened 2016-08-19
sysparam: rework.

ourairquality:
Here's my wip changes to the sysparam code fwiw:

Adds a semaphore used by readers are writers.

Fixes writing to the flash from constant data stored in the flash, using a bounce buffer.

Removed memory allocation from most read and write paths. Only read paths that return a blob of data allocate memory now, and the iterator.

Store small integers as binary values, avoiding parsing and formatting in these paths.

### DIFF
--- a/core/include/sysparam.h
+++ b/core/include/sysparam.h
@@ -169,24 +169,20 @@ sysparam_status_t sysparam_get_info(uint32_t *base_addr, uint32_t *num_sectors);
  */
 sysparam_status_t sysparam_get_data(const char *key, uint8_t **destptr, size_t *actual_length, bool *is_binary);
 
-/** Get the value associate with a key (static buffers only)
+/** Get the value associated with a key (static value buffer)
  *
  *  This performs the same function as sysparam_get_data() but without
- *  performing any memory allocations.  It can thus be used before the heap has
- *  been configured or in other cases where using the heap would be a problem
- *  (i.e. in an OOM handler, etc).  It requires that the caller pass in a
- *  suitably sized buffer for the value to be read (if the supplied buffer is
- *  not large enough, the returned value will be truncated and the full
- *  required length will be returned in `actual_length`).
- *
- *  NOTE: In addition to being large enough for the value, the supplied buffer
- *  must also be at least as large as the length of the key being requested.
- *  If it is not, an error will be returned.
+ *  allocating memory for the result value.  It can thus be used before the heap
+ *  has been configured or in other cases where using the heap would be a
+ *  problem (i.e. in an OOM handler, etc).  It requires that the caller pass in
+ *  a suitably sized buffer for the value to be read (if the supplied buffer is
+ *  not large enough, the returned value will be truncated and the full required
+ *  length will be returned in `actual_length`).
  *
  *  @param[in]  key            Key name (zero-terminated string)
- *  @param[in]  buffer         Pointer to a buffer to hold the returned value
- *  @param[in]  buffer_size    Length of the supplied buffer in bytes
- *  @param[out] actual_length  pointer to a location to hold the actual length
+ *  @param[in]  dest           Pointer to a buffer to hold the returned value.
+ *  @param[in]  dest_size      Length of the supplied buffer in bytes.
+ *  @param[out] actual_length  Pointer to a location to hold the actual length
  *                             of the data which was associated with the key
  *                             (may be NULL).
  *  @param[out] is_binary      Pointer to a bool to hold whether the returned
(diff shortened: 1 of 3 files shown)


core/include/sysparam.h
```diff
@@ -251,17 +247,13 @@ sysparam_status_t sysparam_get_string(const char *key, char **destptr);
  *  @retval ::SYSPARAM_ERR_CORRUPT  Sysparam region has bad/corrupted data
  *  @retval ::SYSPARAM_ERR_IO       I/O error reading/writing flash
  */
-sysparam_status_t sysparam_get_int(const char *key, int32_t *result);
+sysparam_status_t sysparam_get_int32(const char *key, int32_t *result);
```
kanflo: I think this clarification is a good one.

core/sysparam.c
```diff
@@ -719,97 +761,65 @@ sysparam_status_t sysparam_get_string(const char *key, char **destptr) {
...
-
     *result = value;
-    free(buffer);
-    return SYSPARAM_OK;
+    return status;
 }
 
-sysparam_status_t sysparam_get_bool(const char *key, bool *result) {
-    char *buffer;
+sysparam_status_t sysparam_get_int8(const char *key, int8_t *result) {
```
kanflo: I can see the use for a get_int8 but I would like to have the get_bool unchanged. As always, contributions improving functionality are very welcome without discussion but changes like this I think need to be discussed first.

ourairquality: Do we really need a get_bool that parses a string parameter?

With a string we need to malloc the result, but with the proposed get_int8 there is no need to malloc so it might be safer in more places. Patched the reading of the payload so it can be unaligned for this too. With get_uint8 a bool version seems redundant as a byte is the smallest unit of storage.

This is all just wip as I integrate use of sysparam in some code, and it has been really helpful, a big thank you to @foogod

ourairquality:
There are dependent overlapping changes here, that is why I gave up on smaller patches. I doubt I have time to split it all up and incrementally get this from A to B, it's just my working version fwiw.

This set does design for small size and safety, probably at the expense of performance, for example it uses small stack allocated bounce buffers which allows it to avoid dynamic memory allocation in most paths, and this in turn allowed other simplifications.

core/sysparam.c
foogod: Part of the purpose of sysparam was also to create an easy way for the core and applications to read and write human-readable configuration settings for a lot of parameters that humans often want to change.  That is why both the bool routines and the int stuff was designed to work with human-readable strings.  (If you're concerned about size, note that bool values can be stored as single-byte"y"/"n"/etc, and in the case of ints, the majority of int values people are likely to be working with are actually smaller when stored as text than as a 32-bit binary value)

I can see some value in reducing the need for mallocs, but I think this is an unnecessary removal of useful behavior.  If you really want to do this I would recommend perhaps adding a way to read/write small strings internally using the bounce buffer so that bool and int text values could be read/written entirely using that buffer..

core/sysparam.c
```diff
@@ -954,39 +947,22 @@ sysparam_status_t sysparam_set_data(const char *key, const uint8_t *value, size_
...
-    free(buffer);
-
  done:
     xSemaphoreGive(_sysparam_info.sem);
 
     return status;
 }
 
-sysparam_status_t sysparam_set_string(const char *key, const char *value) {
-    return sysparam_set_data(key, (const uint8_t *)value, strlen(value), false);
```
foogod: Is there some reason you removed the const from the input value here?

foogod:
I haven't had a chance to look over all of this in thorough detail, but a couple of things I noticed:

1. I see you've re-removed the IRAM declarations we had already been discussing in the other PR which you withdrew.  I'm assuming this was inadvertent rather than a deliberate attempt to sneak something in that I'd already mentioned I had some issues with.

2. The locking around the basic read routines is, I'm fairly certain, completely unnecessary (at least as I originally wrote them, not sure if your changes change any of that), but I suppose ultimately it doesn't really hurt anything too much..  Earlier you had mentioned the iter routines, which I haven't had a chance to go back and look at, but I acknowledge may have some thread issues, but it looks like you haven't attempted to tackle that here?

I understand the issues with multiple interdependent changes, but the size of this PR makes it really hard to evaluate properly.  If you could find some way to break it up even a bit it might help make the process move quicker, otherwise it may take a while for folks to have enough time to properly review this mass of changes..

ourairquality:
Sorry, I don't like the tone of this discussion. We don't have to agree, I just need something that works and fits my needs so I'll keep it elsewhere. Again, thank you for the sysparam contribution, it's a great step to be able to get all the parameters in this database.

core/sysparam.c
kanflo: I would like to add that the inclusion of sysparam_get_bool sort of was agreed upon when the original PR got merged. This does not mean it is written in stone and must remain that way forever of course. As I commented on one of your PRs, it is a good practice to keep PRs short and, how should I put it, singleton. You have some very good changes in this PR that improves the sysparam framework but those got blocked from being merged due to other parts needing discussion. Had the PR been broken up, we could have merged the good parts without further discussion.

kanflo:
I am sorry for feel that way @ourairquality. Very often we just need something that works as you say. Going the extra mile and turning "what I need" into "something that others may benefit from" is IMHO sometimes painful but in the end quite rewarding. One of the hurdles you need to pass is the scrutinization of your work and that can also be a painful experience. For my part I have always found that the result actually was better code than what I originally submitted.

ourairquality:
@kanflo It's not personal, I just don't have much time, perhaps some other small isolated patches. I took another look at it to see if something could be split out and it's another block of work, the read locking changes requires rewriting the functions, just trivial stuff to make sure the lock is released on all paths, but it's on top of other changes. Also you need to decide if you would want to follow the same direction and use the small stack allocated buffers, take the simplifications and safety, perhaps lose some performance - subjective matters I would rather sit out.

PR #196, opened 2016-08-20
esp8266 headers: A new interrupt no and a new WDT register

pfalcon:


sheinz:
Thanks @pfalcon for the PR.

I checked **WDT.VAL** value on my device. And I discovered the following:
 * The value is decremented from *0x7FFFFFB*. I think *0x7FFFFFF* is the initial value after reset.
 * The decrement frequency is 80MHz and independent from the CPU clock. 
 * Writes to this registers seems to be ignored.

Do you have more information on WDEV_FIQ interrupt?

pfalcon:
> The decrement frequency is 80MHz and independent from the CPU clock.

How did you figure that out?

> The value is decremented from 0x7FFFFFB. I think 0x7FFFFFF is the initial value after reset.

Well, it definitely counts from *F. One way you could measure frequency above is by reading consecutively and comparing wdt counter decrement with instruction cycles. But then you would allow for the fact that between feeding and next read it already counts down a bit. You can see *F value there by disable WDT countdown (0x38 to *_CTRL).

 > Do you have more information on WDEV_FIQ interrupt?

Nope. What more information you'd like to have? It's an IRQ from WiFi controller, definitely.

The idea is to enumerate all IRQs in ESP8266, According to bootrom, there're 14 interrupts, the last one however is level 3, not level 1.

pfalcon:
Well, that level 3 Interrupt is NMI (based on its handler in bootrom being "rfi      0x3").

pfalcon:
And level 2 is debug interrupt, all fits nicely with Xtensa arch refman.

sheinz:
> How did you figure that out?

```
WDT.FEED = WDT_FEED_MAGIC;
prev = WDT.VAL;
sdk_os_delay_us(100);
uint32_t diff = prev - WDT.VAL;
printf("wdt diff: %d\n",);
```
Which outputs WDT decrements for roughly 100 us. I tried different delays and the result is consistent with 80 WDT ticks per 1 us. 
I also tried changing CPU frequency but WDT decrements didn't change.

Issue #66, opened 2015-11-13
Is libhal actually needed?


DagAgren:
In my test build, only two libhal functions are actually used, xthal_get_ccount and xthal_set_intset. xthal_get_ccount is provided in ROM, so it can easily be eliminated. That just leaves one instance of xthal_set_intset in FreeRTOS. If that was replaced, then libhal could be dropped as a dependency.

foogod:
This was another thing I was meaning to look into at some point, but to be honest was never really that high on the priority list..

I don't see any reason not to use the xthal_get_ccount from the ROM instead.. as for xthal_set_intset, it would be good if we didn't have to make custom changes to the FreeRTOS stuff just for this..  We could potentially just put our own implementation (possibly a weak symbol) in our libs, though, so it wouldn't have to pull in libhal for just that one function..

DagAgren:
When I looked at the Espressif FreeRTOS code, it looked somewhat ghastly, so I'm not sure trying to keep their code untouched is the best option. Of course, branching off it would add another maintenance burden...

resetnow:
I assume this can get closed after #198 was merged.

projectgus:
Yes, fantastic - thanks!

PR #198, opened 2016-08-21
FreeRTOS: Drop reference to xthal_set_intset

resetnow:


kanflo:
Hi @resetnow. Could you please elaborate on the need for removing xthal_set_intset?

resetnow:
Hi @kanflo — I saw it being discussed <a href="https://github.com/SuperHouse/esp-open-rtos/issues/66">here</a>, and since xthal_set_intset is just one instruction, I thought that eliminating a library dependency could be good for a code size.

kanflo:
Thanks for the pointer @resetnow. Do you have any input @foogod as you where part of the original discussion?

projectgus:
This looks good, it'd be nice not to have to link libhal (some toolchains don't ship it.)

We have a convenience macro for using WSR from C here: core/include/xtensa_ops.h - is it possible to use this? (I noticed you used register contraint 'a' not 'r' in the PR, but I'm not sure if this makes any difference to the output on xtensa.)

jcmvbkbc:
> I noticed you used register contraint 'a' not 'r' in the PR, but I'm not sure if this makes any difference to the output on xtensa

There's no difference in call0 ABI. In windowed ABI 'a' constraint is 'all registers except the stack pointer'.

resetnow:
I have updated the code to use the macro from xtensa_ops.h

PR #199, opened 2016-08-24
sysparam editor: add an echo command.

ourairquality:
Helpful to be able to disable echo when sending commands fast, so that the editor can consume them faster than they arrive. This adds and 'echo' command to toggle echo on and off.

### DIFF
--- a/examples/sysparam_editor/sysparam_editor.c
+++ b/examples/sysparam_editor/sysparam_editor.c
@@ -6,6 +6,8 @@
 #include <sysparam.h>
 
 #include <espressif/spi_flash.h>
+#include "espressif/esp_common.h"
+#include "esp/uart.h"
 
 #define CMD_BUF_SIZE 5000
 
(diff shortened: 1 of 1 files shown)


yesco:
Hi

I would suggest that it doesn't toggle as the state may not be known.
Better to provide on off methods for predictable behavior.

One issue with the Hayes commands for esp8266 is that it's difficult to
assert a state or make sure something starts in the state you want to have,
when using it programmatically.

Anyway just an idea,

cheers

ourairquality:
@yesco Sounds like a good idea thanks. Do you have a suggestion: `echo-on` `echo-off` or `echo 0` `echo 1` etc?

ourairquality:
@yesco Changed to `echo-on` and `echo-off` fwiw.

sheinz:
This looks good to me. I've tried the example. Works great.

kanflo:
Thanks @ourairquality. Just out of curiosity, are you testing sysparam for speed since you need to turn off echoing?

ourairquality:
@kanflo Not looking at performance. My application loads parameters into RAM at startup, and writes very infrequently, performance is not an issue for me, just needed it lightweight and safe. The echo was a problem using the editor to quickly get some automation running because the delay caused the uart input buffer to overflow.

yesco:
Good. Cheers

PR #202, opened 2016-08-26
faster i2c library

Zaltora:
I create a new i2c library based on other library that exist for esp8266. API is similar with the current i2C library to easily change. I made this library for faster communication (example: ssd1306, sensors )
It is not as complete as the original and there are improvements to do. It is an alternative to the current i2c, not a replacement.
You can use "for loop" (from ws2812 library) and select , 400kHz, 700kHZ ,... ,1.5MHz ( CPU clock dependency )
You can use sdk_os_delay_us() and go up to 300kHz.

I will do measurement for each significant value and upload it as enum for CPU 80Mhz and 160Mhz clock


sheinz:
Hi @Zaltora,

Thanks for contributing!
I didn't quite get why is our current i2c library slow? What is the bottleneck in it?
If there is a key that allows you to write a fast i2c library can we apply it in our current library?
I'm not sure if we need two libraries that do the same but one is fast and another is slow.

Another suggestion is to have at least one example with each new library that uses it. So, the library code is actually compiled during Travis CI check.

kanflo:
Thanks for the i2c improvement @Zaltora . Looking at the old driver (that I actually submitted but I do not feel religious about these things) it seems this one inlines i2c_[read|write]_bit and optionally uses a volatile for loop for short delays. I agree with @sheinz we should not have two separate i2c drivers but the nice improvements from this one could easily go into the current one. One thing I like about the original "Wikipedia i2c driver" is the multitude of comments explaining how i2c actually works.

Zaltora:
i Think it is Clock stretching and multi master support than make original lib slow.  wikipedia describe all case but they are very uncommon. Integrate them slow everything.
Into fast lib clock stretching still exist but only at the end of the transmission because the slave device generally stretch the line to be able to compute without interrupt after receive data. Orignial lib test it all time when transmitting.
Multi master is maybe used on I2C but it is not the principal purpose of this link.

I doing some test on lib with different delay:
`sdk_os_delay_us(1);`
original lib : ~90kHz
fast lib : ~370kHz
`for (volatile uint32_t __j = 4; __j > 0; __j--) ;` at 160 MHz
original lib : ~110kHz
fast lib : ~900kHz

I do one fix and create an example (a i2c scan) . It is my first pull request, i am not sure what to do. What you suggest ?
replace original lib and just keep the principal I2C transmission purpose (one master , multiple slaves) ?
trying to improve the original (i dont think to had same performance) ?

ourairquality:
It's not clear that clock stretching is causing the performance issue, the loops exit as soon as the SCL is read as high. Could the overhead of just reading the SCL line cause such a performance issue? If so then some inlining might help. I think the current driver should be improved to the limited, using inlining etc, and the performance bottlenecks understood before compromising the generality.

ourairquality:
@Zaltora Could you give the change in PR https://github.com/SuperHouse/esp-open-rtos/pull/214 a try. It seems to make a big improvement here. You might also want to change the delay. Testing on only a single i2c device so far.

ourairquality:
@Zaltora Might be best to create an new branch and an new PR for changes to the bmp280 driver.

Zaltora:
OOO, sry, i want create a new pull request for bmp280. I am pretty new with git and github. I don't remenber validate this !! i will fix it

ourairquality:
@Zaltora So `git checkout master`, `git checkout -b <new branch name>`, edit, `git commit`, `git push origin <new branch name>`

projectgus:
All commits in this branch seem to have been merged or reverted, should we close this PR @Zaltora ?

Zaltora:
Yes, i will probably propose change on current i2c library now :) (if needed)

PR #203, opened 2016-08-26
Default flash parameters for 32M bit.

ourairquality:
Suits the Nodemcu and Witty, and many other common boards.

All the boards I am using are 32M bit and DIO. Might there be support to change these defaults?

### DIFF
--- a/parameters.mk
+++ b/parameters.mk
@@ -10,7 +10,7 @@
 
 # Flash size in megabits
 # Valid values are same as for esptool.py - 2,4,8,16,32
-FLASH_SIZE ?= 16
+FLASH_SIZE ?= 32
 
 # Flash mode, valid values are same as for esptool.py - qio,qout,dio.dout
 FLASH_MODE ?= qio


projectgus:
A lot of boards are 32Mbit, but only some are DIO (and it's a performance regression for anyone who supports QIO, only half the SPI flash read performance.)

Did you know you can create a local.mk file in your esp-open-rtos directory and set any defaults you want to override in there?

ourairquality:
@projectgus Ok, thanks, so now just proposing to change the size to 32M, keeping the mode as QIO. All the nodemcu style boards I have work with QIO. The intention was to just reduce the number of configuration issues for people getting started in the common cases.

kanflo:
I second changing to 32M as default.

foogod:
The problem with this is that if the user is using a device that is *not* 32Mbit (only 16), and builds something that needs more space than what their hardware can support, with this change esp-open-rtos will happily build it, saying there's no problems and (last I checked) esptool.py will happily flash it (saying there's no problems) and then it will just crash in usually bizarre (or sometimes subtle) and hard-to-diagnose ways when they actually try to run it.  This is not good user experience (particularly for people just getting started) IMHO..

The advantage of setting the default low is that if the user tries to build too large an image the build process will complain and then they have to think about "can my hardware support this code?" (and if so override the setting), instead of just breaking horribly for no apparent reason.

ourairquality:
@foogod Ok, thanks, shall just add a local.mk file for examples that depend on 32Mbit.

ourairquality:
Problem with using local.mk is that it is currently in the .gitignore so if examples requiring 32M are to be include this then can it be removed from .gitignore ?

ourairquality:
Sorry for the noise, it's possible to override the .gitignore with a local .gitignore including !locak.mk

kanflo:
You have a point @foogod. I would rather have esptool.py check if the image can be flashed by checking flash size but that falls outside the scope of our project. Perhaps the 32M configuration should be mentioned somewhere on the wiki?

PR #205, opened 2016-08-26
core: exception_vectors: initial esp-gdbstub support

resetnow:
This commit adds initial support for <a href="/resetnow/esp-gdbstub">esp-gdbstub</a> (which is based on Espressif implementation and adapted to esp-open-rtos). 

A thing I'd like to hear an opinion on: two global symbols were added: `debug_exception_handler` and `debug_saved_ctx`. I'll take any suggestion regarding renaming them to suit some existing namespace.

Implementation notes:

1. Default debug exception vector now calls the address stored in `debug_exception_handler`. It's `fatal_exception_handler` by default, and gdbstub changes it during initialization.
2. `debug_saved_ctx` is used to save a0 — a4 at the start of debug vector. Also, it stores a0 — a1 in the interrupt vector (the rest of registers is saved later on the stack). The latter allows to use Ctrl-C in GDB.

### DIFF
--- a/core/exception_vectors.S
+++ b/core/exception_vectors.S
@@ -45,6 +45,15 @@ LoadStoreErrorHandlerStack:
         .word   0       # a3
         .word   0       # a4
 
+        .balign 4
+        .global debug_saved_ctx
+debug_saved_ctx:
+        .word   0       # a0
+        .word   0       # a1
+        .word   0       # a2
+        .word   0       # a3
+        .word   0       # a4
+
 /***************************** Exception Vectors *****************************/
 
         .section .vecbase.text, "x"
@@ -66,12 +75,7 @@ VecBase:
         .org    VecBase + 0x10
 DebugExceptionVector:
         .type   DebugExceptionVector, @function
-
-        wsr     a0, excsave2
-        mov     a2, a1
-        movi    a3, 0
-        call0   fatal_exception_handler
-        rfi     2
+        j       DebugExceptionHandler
 
         .org    VecBase + 0x20
 NMIExceptionVector:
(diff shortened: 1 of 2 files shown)


resetnow:
@projectgus could you please confirm whether or not this change can be merged in esp-open-rtos and what possibly should be changed?

core/exception_vectors.S
```diff
@@ -23,6 +23,13 @@
 #define CAUSE_LOADSTORE 3
 #define CAUSE_LVL1INT 4
 
+        .section .data
+        .balign 4
+        .global debug_exception_handler
+
+debug_exception_handler:
```
projectgus: Can you use a weak-linked symbol for the debug_exception_handler, rather than loading it into RAM?

core/debug_dumps.c might be a good place to define a weak symbol (by default it can be a weak alias to fatal_exception_handler, but if the user links a debug_exception_handler then it can replace this.)

core/exception_vectors.S
```diff
@@ -491,7 +530,15 @@ NMIExceptionHandler:
         .balign  4
 UserExceptionHandler:
         .type   UserExceptionHandler, @function
-        xsr     a0, excsave1    # a0 now contains sp
+        // save a0, a1 to debug_saved_ctx before stack pointer is affected
```
projectgus: I'm probably missing something obvious, but why do we need to save a0 & a1 to debug_saved_ctx here? Is it because the 'break' instruction under LUserFailOtherExceptionCause triggers a DebugException? debug_saved_ctx gets written there as well, so I don't fully understand where these values are later used.

projectgus:
Sorry for the slow reply!

In principle I think this is good groundwork to merge (whether or not we merge the full esp-gdbstub into esp-open-rtos or not.) I left a few minor comments on the diff.

core/exception_vectors.S
resetnow: This allows gdbstub to examine context state before C code for interrupt handling is run, so you can press Ctrl-C in GDB and see where the program is going (like prvIdleTask)

resetnow:
Thanks for the review! I'll update the PR.

As for merging esp-gdbstub itself, unfortunately, it's licensed under restrictive "Espressif MIT" license. There was a point when the code in their repo was without a license, but I don't know what are legal implications of this.

core/exception_vectors.S
projectgus: if the "break 1" instruction moves from  LUserFailOtherExceptionCause to UserExceptionVector (which already examines exccause for LOADSTORE, and probably has enough space to check for CAUSE_LVL1INT and break if it's not hat), is it then possible to trigger the debugger before we save interrupt state, and eliminate the need to save these registers twice?

resetnow: Debugger is triggered on UART interrupt, so when user calls breaks the execution, the interrupt is processed through LUserHandleInterrupt (which then calls C functions — thus the need to save a0 and a1). The rest of registers seems to be saved by xthal. Please correct me if I'm wrong, but it looks like when LUserHandleInterrupt is called, original values of a0 and a1 are lost, so these two must be saved right on the user exception entry,

projectgus: Ah, the UART interrupt part is the part that I missed. Thanks for explaining.

resetnow:
I have updated the debug exception handler symbol to be weak and have an alias.

kanflo:
Thanks @resetnow, I'll look forward to testing this.

PR #207, opened 2016-08-27
[WIP] Automatic tests.

sheinz:
I think it was discussed several times that it would be nice to have some automatic test system that would check for regression. It's common for any software project to have such a system. Unfortunately when it comes to an embedded system it's no that trivial as it requires some hardware manipulations.

Here's my attempt to bring some basic automatic testing to EOR. It's currently in very draft state.

The idea is to have one or two modules connected to the host PC. Each test is built and flashed into the module. Than a python script checks the serial output of the test. The script will report with its return code whether the test succeeded or failed.

There's a support for several modules that are tested simultaneously. For example to test wifi communication two modules connect to each other and send/receive test data (wifi basic test).

The modules should have automatic reset/boot-mode support like nodemcu modules. In order to be flashed without pressing any buttons on the device. I currently don't have such modules :( 

Such testing system allows easier rewriting of some binary libraries as a test can be written first and than rewritten implementation can be verified. Or just for general regression testing during optimization or other improvement.

**Usage**
To build all the tests run `make` in the `test` directory:
```
cd tests
make
```
To build and run all the tests run `make test` in the `test` directory:
```
cd tests
make test
```
To build and run a single test run `make flash run_test` in the specific test directory:
```
cd tests/spiffs
make flash run_test
```

For example:
```
> make flash run_test
   ...
Read speed: 732562 bytes/s
Write speed: 50123 bytes/s
Test finished successfully
Test: SUCCESS
```


kanflo:
Very nice @sheinz, I will go through this in detail as soon as possible.

projectgus:
@sheinz, this looks pretty good to me!

This probably predates your involvement with the project, but I also started some automated test support:
https://github.com/SuperHouse/esp-open-rtos/tree/feature/tests

I never finished it though (there are working tests that run on a single ESP, but multi-ESP tests I never finished implementing).

You can probably consider that branch "abandoned" now, I'm not going to pick it back up. But there might be some bits of code or tests there that you would find useful to roll into your efforts. Not a problem if you don't want to dig through it, though. :)

Nice work! :)

sheinz:
Closed in favour of https://github.com/SuperHouse/esp-open-rtos/pull/253

Issue #140, opened 2016-05-09
Clients disassociated every 10 seconds in SoftAP mode


projectgus:
Copied in from mailing list post: https://groups.google.com/d/msg/esp-open-rtos/4BlYYrYRfH8/D72Jo-yyBgAJ

> 10 seconds after a client connects to the ESP in SoftAP mode, the AP disassociates it for inactivity (regardless of whether it was active or not). Some of my devices were transparently re-associating without me noticing, but not all of mine were (and some, like my Android phone, actually seem to get stuck where the device thinks it's still associated even though it isn't.)
> 
> The short/quick answer is that you can work around this by replacing the SDK's internal hostap timeout function. It's hacky (and it means clients will never get booted for inactivity), but you can do it by adding this code to your program:
> 
> void sdk_hostap_handle_timer(void *cnx_node)
> {
> }
> 
> ... full explanation of what I think the behaviour around the bug is:
> 
> * In libnet80211/ieee80211_hostap.o when hostap_input() sees packets from a new station, it calls a static function which adds the station to the internal connection table - via cnx_node_alloc() - and sets up a timer to call hostap_handle_timer() every 10 seconds (repeating). The argument passed to hostap_handle_timer() is the cnx_node structure that was just allocated.
> 
> * hostap_input is also supposed to look up each incoming packet's sender(?) in the internal connection table (cnx_node_search()), and if it finds a matching entry it writes the current count of hardware timer FRC2 to offset 0x108 in the allocated 'cnx_node' structure. This is a "last seen" timestamp.
> 
> * Every ten seconds the hostap_handle_timer() function compares the current FRC2 timer count to the stored timestamp, and if they're too far apart (ie no wifi frame received recently) then it sends a disassociation frame to the idle client and removes it using cnx_node_leave().
> 
> * cnx_node_leave() itself checks some internal state and prints the "max connection!" message each time it removes an "idle" client. (Unclear if this is a bug itself, or a symptom of the bug.)
> 
> 
> ... the bug happens because the "last frame received" timestamp is never written to offset 0x108 in hostap_input, this value is always zero so even an active client is disconnected after 10 seconds.
> 
> I don't know what causes that buggy behaviour. The bug could be in hostap_input, or maybe the "cnx node" is never properly registered.

ourairquality:
[...]

Issue #115, opened 2016-03-13
Fatal exception in prvTimerTask  (access_point example)


elephh:
I start exproring esp-open-rtos sdk, and when I run acces_point example unmodified code I often get that error:

Fatal exception (9):.
epc1=0x40226a4c
epc2=0x00000000
epc3=0x40101367
excvaddr=0x00000016
depc=0x00000000
excsave1=0x40226ad4

addrsource 0x40226a4c says:

0x40226a4c is in prvTimerTask (/home/Espressif/esp-open-rtos/FreeRTOS/Source/timers.c:527).
522
523             while( xQueueReceive( xTimerQueue, &xMessage, tmrNO_DELAY ) != pdFAIL ) /*lint !e603 xMessage does not have to be initialised as it is passed out, not in, and it i
524             {
525                     pxTimer = xMessage.pxTimer;
526
527                     if( listIS_CONTAINED_WITHIN( NULL, &( pxTimer->xTimerListItem ) ) == pdFALSE )
528                     {
529                             /* The timer is in a list, remove it. */
530                             ( void ) uxListRemove( &( pxTimer->xTimerListItem ) );
531                     }
Line 527 of "/home/Espressif/esp-open-rtos/FreeRTOS/Source/timers.c" starts at address 0x40226a4c <prvTimerTask+188> and ends at 0x40226a50 <prvTimerTask+192>.
Dump of assembler code for function prvTimerTask:
   0x40226990 <+0>:     addi    a1, a1, -64
   0x40226993 <+3>:     s32i.n  a12, a1, 56
   0x40226995 <+5>:     s32i.n  a13, a1, 52
   0x40226997 <+7>:     l32r    a12, 0x4022698c
   0x4022699a <+10>:    l32r    a13, 0x40226750
   0x4022699d <+13>:    s32i.n  a0, a1, 60
   0x4022699f <+15>:    s32i.n  a14, a1, 48
   0x402269a1 <+17>:    s32i.n  a15, a1, 44
   0x402269a3 <+19>:    l32r    a3, 0x4022675c
   0x402269a6 <+22>:    movi.n  a15, 0
   0x402269a8 <+24>:    l32i.n  a2, a3, 0
   0x402269aa <+26>:    mov.n   a4, a15
   0x402269ac <+28>:    l32i.n  a3, a2, 0
   0x402269ae <+30>:    movi.n  a5, 1
   0x402269b0 <+32>:    moveqz  a4, a5, a3
[...]

PR #208, opened 2016-08-29
Wifi web based configuration.

ourairquality:
Practically our project needs some web based configuration support as most users would be challenged to run the sysparam editor. This is a quick start. Sharing it here in case other people need something similar and might be able to help, but happy to maintain elsewhere. It implements a minimal http server, attempts to be robust to user input by using a fixed sized stack allocated working buffer, truncating fields if unexpectedly large etc. One benefit of maintaining this here is that the support for this might be refined as the project develops, for example the control of the network management seems rather a black box at present and this might open up in future.


kanflo:
A nice add-on @ourairquality and a must have for creating anything else than one-off ESP nodes. I was just embarking on creating the same so I'll be happy to contribute. @projectgus, would you mind having this "sub project" evolve inside the EOR repository?

ourairquality:
Add more parameters for the AP and STA modes. Not having much luck with an attempt to allow STA static IP addresses. Also added a 'restart' button.

foogod:
This looks potentially really cool..  I'm actually wondering if we might be able to pull out some portions of this into a more generic "simple HTTP server" for other apps to use as well?  (doesn't necessarily have to be now, just an interesting thought for the future)

The plan was originally once sysparam got merged in to define a bunch of "system" parameters for these sorts of wifi/etc settings and have the core routines read them and configure everything on startup (instead of having the app have to do it itself)..  The convention we'd discussed was for core system parameters to be in all-upper-case in sysparam (and mixed/lower-case parameters are for apps to play with as they see fit)..  I'm thinking if we want to just define a bunch of these wifi parameters as "system" (uppercase) parameters here and now, then we could work on integrating support for them more directly into the core going forward as well..

(As a side-note, looks like this is built on #209 so there's a merge-dependency there..  Don't think that will be a big sticking point, though..)

ourairquality:
Thank you for the feedback. Yes I would like to make this more generic, perhaps in the `extras`, that project can include and then customize. Shall keep working on that.

I am also considering using it as a basic web server beyond configuration. For example to present sensor readings, and the AP mode will be great when mobile and away from a base wifi connection. Also some people might not have a wifi AP but still be able to use a phone or tablet to download and present data. The code will need to be better abstracted to support this.

If it is to operate as a web server then I think the configuration pages need to be protected, a password. One thought is an option to disable the configuration once working, and one 'restore' page with a password, or just use a hardware button to re-enable wifi configuration but that would not work so well if the device is installed out of easy reach.

Also people want to be able to erase the flash of their passwords. e.g. Loan out a device, people enter AP passwords, even if they change it then it is still likely in the flash. Perhaps exposing an api to compress the sysparam area would help.

I'm not sure about all the plans to use these parameters internally, it seems a black box at present, but would like to see this more transparent. Subjective, but upper case pokes me in the eye :) but internal naming seems a small matter I can follow.

ourairquality:
Reworked to move most of the code into `extras/wificfg/` and abstracted the path matching and dispatch to allow extra web pages to be added by the application with a small example. Still many issues with the wifi management, more todo, and the restart is also not reliable and can lock up.

kanflo:
Thumbs up for moving it to ```extras/```. I would like to have this merged as soon as your change to int32/int8 in sysparam is merged if no one disagree. Have you thought of the idea of moving the list of configurable parameters to the main application? I think if could be useful as some nodes might require MQTT configuration, some not and so on.

Regarding restart being unreliable I have seen that too (would not want to add an ATTiny as external watchdog but if I have to...).

robertfoss:
I'm not sure how license compatible this project is with the NodeMCU project, but I wrote the first versions of their configuration portal and would happily donate that.

ourairquality:
Made another pass over this code. Added flags to enable/disable the Station and AP mode separately, and a flag to enable or disable the configuration pages (with a password to re-enable). Moved the list of URLs and handlers to `extras/` to hide it from the app, adding a separate optional dispatch list for apps.

@kanflo Not clear on your suggestion to 'moving the list of configurable parameters to the main application'. Do you mean to use these in the core startup code etc? It seems a bit of a black box to me at this point, and I am just trying to have some code to init the network in the extras. Could use some help here. Can it be layered on the existing code for a start.

@robertfoss If you could offer a link then I would be interested to take a look?

A quick search for other esp wifi config tools found this: https://github.com/jeelabs/esp-link

The HTML in the current code need lol.

robertfoss:
@ourairquality This is the last commit which contained only contributions from me, and that you are welcome to use however you like.

There's a simple dns hijacker in that file too, that might be useful, if the current code doesn't already contain one.

https://github.com/nodemcu/nodemcu-firmware/blob/d19c350ead89c629d8ee9e85bac8502a40033ce7/app/modules/enduser_setup.c

ourairquality:
The dhcp server seems to tickle or cause some memory spikes, seemed to be crashing with OOM when used along with other memory demanding code. Works better with this disabled, and rules out this dhcp server code causing other problem noted. Just need to support a static IP address when setting up the wifi connection.

SoftAP mode is frustratingly unreliable. Hitting this crash https://github.com/SuperHouse/esp-open-rtos/issues/115 and Android devices are losing connections probably https://github.com/SuperHouse/esp-open-rtos/issues/140

Make another round or updates, got it working with Chrome, added AP IP address and netmask to the sysparams and the config interface, trimmed the task stack size.

kanflo:
@ourairquality, my idea was that some parameters are "core" and some are application specific. It would be nice to have both parameter sets configurable via the web interface. If nothing else, I will have a look at this when I find the time.

ourairquality:
Add a definition of `sdk_hostap_handle_timer` to address the timer callback argument issue, just does an extra indirection.

Took a pass over the html style, responsive menu, html structure, example css and gradient background and svg favicon!

Survived a day of load testing in station mode. Seems usable enough in softap mode now to get is configured.

toashwin:
Hi
This is a cool project. I have been waiting for this.
If I understand right, the purpose of this project is to have the user provide wifi ssid/password over a website - using esp8266 act as a webserver.
 
Is there a document on how to use this example?

What is the website I need to visit to configure the device?
Where will it store the ssid/password? How can I boot the device again with this new ssid/password?

Thanks

ourairquality:
@toashwin No documentation yet. There is `example/wificfg/` which you could try. It will create an AP that you can connect to, the default password is in the source code. There is no DNS yet, so connect to `http://172.16.0.1/` then follow the menu options to reconfigure the AP and/or configure the 'station' mode too.

I am not sure if there is enough RAM to run a DNS server too, getting very low here just running the DHCP server. But it will depend on what else the device is doing. It seems ok to me to ask people to connect to an IP address, not too much of a barrier, if that is too hard then configuring the device is probably beyond people too.

Have seen some other options, like Mozilla Flyweb that seems to use mDNS to advertise.

If you could report how you go it would help - I would like it to be usable by people in general.

toashwin:
@ourairquality 
I think DNS is not needed. DHCP should be good enough....  If we have an app, we can have it directly go to that IP address. 

I will try out your code, will keep you posted. 

Have you tried out, libesphttpd? why not use this? 

It is released part of Spritetm.
     Library: https://github.com/Spritetm/libesphttpd
     Example: https://github.com/Spritetm/esphttpd-freertos

I was having difficulty integrating this code into SuperHouse/esp-open-rtos.

Thanks
Ashwin

robertfoss:
DNS redirection is rather helpful.
Especially for having smartphones autodetect that the device it is
connected to has to be configured.

ourairquality:
@toashwin The URL dispatch is similar to libesphttpd. The wifisfg uses sockets and threading, not the lower level lwip interface, so a large response can be sent without coding callbacks etc. it also uses a single small stack allocated buffer for most work, so probably has lower memory usage. It also uses const storage for responses, could use a C structure to break it up with codes etc, this is lighter. People can write the interface in another web server and re-use the same sysparams if they wish, but perhaps as-is it is about a light weight as practical.

@robertfoss I've seen Android pop up the config option, not sure how that works? Would need a really lightweight DNS. It does not seem to be necessary, rather just nice, so perhaps it is something to explore in future and could be a compile-time option. Also mDNS might be even better.

robertfoss:
Included in the code I linked above is a ~100 line dns server that only
replies to some queries and only replies with the IP of the ESP.
On Android and IOS internet connectivity is determined by making http
queries to some host, if those fail, a captive portal is assumed.
Have a look at line 1111 here:
https://github.com/nodemcu/nodemcu-firmware/blob/master/app/modules/enduser_setup.c

Really, the config module is what a noob using an ESP would interact with
first, and knowing which ip (or anything about ip addresses for that
matter) reduces the utility of any config module in my mind.

toashwin:
@ourairquality @robertfoss 

I was able to flash my ESP8266 with your example. 
When I scan the available networks on my computer/ Android phone – I see the network name:  “ESP_D1DE60”

When I try to connect to this wifi network: 

On Android, it says “Obtaining IP Address”.. after some time it gives up.
On Windows, it connects and the status says “No Internet, Open”.  But when I try to visit the website: http://172.16.0.1/ , chrome browser it says, “site cant be reached”

ourairquality:
@toashwin The PR here includes an example that defaults the ssid to 'esp-open-rtos AP' so if you are not seeing that then something has gone wrong. The ssid ESP_D1DE60 must be from some other code.

toashwin:
@ourairquality  Here is some debug data I was able to print from the device. 
It seem to be getting into the code to set the SSID and PSWD. Not sure if it was really able to set.

Also, I wonder, why is it connecting to my home wifi at the end? Thought it will only act as a server..


-gSDK version:0.9.9

Here1   => sysparam_set_string SSID esp-open-rtos AP

Here4  => sysparam_set_string PSWD esp-open-rtos

sleep enable,type: 2
mode : sta(xx:fe:34:yy:de:60) + softAP(zz:fe:34:yy:de:60)
add if0
add if1
bcn 100
scandone
add 0
aid 8
cnt 

connected with “My Home Wireless”, channel 7
dhcp client start...
ip:xxx.yyy.0.9,mask:255.255.255.0,gw:xxx.yyy.0.1

ourairquality:
@toashwin Looks like a good idea to default the ssid based on the chipid, see the latest update, now the ssid is `eor-<chip id>`. Also reworked the validation and initialization of the wifi modes. Tested on an erased esp connecting from an Android device and it all worked ok. Also moved the defaulting of the ssid and password into the extras.

toashwin:
@ourairquality  Couple of things
1) Do I need to erase the ESP, before I flash wificfg? The example was not working for me. So, ran command "esptool.py erase_flash". After that I could not even connect through USB serial. I am not able to reflash it either.

For now I changed the board. 
2) My new board is already programmed for a different application. And wifi might have been enabled before in that application. 

Now when I flash it, with wificfg. And try to check my android phone.. I dont see this device.  I think this code is crashing..  I put debug statements. I added following lines @1463 in function wificfg_init in extras/wificfg.c

    printf("Here0 init\n");
    printf("wifi_sta_ssid:%s\n", wifi_sta_ssid);

Below is the crash output. 

  SDK version:0.9.9
sleep enable,type: 2
**Here0 init
wifi_sta_ssid:Fatal exception (28):** 
epc1=0x40217240
epc2=0x00000000
epc3=0x40215784
excvaddr=0x00000000
depc=0x00000000
excsave1=0x40205351
Registers:
a0 40205351 a1 3fff3b40 a2  00000000 a3  00000000
a4  ffffffff a5  40106f1c a6  3fff3bf0 a7  0000000e
a8  3fff3bd3 a9  0000000d a10 3fff155c a11 0000003a
a12 00000000 a13 3fff1514 SAR 0000001f

Stack: SP=0x3fff3b40
0x3fff3b40: 4021a422 4021a421 00000000 40106f12
0x3fff3b50: 3ffe800c a5a5a5a5 a5a5a5a5 40106f6c
0x3fff3b60: 3ffe800c 3ffe800c 3fff1514 40207ded
0x3fff3b70: a5a5a5a5 a5a5a5a5 a5a5a5a5 3fff1514
0x3fff3b80: 4021a424 00000000 3ffe800c 401072b0
0x3fff3b90: 00000000 ffffffff 00000000 00000000
0x3fff3ba0: 3fff3c83 0000000e 00302073 00000001
0x3fff3bb0: 3fff155b 3ffe800c 3fff1514 40207ded

Free Heap: 43844
_heap_start 0x3fff1498 brk 0x3fff4fc4 supervisor sp 0x3ffffae0 sp-brk 43804 bytes
arena (total_size) 15148 fordblks (free_size) 40 uordblocks (used_size) 15108

 ets Jan  8 2013,rst cause:2, boot mode:(3,6)

load 0x40100000, len 2292, room 16 
tail 4
chksum 0x57
load 0x3ffe8000, len 772, room 4 
tail 0
chksum 0x0b
csum 0x0b

rBoot v1.4.0 - richardaburton@gmail.com
Flash Size:   16 Mbit
Flash Mode:   QIO
Flash Speed:  40 MHz
rBoot Option: Big flash
rBoot Option: RTC data

Booting rom 0.
W[00]"[1B]$„|ò[1F][00]äûNòƒoÜ„noŸlNoä[1B]Nàp[12]n[12][03]ä[13]r[02][1B]b[1B]’n|[00][0C]cpbl$`[02]€[03]ò[1B]žàŽ{o[1B]oœn„[04]ì„Üœœpœ„[03][0C]B[0C]B[12]no[1E]‚[0C]Žbì‚[1B]rso>’lŒ$`[02]€’o[12]ßàrn[12]nÜNŒ[1F][00]äŒœÜÜ[1C]€[0C]Ž[04]c[04]c2Nn>‚[0C]cäƒ[12]srN[1F]ƒäœl [03][1C]"sBä[1B]žär²œ|_task_hdl:3fff4ba0, prio:12, stack:200

ESP-Open-SDK ver: 0.0.1 compiled @ Sep 20 2016 22:17:18
phy ver: 273, pp ver: 8.3

ourairquality:
@toashwin Thank you for testing this. Note `wifi_sta_ssid` will be NULL until initialized, so the string can not be printed - this is the likely cause of the crash reported. Sorry, I don't know what is going wrong after erasing the flash but it has always worked here. I might be that if something else had written to the sysparam area of the flash that it is not usable.

ourairquality:
@robertfoss Thank you. Added AP DNS support, and support to redirect host names to ip addresses. On Android the 'sign in to network' appears, but the landing page is not great yet.

toashwin:
@ourairquality 
Even if it do not print `wifi_sta_ssid`, it is crashing...  I suspect it is crashing at the following if statement. It is never reaching either of the print statements. Let me debug further. Thank you for all the work.. 

And even without any print-statements, I dont see a wifi ssid from ESP on my phone.. I suspect it is because it crashed.

```
    if (strlen(wifi_ap_ssid) <= 0 || strlen(wifi_ap_ssid) >= 32 ||
        strlen(wifi_ap_password) >= 64) {
        printf("Here4 strlen\n");
        wifi_ap_enable = 0;
}
printf("Here5 after if - wifi_ap_enable\n");
```

ourairquality:
@toashwin Thank you. Could you please check if "WARNING: Could not initialize sysparams" is being printed on startup? This might explain the crash as the wificfg currently assumes it has been initialized. I can't see this working usefully if the sysparams can not be initialized. Another explanation is a bug in the sysparam code, and I work on a different fork of that sorry.

toashwin:
@ourairquality - My problem is somewhat related to sysparams initialization. And making some modifications in wificfg will avoid crashes. 

I was experimenting with examples/sysparam_editor. What I understood is that my device has no sysparam defined. So I used sysparam_editor "reformat" option to create a sysparam region. Since then sysparam_editor worked for me.

But I still see hangs with wificfg. It may be because, `wifi_ap_ssid` is not yet defined for me. So it was crashing on following line. 
    `if (!wifi_ap_ssid) {`

sysparam_editor, does the checks following way. It uses the `status` value from the `sysparam_get_string` as below
        ` status = sysparam_get_string(cmd_buffer, &value);`

If you do that, you may avoid the crashes.

toashwin:
@ourairquality  I have verified our code.. it works now.. with the changes you have at: https://github.com/ourairquality/esp-open-rtos

I have verified that the AP is enabled. And I can see the wifi ssid on Android. I briefly looked at the webpage.. But did not verify the functionality. Will check it this weekend. 

But I felt the webpage can be made little bit cleaner.. Instead of making the webpage as a char string, can we make it a html page?

ourairquality:
@toashwin Great, thank you for giving it a try. Split out the web content into separate files, still included as the content of C char strings so it still needs to be escaped correctly, but it is a least separated a little. Might be possible to write a script to escape the html into a C string and even do this in the build process but I'll have to leave that to someone else to explore.

ourairquality:
Add a FreeRTOS tasks list page that gets compiled with the wificfg when `configUSE_TRACE_FACILITY` is defined, see the example which now uses this. Found this very helpful for checking that the stack depths were ok, and the task runtimes were interesting too.

resetnow:
>Might be possible to write a script to escape the html into a C string and even do this in the build process but I'll have to leave that to someone else to explore.

GNU linker feature can also be used to link data files directly into the binary as described here http://stackoverflow.com/a/4865249

sheinz:
I tried this PR and I'm getting a crash:
```
ESP-Open-SDK ver: 0.0.1 compiled @ Nov 23 2016 22:06:20
phy ver: 273, pp ver: 8.3

SDK version:0.9.9
sleep enable,type: 2
Fatal exception (28):
epc1=0x4020ad1c

0x4020ad1c: wificfg_init at /Volumes/esp8266/esp-open-rtos/extras/wificfg/wificfg.c:1438

epc2=0x00000000
epc3=0x00000000
excvaddr=0x00000000
depc=0x00000000
excsave1=0x4020ad19

0x4020ad19: wificfg_init at /Volumes/esp8266/esp-open-rtos/extras/wificfg/wificfg.c:1434

Registers:
a0 4020ad19 a1 3fff3c90 a2  00000000 a3  3fff1324

0x4020ad19: wificfg_init at /Volumes/esp8266/esp-open-rtos/extras/wificfg/wificfg.c:1434

a4  00000001 a5  3fff5048 a6  04000002 a7  3feffe00
a8  00000100 a9  0000001a a10 00000018 a11 04000102
a12 40218548 a13 00000050 SAR 00000014

0x40218548: _malloc_usable_size_r at ??:?


Stack: SP=0x3fff3c90
0x3fff3c90: 30726f65 64616330 00356261 00000000
0x3fff3ca0: 00320001 0000000d 33302064 34666666
0x3fff3cb0: 00303032 a5a578a5 00000000 a5a5a5a5
0x3fff3cc0: 3fff3d20 3fff3d10 00000004 00000001
0x3fff3cd0: 4021c955 00000001 00000000 00000003

0x4021c955: strchr at ??:?

0x3fff3ce0: 3fff3d40 3fff3d30 00000008 00000000
0x3fff3cf0: 00000000 4021c954 3ffe800c 4010606a

0x4021c954: strchr at ??:?


0x4010606a: printf at /home/gus/dev/esp/rtos/newlib/build/xtensa-lx106-elf/newlib/libc/stdio/../../../../../newlib/libc/stdio/printf.c:61

0x3fff3d00: 3fff3d20 3fff3d10 00000004 00000000

Free Heap: 43828
_heap_start 0x3fff1460 brk 0x3fff5048 supervisor sp 0x3ffffb00 sp-brk 43704 bytes
arena (total_size) 15336 fordblks (free_size) 124 uordblocks (used_size) 15212
```

ourairquality:
@sheinz Updated this PR with the latest code. If it still crashes then one likely difference is the sysparam code. If you had the time then you could try to cherry-pick the sysparam commit from https://github.com/ourairquality/esp-open-rtos and see if that fixes the crash.

sheinz:
@ourairquality I pulled **sysparam** changes from your fork and **wificfg** seems to work fine. Do you mind if I open a PR for your **sysparam** changes? Because this PR without updated **sysparam** will not work. Moreover you've added a really good stuff in the **sysparam** and it would be nice to have it here.

ourairquality:
@sheinz Yes, you are welcome to make whatever use you want of the code, and if you can finesse landing some improvement that would be great. Many of the changes were to work around alignment issues with the SPI flash code, but it was later pointed out that there are other functions that have already addressed this and it might be better to someday re-factor the code to use them.

On your sysparm branch: I think it is better to not use strings for bool values because even reading a sysparam string requires dynamic memory allocation which might require locking and may well fail, so it seems much better to just use a numeric value which can be safely read in more contexts. Also my fork uses a single byte of storage for a uint8 parameter value, rather than a uint32. But I have my own fork, so do what you wish, and if the API can be compatible then that would be great, and I can add a bool API for my fork that uses an uint8 for storage to keep other code compatible.

malachib:
Very late in the game I know, but I must chime in and say it is *critical* for a provisioning feature like this to go over some kind of encryption channel.  With @ourairquality awesome LWIPv2 work, that moves us closer . 

LWIPv2 contains lwip/apps/httpd code with an HTTPS mode -- **BUT** it is not working out of the box and will need our attention

ourairquality:
@malachib There are probably many use cases in which the wifi security is more than adequate, and the device ram is very limited, so this wificfg option is not planned to use https and rather be one very lightweight option.

It seems secure to setup the device when it is running as a secure access point and then there is an option to disable the configuration. It also seems adequate on a trusted local network. The logic could be extended to allow configuration only via the AP netif.

I am not even sure if running a web server on the device is a great solution. It should be possible to keep all the web side code on another device in a web app. But people have many different uses. I am interested to learn more about standards that emerge.

malachib:
For non-provisioning use cases, I agree, which I surmise covers at least 75% of this thread.  The thing is, the inbuilt WiFi security isn't specifically the issue.

Let's look at provisioning use cases using HTTP:

Case 1: Open AP, user configures ssid + pass.  
--Pass travels in plaintext, so this use case fails immediately
Case 2: Closed AP, user configures ssid + pass over secured WiFi AP.   Mass produced product
--In this case, password is distributed to get into the AP.  This widens attack surface, and likelihood someone discovers the password is high (think DeCSS).  Not a fail outright, but problematic
Case 3: Closed AP, user configures ssid + pass over secured WiFi AP.   One-off/hobby product
--In this case, password distribution is a very tight loop.  So Case 3 is tenable.

Evaluating these cases we see that the only case where you are protected is a low-volume one-off scenario

ourairquality:
@malachib I basically agree with you, but I am at the 'case 3' at present. I recall seeing some work that had the station mode configured from a web browser and it was almost automated, see https://wiki.mozilla.org/FlyWeb and this involved mdns, but that work might have all been canned. Sorry I am just not an expert on where things are heading but I am keen to follow solutions.

malachib:
I plan to open a separate issue/pull request to really get into the nuts and bolts of this rather than hijacking your thread, but to summarize: HTTPS RAM usage is actually very manageable because it can be shut off most of the time.

ourairquality:
@malachib fwiw I think I can get you around 8k more ram, but it is sdk static memory at known offset from the `sdk_wDevCtrl` - the sdk seems to allocate 8k of TX buffers that are not used!

malachib:
Wow, that's a lot! Sounds too good to be true...
You should get on the esp-open-rtos irc channel on freenode...

PR #209, opened 2016-08-29
sysparams: get/set int32 and int8 apis.

ourairquality:
Rename the get/set_int api functions to get/set_int32.

Add get/set_int8 api functions too in anticipation of more efficient implementations for these.

### DIFF
--- a/core/include/sysparam.h
+++ b/core/include/sysparam.h
@@ -169,7 +169,7 @@ sysparam_status_t sysparam_get_info(uint32_t *base_addr, uint32_t *num_sectors);
  */
 sysparam_status_t sysparam_get_data(const char *key, uint8_t **destptr, size_t *actual_length, bool *is_binary);
 
-/** Get the value associate with a key (static buffers only)
+/** Get the value associated with a key (static buffers only)
  *
  *  This performs the same function as sysparam_get_data() but without
  *  performing any memory allocations.  It can thus be used before the heap has
@@ -231,8 +231,7 @@ sysparam_status_t sysparam_get_string(const char *key, char **destptr);
 /** Get the int32_t value associated with a key
  * 
  *  This routine can be used if you know that the value in a key will (or at
- *  least should) be an integer value.  It will parse the stored data as a
- *  number (in standard decimal or "0x" hex notation) and return the result.
+ *  least should) be an int32_t value.
  *
  *  Note: If the status result is anything other than ::SYSPARAM_OK, the value
  *  in `result` is not changed.  This means it is possible to set a default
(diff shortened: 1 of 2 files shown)


foogod:
Changing the name of the get/set_int to int32 is fine by me.. I'm not sure what kind of "more efficient" implementation you have in mind for int8, though, since even if we read/write it as a single binary byte, it will still take up 4 bytes of payload space (due to limitations in the ESP8266 flash write routines/hardware), so it will still not be any smaller or more efficient than an int32..

ourairquality:
@foogod We can write and read a single byte from the flash, so an int8 is more efficient in terms of storage. There is an alignment restriction on the source pointer for `sdk_spi_flash_write` but no restriction on the destination alignment or the size. One suggestion is to use a bounce buffer to overcome the source alignment restriction.

foogod:
That is actually not true (I've tested it).  Regardless of what length you specify, the flash write routines will always write a multiple of 4 bytes out to the flash (including trailing garbage if you haven't set anything for those bytes in the source buffer).  I did a little bit of digging into the routines and it appears this is actually a limitation of the dedicated SPI flash write-from-memory hardware built into the ESP8266 which the routines use.  This is why sysparam always writes everything (including payloads) in multiples of 4 bytes, because otherwise it resulted in corruptions due to this limitation.

foogod:
Though it does occur to me that now that we're doing all writes through the bounce buffer, we could conceivably do short writes by padding the extra bytes with 0xFF in the bounce buffer before submitting the write call..  They would technically still get written, but would have no effect..  (would want to do a fair bit of testing to verify that works as expected, and would require changing several other places in the code where it rounds-up payload sizes, etc..)

If we could get that working, then yeah, maybe an int8 version of the routines would make more sense..

ourairquality:
@foogod Ok, thank you again, I see a lot of alignment issues now. So the flash address needs to be aligned if the write can cross a flash page boundary, the source address and size need to be word aligned. The flash read need to be page aligned and multiples of the word size. Made some updates to my sysparam code using bounce buffers, so entries no long need to be word aligned and it can store a single byte payload more efficiently (if that is really significant), should also save a little payload space for strings hear and there.

sheinz:
There's already an [alternative implementation](https://github.com/SuperHouse/esp-open-rtos/blob/master/extras/spiffs/esp_spiffs_flash.h) of SPI flash read/write that was written for SPIFFS integration. It doesn't require any alignment and allows read/write with byte precision.

Can it be of any use for sysparams?

kanflo:
@sheinz, I for one hade missed your alternative implementation. It could for sure be used in sysparam (PRs welcome :) and we might need to do more advertisement for it. I would not mind having it moved to core.

kanflo:
I think this one could be merged if others agree. I think we should merge sooner rather than later and agree on a frozen sysparam API before it sees widespread use.

Issue #60, opened 2015-10-28
Support for GPIO 16 (and beyond)


kanflo:
One thing we are currently lacking is support for handling GPIO > 15. Looking at the Arduino port, GPIO 16 is handled in other registers than GPIOs 0..15:

```
extern void __digitalWrite(uint8_t pin, uint8_t val) {
  if(pin < 16){
    if(val) GPOS = (1 << pin);
    else GPOC = (1 << pin);
  } else if(pin == 16){
    if(val) GP16O |= 1;
    else GP16O &= ~1;
  }
}
```

`GPIOS` above corresponds to our `GPIO.OUT_SET` and `GP16O` resides at 0x60000768 where we have no mapping.

projectgus:
Yes, the standard GPIO interface in core/esp/gpio.h should support number 16.

GPIO16 is connected internally to the RTC hardware not the GPIO hardware, which is why it can be used for RTC wakeup and things like that.

@foogod has mapped the control registers in the open-libmain branch:
https://github.com/SuperHouse/esp-open-rtos/blob/open-libmain/core/include/esp/rtc_regs.h#L34

He's mentioned this file can probably be brought over into master separately to the rest of the branch.

Some investigation probably still required to add register bit mappings for RTC.GPIO_CONF, etc.

kanflo:
Ahaa, that explains why GPIO 16 got the special treatment. Right now I have all the GPIOs I need but if I need #16 and none else has looked into this I'll have a go at it.

projectgus:
I just opened a duplicate issue and dumped out some more thoughts about this. Repeating below:

Our `gpio.h` functions currently only support GPIOs 0-15. GPIO16 is controlled by the RTC register bank and is "special" (connected to RTC hardware, alive during sleep, has soft pulldown instead of pullup, takes more clock cycles to toggle, etc.)

However, people using the gpio_xxx wrapper functions should be able to control gpio16 same as the other GPIOs.

My suggested design would be:
* We add specific `gpio16_enable`, `gpio16_read`, etc. functions specific to gpio16. Could be defined in `gpio.h` or maybe in `rtc.h`, as relates to RTC hardware...
* We add `if(gpio == 16)` tests to all generic the `gpio.h` functions that call the gpio16_xxx version if necessary.

... but I'd be open to other designs.

kanflo:
[...]

PR #211, opened 2016-08-30
GPIO16 support.

ourairquality:
Need this to blink the LED on nodemcu boards! Was discussed here https://github.com/SuperHouse/esp-open-rtos/issues/60 Have not integrated this with the other gpio functions to avoid a performance regression, and if desired then perhaps that could be a separate PR.

### DIFF
--- a/core/esp_gpio.c
+++ b/core/esp_gpio.c
@@ -5,9 +5,26 @@
  * BSD Licensed as described in the file LICENSE
  */
 #include <esp/gpio.h>
+#include <esp/rtc_regs.h>
 
 void gpio_enable(const uint8_t gpio_num, const gpio_direction_t direction)
 {
+    if (gpio_num == 16) {
+        RTC.GPIO_CFG[3] = (RTC.GPIO_CFG[3] & 0xffffffbc) | 1;
+        RTC.GPIO_CONF = (RTC.GPIO_CONF & 0xfffffffe) | 0;
+        switch (direction) {
+        case GPIO_INPUT:
+            RTC.GPIO_ENABLE = (RTC.GPIO_OUT & 0xfffffffe);
+            break;
+        case GPIO_OUTPUT:
+        case GPIO_OUT_OPEN_DRAIN:
+            /* TODO open drain? */
+            RTC.GPIO_ENABLE = (RTC.GPIO_OUT & 0xfffffffe) | 1;
+            break;
+        }
+        return;
+    }
+
     switch (direction) {
     case GPIO_INPUT:
         GPIO.ENABLE_OUT_CLEAR = BIT(gpio_num);
--- a/core/include/esp/gpio.h
+++ b/core/include/esp/gpio.h
@@ -10,6 +10,7 @@
 #define _ESP_GPIO_H
 #include <stdbool.h>
 #include "esp/gpio_regs.h"
+#include "esp/rtc_regs.h"
 #include "esp/iomux.h"
 #include "esp/interrupts.h"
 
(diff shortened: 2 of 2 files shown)


ourairquality:
Perhaps performance would not be impacted if the gpio number is a constant, because the comparison and dead code would be optimized away. Happy to just bundle this in the gpio_read and gpio_write is that is what people want?

resetnow:
For static inline functions (read/write) yes, branching should be eliminated as a part of constant propagation. And gpio_enable is unlikely to be called very often so I think performance impact will be minimal.

foogod:
Yeah, I think including it as a conditional in the standard `gpio_read`/`gpio_write` would be fine, actually..  If someone's concerned about speed there they should already be using a compile-time constant for pin number (as it optimizes out a lot of the other internals of those functions anyway) and so this would be just another compile-time optimization, so that should be fine..

(Alternately, I'm fine with merging this in as-is for now too if you don't want to make that change for some reason.. just let me know)

ourairquality:
Thank you for the feedback. Actually the i2c driver might be an example that is compromised by poor performance in these paths and uses a dynamic pin configuration, and it also uses the `gpio_enable` function to toggle pins high an low rather than using open drain. There is already an open PR to write a separate 'fast' i2c driver.

I was even thinking of abstracting the i2c driver to support multiple i2c buses so wanted to keep the dynamic pin definitions. The gpio16 pin seems to be in a different class, so perhaps it's fine with separate api functions.

So have split out `gpio16_enable` too for now. Perhaps we need to add inline api functions to switch between input and output on a pin quickly too, without all the configuration too.

foogod:
Hmm, actually now that you mention it, things like the i2c driver also seem to be a good argument for including GPIO16 into the main routines if we can manage it.  It would be really nice if somebody could, for example, do I2C over GPIO16 by just passing "16" to the i2c driver, which they can't do if we keep things separate..

I haven't actually looked at the i2c driver in a while, but if it's using `gpio_enable` to switch modes instead of open-drain then that's probably something we should fix in the i2c driver (and shouldn't be using that bad design as a reason for design decisions elsewhere in the code)..

Having an extra conditional in these routines probably isn't the end of the world either, even if the value isn't a compile-time constant, using things like `__builtin_expect` we should be able to ensure that the GPIO0-15 code remains the fast-path and it will only introduce an extra 2 opcodes (25ns at 80MHz) into that path.  For something like the i2c driver I really doubt that will make any difference to anyone (and if it does we can probably more than make up for it with improvements in the i2c driver (like not using `gpio_enable` to switch modes) instead)..

ourairquality:
@foogod Ok, thanks. Changed the PR to bundle the gpio16 support into the existing api functions. Looking at the I2C driver I also don't think the overhead is significant, and it has delays after changes anyway. Fir the I2C driver, changing the pin from input to output rather than using open drain was a feature, there was a discussion somewhere on the merits of the approach.

resetnow:
Speaking about performance, I have seen mentions of rather slow MMIO register access — people say it takes about <a href="http://www.esp8266.com/viewtopic.php?f=156&t=11275">17 cycles to write a peripheral register</a>. Pure speculation here, but I think cycles could be lost in the glue logic between lx106 memory interface and peripheral bus, which probably doesn't have pipeline support implemented. So these extra branch delays are quite small compared to IO access time.

projectgus:
This looks good to me, thanks @ourairquality . +1 to merge here if noone has any other concerns.

@resetnow , probably the most relevant thing about the ESP8266 CPU is that it has the xtensa "write buffer" feature, with buffer size one. So there can always be one peripheral write "in flight" at a given time without blocking the CPU, as long as it doesn't issue another write that blocks behind it (or a MEMW instruction). This means that some additional conditional tests like the one mentioned may not cause any performance overhead, even if they're not known at compile time, because in a tight loop the conditional can be evaluated during the window when the write buffer is still flushing. I wouldn't guarantee this without measuring it, but that's my suspicion!

PR #212, opened 2016-08-30
esp/interrupts.h: Add RTC interrupt number.

pfalcon:
40002a58     $a3 = rtc_intr_handler
40002a5e     $a4 = 0x0
40002a60     $a2 = 0x3
40002a62     call ets_isr_attach


pfalcon:
With this, asisgnment of interrupt numbers 0-10, are known, which is pretty good. Per BootROM functions, there's 14 total settable interrupt vectors, 13 level1 + NMI. So, we know 11 of 13 level1 vector assignments.. Unknown vectors 11, 12. Of course, they can be unused actually.

projectgus:
Great, thanks!

PR #213, opened 2016-08-31
sysparam: export the 'compact' function, added it to the editor.

ourairquality:
Useful after editing parameters to get best read performance. Also to clear out the history, such as old passwords, but would take two calls for that.

### DIFF
--- a/core/sysparam.c
+++ b/core/sysparam.c
@@ -407,7 +407,7 @@ static inline sysparam_status_t _delete_entry(uint32_t addr) {
  *  the result to the alternate region, then make the new alternate region the
  *  active one.
  *
- *  @param key_id  A pointer to the "current" key ID.
+ *  @param key_id  A pointer to the "current" key ID, or NULL if none.
  *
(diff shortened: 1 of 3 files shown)


core/sysparam.c
```diff
@@ -634,6 +636,22 @@ sysparam_status_t sysparam_get_info(uint32_t *base_addr, uint32_t *num_sectors)
...
+        status = SYSPARAM_ERR_NOINIT;
+        goto done;
+    }
+
+    status = _compact_params(NULL, NULL);
+
+ done:
+    xSemaphoreGive(_sysparam_info.sem);
+    return status;
+}
```
sheinz: Why do you need a **goto** here?

Isn't it the same?
```
    if (_sysparam_info.cur_base) {
        status = _compact_params(NULL, NULL);
    } else {
        status = SYSPARAM_ERR_NOINIT;
    }
    xSemaphoreGive(_sysparam_info.sem);
    return status;
}
```

ourairquality:
Fixed, thank you for the reminded.

PR #214, opened 2016-09-01
i2c: use open drain for output rather than flipping between input and output.

ourairquality:
Using gpio_enable and flipping the configuration appears to be a major cause of poor performance in the i2c driver, so this changes to using the outputs as open-drain only.

### DIFF
--- a/extras/i2c/i2c.c
+++ b/extras/i2c/i2c.c
@@ -49,48 +49,47 @@ void i2c_init(uint8_t scl_pin, uint8_t sda_pin)
     gpio_set_pullup(g_scl_pin, 1, 1);
     gpio_set_pullup(g_sda_pin, 1, 1);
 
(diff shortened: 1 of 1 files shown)


Zaltora:
Great fix, the performance gain is significant. I did a test with DS3231 example with a delay of 1 us from sdk :

projectgus:
Thanks @ourairquality, this looks good to me! :+1:

sheinz:
Thanks @ourairquality.

Has anyone tried GPIO 16 with i2c already?

ourairquality:
@sheinz It's not expected for work on GPIO 16 now as GPIO 16 has no open drain support (or at least it's not in the code yet).

Zaltora:
One other inconvenient of GPIO 16, this pin is very slow to switch between state.

PR #215, opened 2016-09-04
esp/gpio_regs.h: Add info about strapping pins captured in GPIO.IN register.

pfalcon:
Specifically, in the high 16 bits of GPIO.IN register.


sheinz:
@pfalcon Thank you for the contribution!

PR #217, opened 2016-09-06
Comment the sdk_station_config slots.

ourairquality:
See mention in some other esp code that if the bssid_set value is not zero then the bssid must be set. Should the examples be initializing this to zero explicitly?

### DIFF
--- a/include/espressif/esp_sta.h
+++ b/include/espressif/esp_sta.h
@@ -15,10 +15,10 @@ extern "C" {
 #endif
 
 struct sdk_station_config {
-    uint8_t ssid[32];
-    uint8_t password[64];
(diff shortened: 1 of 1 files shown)


projectgus:
The C99 standard specifies that if there's an intiializer list for an aggregate type (aggregates include structs) then any non-specified fields are initialized to zero. So this PR doesn't change the compiler output.

That said, there is a bit of a gap around understanding of these structures. I think it would be really useful to have some comments in include/espressif/esp_sta.h explaining what each field should be set to (including details like "bssid_set should be 0 if using SSID, 1 if using bssid").

ourairquality:
Thanks. Changed the PR to just add some comments as suggested.

PR #218, opened 2016-09-08
wifi_get/set_sleep_type: add source code implementations plus fixes.

ourairquality:
- wifi_set_sleep_type returns a bool success flag.
- wifi_get_sleep_type seemed useless, just returning an argument. Added an implementation using sdk_pm_get_sleep_type.

### DIFF
--- a/open_esplibs/libmain/user_interface.c
+++ b/open_esplibs/libmain/user_interface.c
@@ -592,4 +592,16 @@ void sdk_system_uart_de_swap()
     DPORT.PERI_IO &= ~DPORT_PERI_IO_SWAP_UART0_PINS;
 }
 
+enum sdk_sleep_type sdk_wifi_get_sleep_type()
+{
+    return sdk_pm_get_sleep_type();
+}
+
+bool sdk_wifi_set_sleep_type(enum sdk_sleep_type type)
+{
+    if (type > WIFI_SLEEP_MODEM) return false;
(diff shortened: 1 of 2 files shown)


open_esplibs/libmain/user_interface.c
```diff
@@ -592,4 +592,19 @@ void sdk_system_uart_de_swap()
...
 
+enum sdk_sleep_type sdk_pm_get_sleep_type(void);
+void sdk_pm_set_sleep_type_from_upper(enum sdk_sleep_type);
+
+enum sdk_sleep_type sdk_wifi_get_sleep_type()
+{
+    return sdk_pm_get_sleep_type();
+}
+
+bool wifi_set_sleep_type(enum sdk_sleep_type type)
```
projectgus: do you mean sdk_wifi_set_sleep_type here?

projectgus:
This looks good to me. @foogod do you have any thoughts on this?

Power management in version 0.9.9 of the Espressif SDK wasn't very well documented. Do you know if using these functions has much effect on power consumption?

ourairquality:
Yes, updated the name but the definitions added seems to be best placed in libmain.h and were moved there in https://github.com/SuperHouse/esp-open-rtos/pull/267 so this depends on https://github.com/SuperHouse/esp-open-rtos/pull/267 landing first. Once we get the definitions landed it will be possible to start landing some of these.

PR #220, opened 2016-09-15
Bmp280 option

Zaltora:
Add skip option and possibility to select oversampling for temperature with bmp/bme280  (see datasheet)

### DIFF
--- a/extras/bmp280/bmp280.h
+++ b/extras/bmp280/bmp280.h
@@ -94,8 +94,8 @@ typedef enum {
 typedef struct {
     BMP280_Mode mode;
     BMP280_Filter filter;
-    BMP280_Oversampling oversampling_pressure;  // pressure oversampling
-    BMP280_Oversampling oversampling_temperature;  // temperature oversampling
+    BMP280_Oversampling oversampling_pressure;
+    BMP280_Oversampling oversampling_temperature;
     BMP280_Oversampling oversampling_humidity;
     BMP280_StandbyTime standby;
 } bmp280_params_t;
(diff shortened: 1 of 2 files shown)


extras/bmp280/bmp280.h
```diff
@@ -93,7 +94,8 @@ typedef enum {
 typedef struct {
     BMP280_Mode mode;
     BMP280_Filter filter;
-    BMP280_Oversampling oversampling;  // pressure oversampling
+    BMP280_Oversampling oversampling_pressure;  // pressure oversampling
```
sheinz: The comments seems redundant now :)

extras/bmp280/bmp280.h
```diff
@@ -93,7 +94,8 @@ typedef enum {
 typedef struct {
     BMP280_Mode mode;
     BMP280_Filter filter;
-    BMP280_Oversampling oversampling;  // pressure oversampling
+    BMP280_Oversampling oversampling_pressure;
+    BMP280_Oversampling oversampling_temperature;
```
sheinz: Enum `BMP280_Oversampling` was initially create only for pressure oversampling values.  Even though temperature oversampling value is also 3 bits wide the meaning of the enum values are not quite the same. For example from the datasheet temperature oversampling values 2 is used for "Ultra Hi Resolution" pressure measurement. 

I originally didn't provide option for temperature oversampling because the datasheet recommends to select temperature oversampling depending on the desired pressure resolution.

Zaltora: If people want for example, read faster the pressure, they can disable other measurement like temperature. 
Lets open temperature oversampling for more custom choice. User will probably do some tests if professional usage.
Maybe we can preset an optimized selection ? or comment the tab from datasheet.

PR #221, opened 2016-09-15
timers: esp library timer callbacks expect their argument not a xTimerHandle

ourairquality:
Fixes https://github.com/SuperHouse/esp-open-rtos/issues/140 and probably lots of other issues.


sheinz:
I also discovered this issue and it was discussed in https://github.com/SuperHouse/esp-open-rtos/issues/176 

We might eventually leave single implementation of timers. Currently there are two kind of timers: **sdk_os_timer_** and **sdk_ets_timer_**

For now the fix is good for me. Thanks @ourairquality!

kanflo:
Thanks @ourairquality!

PR #222, opened 2016-09-15
core: add phy regs definition

resetnow:
I plan to implement some of libphy functions in C later, so I thought it would be good to have an attempt of phy registers description.

Also, if somebody is interested in doing something similar, radare2 has recently been updated with better xtensa support — I posted a message <a href="https://groups.google.com/forum/#!topic/esp8266-re/cSWvEex6hpI">here</a>.


sheinz:
Thanks @resetnow!

PR #224, opened 2016-09-16
libnet80211: start an open_esplib library.

ourairquality:
Would like to start add some functions for libraries other than libmain, and here's two I wrote recently.

Need help with how to actually get this to be linked. It seems to get built ok, just not used?


ourairquality:
Solved the makefile problem, sorry for the noise. The `*_DEFAULT_ROOT` was not correct, but we have `open_esplibs_ROOT` so using that.

ourairquality:
Unfortunately still got problems here, the new functions are not used in preference to the sdk functions. Have been able to get one of them to be used, buy adding a dummy function in the same object file, and referencing it from the application, but even this does not help with the other. Linker challenge I assume?

ourairquality:
Added some hack references to demonstrate how I've seen these used. Perhaps someone with good knowledge of the linker can stop the problem? But any there is a hack workaround. Once a complete object file had open definitions I presume removing it would also address the problem, so the hack might only be temporary.

open_esplibs/libnet80211/ieee80211_hostap.c
```diff
@@ -0,0 +1,34 @@
...
+
+void sdk_ieee80211_send_mgmt(void *, int, int);
+void sdk_cnx_node_leave(struct sdk_g_ic_netif_info *netif, struct sdk_netif_conninfo *conn);
+
+void sdk_hostap_handle_timer(struct sdk_netif_conninfo *cnx_node)
+{
+    uint32_t count1 = *(uint32_t *)(((void *)cnx_node) + 0x108);
+    uint32_t count2 = TIMER_FRC2.COUNT;
+
+    if ((count2 - count1) > 93599688U) {
```
sheinz: @ourairquality You seems to have optimised the code here :)
I looked into the disassembly of that part and don't quite understand the intended logic:
```
            l32i         a0, a12, 0x108 /* [arg0+0x108] */     #  52b: 022c42
// a0 = cnx_node->0x108
            l32i         a2, a2, 36 /* [0x60000624] */         #  52e: 2892
// a2 = FRC2.COUNT
            bltu         a2, a0, .Lbr021                       #  530: 07320e
// if (a2 >= a0) {
            movi         a3, 312 /* 0x00000138 */              #  533: 32a138
            sub          a2, a2, a0                            #  536: 0022c0
            call0        __udivsi3                             #  539: 01f3ffc0
                                                               #  53d: 0000
            j            .Lbr022                               #  53f: 460400
// if (a2 < a0) {
.Lbr021:    movi         a3, -1 /* 0xffffffff */               #  542: 7cf3
            xor          a3, a0, a3                            #  544: 303030
            add          a2, a2, a3                            #  547: 3a22
            addi         a2, a2, 1                             #  549: 1b22
            movi         a3, 312 /* 0x00000138 */              #  54b: 32a138
            call0        __udivsi3                             #  54e: 01efffc0

.Lbr022:    movi         a4, 0x000493df                        #  554: 41e9ff
            bgeu         a4, a2, .Lbr023                       #  557: 27b43b

```
So, the the logic seems to be the following:
```
if (curr_time < timestamp) {
    t = (curr_time + (~timestamp) + 1) / 312;
} else {
    t = (curr_time - timestamp) / 312;
}

if ((300000 - 1) < t) {  // (300000-1) == 0x000493df
```
But the expression `(~timestamp) + 1` equals `-timestamp`. So, both branches basically do the same.
@ourairquality Do you have any clues what might be the reason behind all this?

BTW, great job on reverse engineering those functions!

sheinz:
I have no idea why those functions are not linked instead of the Espressif's.
The build stage 2 that renames symbols in the Espressif's libraries are also weakens the symbols, so they should be replaced with no-weak versions.
```
SDK processing stage 2: Renaming symbols in SDK library build/sdklib/libnet80211_stage1.a -> build/sdklib/libnet80211.a
xtensa-lx106-elf-objcopy --redefine-syms ../../lib/allsymbols.rename --weaken build/sdklib/libnet80211_stage1.a build/sdklib/libnet80211.a
```
The linking workaround might get unnecessary eventually when enough of libnet80211 will be open source and it will be directly referenced from other source code.

open_esplibs/libnet80211/wl_cnx.c
```diff
@@ -0,0 +1,37 @@
...
+#if OPEN_LIBNET80211_WL_CNX
+// The contents of this file are only built if OPEN_LIBNET80211_WL_CNX is set to true
+
+#include "espressif/esp_misc.h"
+#include "sdk_internal.h"
+#include <string.h>
+
+void *sdk_cnx_node_search(uint8_t mac[6])
+{
+    int count = sdk_g_ic.s._unknown310 + 2;
```
JSStats: The disassembly did load this slot using `l8ui` so changing to signed would not appear to be appropriate and would cause a sign extension for values over 0x7f. The `count < 1` test does not look right but I don't think it is ever triggered, so until we know more perhaps it's best left as-is, perhaps adding a comment?

PR #228, opened 2016-09-24
Add a TSL2561 driver and example usage

bschwind:
This pull request adds support for the [TSL2561 Light Sensor](https://cdn-shop.adafruit.com/datasheets/TSL2561.pdf)

It uses the I2C module from the `extras` directory of this project.

This driver supports setting an integration time from the presets of (13ms, 101ms, and 402ms), as well as a 1X and 16X gain. The driver uses `vTaskDelay` when the lux value is requested from the sensor so it has time to integrate the sensors and get a reading. The sensor remains in a "powered down" mode when not in use.

Interrupts and custom integration times are not supported in this PR.

### DIFF
--- a/extras/tsl2561/tsl2561.c
+++ b/extras/tsl2561/tsl2561.c
@@ -168,7 +168,7 @@ void tsl2561_init(tsl2561_t *device)
     disable(device->i2c_addr);
 }
 
-void tsl2561_set_integration_time(tsl2561_t *device, uint8_t integration_time_id)
+void tsl2561_set_integration_time(tsl2561_t *device, tsl2561_integration_time_t integration_time_id)
 {
     enable(device->i2c_addr);
     write_register(device->i2c_addr, TSL2561_REG_TIMING, integration_time_id | device->gain);
@@ -177,7 +177,7 @@ void tsl2561_set_integration_time(tsl2561_t *device, uint8_t integration_time_id
     device->integration_time = integration_time_id;
 }
 
-void tsl2561_set_gain(tsl2561_t *device, uint8_t gain)
+void tsl2561_set_gain(tsl2561_t *device, tsl2561_gain_t gain)
 {
     enable(device->i2c_addr);
     write_register(device->i2c_addr, TSL2561_REG_TIMING, gain | device->integration_time);
--- a/extras/tsl2561/tsl2561.c
+++ b/extras/tsl2561/tsl2561.c
@@ -10,7 +10,93 @@
 #include "task.h"
 #include "tsl2561.h"
 
-bool write_register(uint8_t i2c_addr, uint8_t reg, uint8_t value)
+// Registers
+#define TSL2561_REG_COMMAND 0x80
+#define TSL2561_REG_CONTROL 0x00
+#define TSL2561_REG_TIMING 0x01
+#define TSL2561_REG_THRESHOLD_LOW_0 0x02
+#define TSL2561_REG_THRESHOLD_LOW_1 0x03
+#define TSL2561_REG_THRESHOLD_HIGH_0 0x04
+#define TSL2561_REG_THRESHOLD_HIGH_1 0x05
+#define TSL2561_REG_INTERRUPT 0x06
+#define TSL2561_REG_PART_ID 0x0A
+#define TSL2561_REG_CHANNEL_0_LOW 0x0C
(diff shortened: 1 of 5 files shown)


extras/tsl2561/tsl2561.h
```diff
@@ -0,0 +1,125 @@
...
+ */
+
+#ifndef __TSL2561_H__
+#define __TSL2561_H__
+
+#include <stdint.h>
+#include <stdbool.h>
+
+// I2C Addresses
+#define TSL2561_I2C_ADDR_VCC 0x49
```
sheinz: All hardware related defines can be moved to .c file as it is not part of the driver's interface.

extras/tsl2561/tsl2561.c
```diff
@@ -0,0 +1,287 @@
...
+ * BSD Licensed as described in the file LICENSE
+ */
+
+#include <stdio.h>
+#include "FreeRTOS.h"
+#include "i2c/i2c.h"
+#include "task.h"
+#include "tsl2561.h"
+
+bool write_register(uint8_t i2c_addr, uint8_t reg, uint8_t value)
```
sheinz: All internal functions could be made static.

extras/tsl2561/tsl2561.c
```diff
@@ -0,0 +1,287 @@
...
+            }
+            else if (ratio > K8T)
+            {
+                b = B8T;
+                m = M8T;
+            }
+
+            break;
+        default:
+            printf("Invalid package type in CalculateLux\n");
```
sheinz: In this case the function will still return true.

sheinz:
Thanks @bschwind for the contribution!

This looks good apart from a few minor comments.

extras/tsl2561/tsl2561.h
bschwind: I moved most of the defines into the .c file, but there are a few I left in the header which you can use when calling particular TSL2561 functions, such as when choosing an I2C address or an integration time constant.

extras/tsl2561/tsl2561.c
bschwind: Fixed!

extras/tsl2561/tsl2561.c
bschwind: Good catch!

bschwind:
@sheinz Thanks for the review!

I addressed your comments and updated the PR. When it's ready to merge I'll squash it back down to one commit, if that's preferred.

extras/tsl2561/tsl2561.h
UncleRus: I tnink enums would be better:
```C
typedef  enum {
    TSL2561_GAIN_1X = 0x00,
    TSL2561_GAIN_16X = 0x10
} tsl2561_gain_t;

...

void tsl2561_set_gain(tsl2561_t *device, tsl2561_gain_t gain);
```

extras/tsl2561/tsl2561.h
```diff
@@ -0,0 +1,39 @@
...
+// Integration time IDs
+#define TSL2561_INTEGRATION_13MS 0x00
+#define TSL2561_INTEGRATION_101MS 0x01
+#define TSL2561_INTEGRATION_402MS 0x02 // Default
+
+// Gain IDs
+#define TSL2561_GAIN_1X 0x00
+#define TSL2561_GAIN_16X 0x10
+
+typedef struct {
```
UncleRus: It's hard to use in cpp without 
```C
#ifdef __cplusplus
extern "C" {
#endif
```

bschwind: To be clear, should the final code end up like this?

```
#ifdef  __cplusplus
extern "C" {
#endif

typedef struct {
    tsl2561_i2c_addr_t i2c_addr;
    uint8_t integration_time;
    uint8_t gain;
    uint8_t package_type;
} tsl2561_t;

#ifdef  __cplusplus
}
#endif
```

Or should the `extern "C"` block cover all symbols in tsl2561.h?

resetnow: It should cover all symbols (so C++ compiler knows no name mangling should be used).

bschwind:
@sheinz @UncleRus @resetnow, Thanks for your reviews and suggestions

PR #232, opened 2016-10-09
DS1307 RTC driver + example

UncleRus:
Driver for the most common RTC DS1307.

### DIFF
--- a/extras/ds1307/ds1307.h
+++ b/extras/ds1307/ds1307.h
@@ -1,0 +1,114 @@
+/*
+ * Driver for DS1307 RTC
+ *
+ * Part of esp-open-rtos
+ * Copyright (C) 2016 Ruslan V. Uss <unclerus@gmail.com>
+ * BSD Licensed as described in the file LICENSE
+ */
+#ifndef EXTRAS_DS1307_H_
+#define EXTRAS_DS1307_H_
+
+#include <stdint.h>
+#include <stdbool.h>
+#include <time.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * Squarewave frequency
+ */
+typedef enum _ds1307_squarewave_freq_t
+{
+    DS1307_1HZ = 0, //!< 1 Hz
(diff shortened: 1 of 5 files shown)


ourairquality:
There is already a ds3231 driver, and I understood these were largely compatible, so is a separate driver really necessary?

UncleRus:
Yes, they looks similar, but:
- registers map is same only at lower addresses (0x00 - 0x06)
- DS1307 has no alarms
- DS1307 has no temperature sensor
- Status and control registers are completely different
and so on.
I think specialized drivers for DS1307 and DS3231 is better.

sheinz:
Thanks @UncleRus for the driver!

This looks really good to me.

Even though generic driver that supports both ds1307 and ds3231 seems like a reasonable approach. Most likely in a real project only one kind of chip will be used and a simple clean driver that supports only the hardware you have seems like a good thing. Nevertheless I don't think theres only one right way. For now this driver looks good as it is.

UncleRus:
Please review

extras/ds1307/ds1307.h
```diff
@@ -0,0 +1,121 @@
...
+    DS1307_8192HZ,  //!< 8192 Hz
+    DS1307_32768HZ  //!< 32768 Hz
+} ds1307_squarewave_freq_t;
+
+/**
+ * \brief Init RTC
+ * \param scl_pin GPIO pin for I2C SCL
+ * \param sda_pin GPIO pin for I2C SDA
+ */
+void ds1307_init(uint8_t scl_pin, uint8_t sda_pin);
```
sheinz: I think it was @ourairquality who pointed out that initialising I2C driver inside another driver might be inconvenient when you have several drivers in the system that use the same I2C.

sheinz:
I've looked through the code and apart from a small comment about I2C initialisation everything looks good to me.

sheinz:
I've just checked it on my hardware. The example worked from the first try :)

@UncleRus Could you squash your commits for this PR, and I will merge it.

UncleRus:
Done

PR #235, opened 2016-10-11
Driver for ultrasonic range meters + example

UncleRus:
Driver for ultrasonic range meters like HC-SR04, HY-SRF05 and so on.

### DIFF
--- a/examples/ultrasonic/Makefile
+++ b/examples/ultrasonic/Makefile
@@ -1,0 +1,4 @@
+PROGRAM = ultrasonic
+EXTRA_COMPONENTS = extras/ultrasonic
+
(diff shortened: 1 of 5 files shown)


sheinz:
That is great!

I actually had in mind a little project involving ultrasonic distance measurement and esp8266 but hadn't have time to start working on it.
I will definitely check out this PR today evening.

Thanks @UncleRus!

sheinz:
I have tried the PR with the module I have and it works OK.

It would be interesting to try it under heavy WiFi traffic. I suspect the reading might be inconsistent due to WiFi interrupts. 

In my opinion the PR is good for merge.

PR #239, opened 2016-10-18
Driver for SD/MMC cards

UncleRus:
Low-level read/write/erase for MMC/SD/SDHC on SPI.
Good base for FatFs integration.


sheinz:
Very cool addition, thanks @UncleRus !

extras/sdio/sdio.h
```diff
@@ -0,0 +1,228 @@
...
+{
+    sdio_error_t     error;        //!< Last operation result
+    uint8_t          cs_pin;       //!< Chip Select GPIO pin
+    sdio_card_type_t type;         //!< Card type
+    bool             crc_enabled;  //!< True if CRC enabled for IO
+    sdio_ocr_t       ocr;          //!< OCR register
+    sdio_csd_t       csd;          //!< CSD register
+    sdio_cid_t       cid;          //!< CID register
+    uint32_t         sectors;      //!< Card size in 512 byte sectors
+} sdio_card_t;
```
sheinz: All those structures are for internal usage. They are not intended to be manipulated from outside as far as I understand.
@UncleRus  do you think it is a good idea to hide them in some kind of sdio_private.h (or sdio_impl.h) and include it here? I understand it will not change the name space but it will reduce visual noise in the main sdio.h header? So, the main sdio.h will contain only dirver's interface. What do you think?

UncleRus:
Example FatFs integration:
https://gist.github.com/UncleRus/e5ee407ec92b96df7964fd75b80df69f

sheinz:
Is this driver intended to be used with SD card modules like this?
sd card module
sd card module

resetnow:
I assume any module will to (although the first one has a voltage shifter which isn't required with esp8266)

UncleRus:
Yes, I'm using module just like this one and it works just fine.

sheinz:
I have laying around a bare microSD card holder without any components. So, as far as I understand I don't really need any additional circuitry to connect it to esp8266? Both esp8266 and sd card are 3.3V, so no voltage regulator and no level sifters. Is it correct?

resetnow:
You only need a bypass capacitor near the card and pullup resistors for SPI lines (SD card will start in SPI mode if these lines are pulled up).

sheinz:
Thanks @resetnow! I will try it out today evening.
Also thanks @UncleRus for the wiring diagram!

UncleRus:
@sheinz I'm ready to create a PR with FatFs integration after you merge this one :)
https://github.com/UncleRus/esp-open-rtos/tree/extras/fatfs/extras/fatfs

PR #246, opened 2016-10-22
Driver for HD44780 LCDs + examples

UncleRus:
Driver allows to choose GPIO expander to LCD pinout.
Example of most common module:
Most common module


extras/i2c_lcd/i2c_lcd.c
```diff
@@ -0,0 +1,151 @@
+#include "i2c_lcd.h"
+
+#include <pcf8574/pcf8574.h>
+#include <espressif/esp_common.h>
+
+#define MS 1000
+
+#define DELAY_CMD_LONG  (3 * MS) // >1.53ms accordint to datashit
```
resetnow: according to datasheet :)

extras/i2c_lcd/i2c_lcd.c
```diff
@@ -0,0 +1,151 @@
...
+#define DELAY_CMD_SHORT (60)     // >39us accordint to datashit
+#define DELAY_TOGGLE    (10)
```
resetnow: Every I2C write takes more than 10 us even when running on 400 kHz, wouldn't it work without this delay? Although if this driver is used to drive GPIO directly it would be needed. Maybe wrap delay block in `#ifdef` and provide different delays depending on whether a port expander or GPIO is used?

UncleRus:
Thanks @resetnow! Driver supports now both direct LCD connection to GPIO and I2C GPIO expanders.

Issue #248, opened 2016-10-22
conflicting '_read_r'


SaimenSays:
Hello,

when using components "extras/spiffs" and "extras/stdin_uart_interrupt", compiler brings up an error because the function `_read_r()` is implemented twice. 
I think these `_r` functions are for posix support of spiffs. But there should be a compiler flag to enable this, because it can cause also some other sideefects.

ourairquality:
A hack workaround would be to rename the definition in stdin_uart_interrupt and call this from the definition in spiffs if it is not a spiffs file descriptor, replacing the current uart rx handling code in spiffs.

Perhaps a more permanent fix would be to integrate stdin_uart_interrupt into the core, perhaps behind a build option, and then conditionalize the spiffs code on this option. The stdin_uart_interrupt code seems like a core function to me.

SaimenSays:
@ourairquality: I don't need a hack, because I'm not using this posix read functions on spiffs.

The switch between outputs should be in core, but the functionalities configurable/overwritable...

sheinz:
Hi @SaimenSays,

Thanks for bringing this issue. I've created a [fix](https://github.com/SuperHouse/esp-open-rtos/pull/249). You can try it.

SaimenSays:
@sheinz: Thanks for this hint. I tested it right now with files from #249, but it does not work as expeccted! My call on `read(0, (void*)&byte, 1)` blocks all lower priority threads. I did no investigation, but I think something might be wrong.

Rethinking about the blocking call, I seriously want to have the interrupt driven uart as core functionality. The Idle task gets not executed anymore when calling the blocking read, so RTOS does not work properly anymore.

sheinz:
Hi @SaimenSays, can you try the updated https://github.com/SuperHouse/esp-open-rtos/pull/249
It should work as expected now.

SaimenSays:
Thanks!

PR #249, opened 2016-10-23
Fix spiff and stdin_uart_interrupt overiding the same read function

sheinz:
Fix for https://github.com/SuperHouse/esp-open-rtos/issues/248

In this PR function `_read_r` makes additional call to `_read_stdin_r`.
So, **stdin_uart_interrupt** can override only part that it is interested in.
Also **spiffs** overrides `_read_r`, handles file descriptors and leave stdin to `_read_stdin_r` whoever overrides it.

### DIFF
--- a/common.mk
+++ b/common.mk
@@ -134,9 +134,15 @@ $$($(1)_OBJ_DIR)%.o: $$($(1)_REAL_ROOT)%.S $$($(1)_MAKEFILE) $(wildcard $(ROOT)*
 	$$($(1)_CC_BASE) -c $$< -o $$@
 	$$($(1)_CC_BASE) -MM -MT $$@ -MF $$(@:.o=.d) $$<
 
-# the component is shown to depend on both obj and source files so we get a meaningful error message
-# for missing explicitly named source files
-$$($(1)_AR): $$($(1)_OBJ_FILES) $$($(1)_SRC_FILES)
+$(1)_AR_IN_FILES = $$($(1)_OBJ_FILES)
+
+# the component is shown to depend on both obj and source files so we get 
+# a meaningful error message for missing explicitly named source files
+ifeq ($(INCLUDE_SRC_IN_AR),1)
+   $(1)_AR_IN_FILES += $$($(1)_SRC_FILES)
+endif
+
+$$($(1)_AR): $$($(1)_AR_IN_FILES)
 	$(vecho) "AR $$@"
 	$(Q) mkdir -p $$(dir $$@)
 	$(Q) $(AR) cru $$@ $$^
--- a/core/newlib_syscalls.c
+++ b/core/newlib_syscalls.c
@@ -59,14 +59,9 @@ __attribute__((weak)) long _write_r(struct _reent *r, int fd, const char *ptr, i
 }
 
 /* syscall implementation for stdio read from UART */
-__attribute__((weak)) long _read_r( struct _reent *r, int fd, char *ptr, int len )
+__attribute__((weak)) long _read_stdin_r(struct _reent *r, int fd, char *ptr, int len)
 {
     int ch, i;
-
-    if(fd != r->_stdin->_file) {
-        r->_errno = EBADF;
-        return -1;
-    }
     uart_rxfifo_wait(0, 1);
     for(i = 0; i < len; i++) {
         ch = uart_getc_nowait(0);
@@ -74,6 +69,15 @@ __attribute__((weak)) long _read_r( struct _reent *r, int fd, char *ptr, int len
         ptr[i] = ch;
     }
     return i;
+}
+
+__attribute__((weak)) long _read_r( struct _reent *r, int fd, char *ptr, int len )
+{
+    if(fd != r->_stdin->_file) {
+        r->_errno = EBADF;
+        return -1;
+    }
+    return _read_stdin_r(r, fd, ptr, len);
 }
 
 /* Stub syscall implementations follow, to allow compiling newlib functions that
--- a/extras/spiffs/esp_spiffs.c
+++ b/extras/spiffs/esp_spiffs.c
@@ -141,21 +141,16 @@ long _write_r(struct _reent *r, int fd, const char *ptr, int len )
     return len;
 }
 
+// This function is weakly defined in core/newlib_syscalls.c
+long _read_stdin_r(struct _reent *r, int fd, char *ptr, int len);
+
 // This implementation replaces implementation in core/newlib_syscals.c
 long _read_r( struct _reent *r, int fd, char *ptr, int len )
 {
-    int ch, i;
-
     if(fd != r->_stdin->_file) {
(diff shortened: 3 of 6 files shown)


SaimenSays:
Tried this code, but it seems as new function `_read_stdin_r()` from stdin_uart_interrupt.c does not overwrite function defined in newlib_syscalls.c. In my case I get a blocking read instead of an interrupt driven.

sheinz:
@SaimenSays is right. A new function `_read_stdin_r` is not called.
I have not idea why :(

sheinz:
The only way to make a strong function definition replace a weak one is to link the whole library with a strong function definition.

If anybody knows a better way, please tell.

I updated the PR with changes that link **stdin_uart_interrupt** with `--whole-archive` option.

common.mk
```diff
@@ -134,9 +134,15 @@ $$($(1)_OBJ_DIR)%.o: $$($(1)_REAL_ROOT)%.S $$($(1)_MAKEFILE) $(wildcard $(ROOT)*
...
 
-# the component is shown to depend on both obj and source files so we get a meaningful error message
-# for missing explicitly named source files
-$$($(1)_AR): $$($(1)_OBJ_FILES) $$($(1)_SRC_FILES)
+$(1)_AR_IN_FILES = $$($(1)_OBJ_FILES)
+
+# the component is shown to depend on both obj and source files so we get 
+# a meaningful error message for missing explicitly named source files
+ifeq ($(INCLUDE_SRC_IN_AR),1)
+   $(1)_AR_IN_FILES += $$($(1)_SRC_FILES)
```
sheinz: If source files is added into an archive it can not be included into a binary with option --whole-archive

resetnow:
Is it possible to check whether any of extra components is enabled in core and change core _read_r implementation accordingly?

sheinz:
> Is it possible to check whether any of extra components is enabled in core and change core _read_r implementation accordingly?

I think so. Some definition flag might be added and checked in core. If the flag is defined _read_r would be skipped in core. This definition flag might be declared in extras/component/component.mk
It is possible but the weak symbol feature of gcc is designed to do exactly the same during linkage stage without any additional flags.

resetnow:
(I replied via email and it didn't go through for some reason, if it did but was deleted let me know)

>the weak symbol feature of gcc is designed to do exactly the same during linkage stage without any additional flags.

Maybe the order of linked libraries makes difference then? I did essentially the same thing for esp-gdbstub and it worked. As far as I understand, when linking static libraries, ld stops looking for a symbol after a first occurrence unless a flag is specified. Maybe moving extra libraries to the start of -l list would help?

sheinz:
@resetnow I also thought that the order might count but my tests didn't prove that.
Here's a small example that shows when a weak function is replaced by a strong one:
*main.c*:
```
#include <stdio.h>

__attribute__((weak)) void some_func()
{
    printf("Weak func\n");
}

int main(int argc, char *argv[])
{
    printf("Main function\n");

    some_func();

    return 0;
}
```
*test.c*:
```
#include <stdio.h>

void some_func()
{
    printf("Strong func\n");
}
```
*Makefile*:
```
all: test

test.o: test.c

main.o: main.c

# case 1: weak function is not replaced
test: main.a test.a 
	gcc $^ -o $@

# case 2: weak function is not replaced
#test: main.a test.a 
#	gcc test.a main.a -o $@

## case 3: weak function is not replaced
#test: main.a test.a 
#	gcc -Wl,--start-group $^ -Wl,--end-group -o $@

## case 4: weak function is replaced
#test: main.a test.a 
#	gcc -Wl,--whole-archive test.a -Wl,--no-whole-archive main.a -o $@

%.a: %.o 
	ar rcs $@ $^

clean:
	-rm -f test.o test test.a main.a main.o
```
So, a weak function is only replaced if the library with the strong definition is linked as a whole archive.

resetnow:
I think I figured out why it worked for me: I replaced <a href="https://github.com/SuperHouse/esp-open-rtos/blob/8368929a6658082ebe5eee6926232986e7895f60/core/include/debug_dumps.h#L24">a weak symbol with an alias</a>. So a possible solution is to define _read_r in core as a weak symbol with alias to _read_r_core which could be overridden when needed.

I still don't see how it would help with multiple symbols, it would be better to have _r_read in core to call other implementations depending on the project configuration.

sheinz:
Still can't make the example above to replace a weak symbol. Even if I declare a weak alias to a function:
```
__attribute__((weak, alias("main_some_func"))) void some_func();

void main_some_func()
{
    printf("Weak func\n");
}
```
I don't think that there's a way to make it work from the source code. Because the static libraries are generated correctly:
```
> nm main.a

main.o:
00000000 r .LC0
0000000c r .LC1
00000014 T main
00000000 T main_some_func
         U puts
00000000 W some_func
```
```
> nm test.a

test.o:
00000000 r .LC0
         U puts
00000000 T some_func
```
Please note the **W** in *main.a*, which indicates that a symbol is weak and should be replaced with **T** from *test.o*.

resetnow:
Well the flag is needed so LD would not stop at the first encountered
symbol with a matching name. It's probably supposed to provide better
performance (?).

However I did the same thing (weak symbol) in esp-gdbstub and it worked
somehow. Maybe -l flags order matters?

24 окт. 2016 г. 16:48 пользователь "sheinz" <notifications@github.com>
написал:

projectgus:
Something that might explain some of this behaviour, referencing another symbol in the same object file as the strong replacement symbol will cause the linker to examine the whole object file, and pull in the replacement symbol. (May explain the esp-gdbstub case, I don't know.)

For my 2c, I think `--whole-archive` is what is wanted here. Combined with `--gc-sections` I don't think it will create larger binaries, just a slower link pass.

sheinz:
@projectgus is absolutely right!

Just adding another regular function and referencing it from the main causes the linker to examine the archive for weak symbols replacement!
*main.c*:
```
__attribute__((weak)) void some_func()
{
    printf("Weak func\n");
}
int main(int argc, char *argv[])
{
    some_func();
    other_func();
    return 0;
}
```
*test.c*:
```
void some_func()
{
    printf("Strong func\n");
}
void other_func()
{
}
```
So, if the linker doesn't find any references to the library it skips it entirely and do not consider any symbols from the library as week symbols replacement.

ourairquality:
Saw the same when trying to start an open libnet80211but recall that it might have been per-object file within a library, not just per library, so a reference to a function in each object file was required. Would be great to get a workaround for this in the linking of the open libraries too.

resetnow:
I think this issue may become more relevant over time if more functions will be implemented in source. So adding  `--whole-archive` by default makes sense.

phlemoine:
I have been facing a similar issue when using spiffs and lwip with the the
read and write functions. I solved this by a correct order of #include and
defining LWIP_POSIX_SOCKETS_IO_NAMES to 0

PR #250, opened 2016-10-24
Make headers in extras more cpp friendly

UncleRus:
Basically "extern "C" was added to some of the extras headers.

### DIFF
--- a/extras/i2c/i2c.h
+++ b/extras/i2c/i2c.h
@@ -24,10 +24,13 @@
 
 #ifndef __I2C_H__
 #define __I2C_H__
-#endif
 
 #include <stdint.h>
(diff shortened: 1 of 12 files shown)


extras/i2c/i2c.h
```diff
@@ -49,3 +52,9 @@ bool i2c_slave_read(uint8_t slave_addr, uint8_t data, uint8_t *buf, uint32_t len
 // devices where the i2c_slave_[read|write] functions above are of no use.
 void i2c_start(void);
 void i2c_stop(void);
+
+#ifdef	__cplusplus
+}
+#endif
+
+#endif /* __I2C_H__ */
```
sheinz: Nice catch :)

PR #254, opened 2016-10-26
Driver and example for SSD1306 128x64 I2C display

urx:
Hi there. I wrote simple library for deal with SSD1306 128x64 display connected via I2C. It allows you to configure SSD1306 display and load an image in XBM format.


UncleRus:
Very cool addition! Thank you, @urx!

examples/ssd1306_i2c/ssd1306_i2c.c
```diff
@@ -0,0 +1,70 @@
+/* vim: set expandtab ts=4 sw=4 :
```
sheinz: Hello fellow vim user :)
I don't think we should include configuration of our favourite editors in esp-open-rtos, as not everybody uses them and it might confuse others. It just my personal opinion.
I think it's better to include more descriptive information in the header comment or/and licensing information.

extras/ssd1306/ssd1306.h
```diff
@@ -0,0 +1,93 @@
...
+#define SSD1306_ALT_COM_PINS_CFG    (0x12)
+#define SSD1306_COM_LR_REMAP_OFF    (0x02)
+#define SSD1306_COM_LR_REMAP_ON     (0x22)
+
+#define SSD1306_SET_DESEL_LVL       (0xDB)
+#define SSD1306_SET_NOP             (0xE3)
+
+#define SSD1306_SET_CHARGE_PUMP     (0x8D)
+#define SSD1306_CHARGE_PUMP_EN      (0x14)
+#define SSD1306_CHARGE_PUMP_DIS     (0x10)
```
sheinz: Does user need those definition to use the library? Can they be moved into .c file to make header shorter? So the header  would contain only library interface information.

sheinz:
Thanks @urx! Apart from a few cosmetic comments this PR looks really good to me.
I was just wandering what is the picture in the file? :)

extras/ssd1306/ssd1306.h
urx: May be I move this defines into some separate header like ssd1306_priv.h?

urx:
@sheinz pic in the file not so interesting :) [you can view it here](http://savepic.ru/11989725.jpg)

examples/ssd1306_i2c/ssd1306_i2c.c
urx: May be you are right, but i thought this vim configuration may be useful for peoples like me who working on several projects with different indenting.

extras/ssd1306/ssd1306.h
urx: ok. you win :))

Issue #210, opened 2016-08-29
spiffs build artifacts


phlemoine:
i am using spiffs and it seems that these files are created at build time

        extras/spiffs/mkspiffs/mkspiffs
        extras/spiffs/mkspiffs/mkspiffs.o
        extras/spiffs/mkspiffs/spiffs_cache.o
        extras/spiffs/mkspiffs/spiffs_check.o
        extras/spiffs/mkspiffs/spiffs_gc.o
        extras/spiffs/mkspiffs/spiffs_hydrogen.o
        extras/spiffs/mkspiffs/spiffs_nucleus.o

i wonder if this is expected or not. in any case, the .o should be excluded from .gitignore.
having said that this won't solve the issue caused by the 1st file in this list. any guidance on how to handle this properly ?

sheinz:
Hi @phlemoine,

Yes. mkspiffs is currently built in the source directory. I was thinking about placing object files in the application build directory but didn't implement that at the time.
I will try to move mkspiffs objects and executable into the application's build directory.

PR #256, opened 2016-10-27
Add mkspiffs binary to .gitignore

UncleRus:
mkspiffs binaries added to .gitignore.


sheinz:
Arghh, I was planing to move the build artifacts of mkspiffs into the application's build directory but hadn't have time for that so far. There's an issue about it https://github.com/SuperHouse/esp-open-rtos/issues/210

But adding those in .gitignore wouldn't hurt.

PR #257, opened 2016-10-28
dhcpserver.c corrections.

DanielCerejo:
initialize dhcpserver_task_handle = NULL
corrert error printf "OTA TFTP" to "DHCP Server Error"


sheinz:
Thanks @DanielCerejo  for the fix!

PR #258, opened 2016-10-31
DS1302 RTC driver + example

UncleRus:
Driver for yet another widely spread RTC


sheinz:
Thanks @UncleRus, looks good to me.

PR #245, opened 2016-10-21
FreeRTOS v9.0.0 upgrade

ourairquality:
Let me try again, sorry for all the github noise.

PR #261, opened 2016-11-02
FreeRTOS type updates.

ourairquality:


sheinz:
Thanks @ourairquality. 
As far as I can see only types definitions are changed. Looks good to me.
This is a nice addition to https://github.com/SuperHouse/esp-open-rtos/pull/245

sheinz:
I checked the PR with the tests and everything seems good. I will merge it if there's no objections.
It's better to merge it quicker so the following PRs are based on the new types.

Issue #133, opened 2016-04-26
HTTP client library


projectgus:
This is related to #132, but I thought it'd warrant its own issue to discuss/track.

For a long time I've wanted to include an HTTP client library in esp-open-rtos. Making raw socket requests works for simple actions (apart from obvious problems like #132), but it'd be good to have some wrappers for common HTTP operations.

* I tried porting a stripped down libcurl, but even stripped down it's not suitable for this size of processor
* mbed has its own C++ HTTPClient library (although I'm not sure where the canonical source for it is), but it's in C++ and fairly mbed-centric so it may not be suitable.

Most people seem to just DIY on raw sockets for the subset of HTTP they need. We could write our own I guess, it might be of wider interest if it only depends on LWIP.

foogod:
Heh, that's funny.. I was just thinking of creating a ticket for this as well.  I agree that this would be a very useful general-purpose module to have in esp-open-rtos (and I also agree that an embedded-friendly C-only HTTP client library could be useful in general)..

This might not be a bad project for somebody new to esp-open-rtos, as it shouldn't require much beyond learning the ins and outs of the LWIP stuff (and familiarizing onesself with the HTTP protocol, of course).  I'm sorta envisioning this as something that could potentially be built in several stages:

1) Really minimal, but still useful:
* Supports GET requests
* Supports basic HTTP/1.0 protocol
* Just call the API routine and get back the payload (HTML page or whatever).

2) Basic two-way capabilities:
* Supports POST requests
* Support for sending/receiving custom HTTP headers
* Possibly more of a callback-based interface which could then be wrapped by simpler APIs

3) Reasonably full-featured basic HTTP client
* Reporting/custom handling of HTTP status codes
* Auto-handling HTTP redirects, etc
* HTTP basic auth support
* Other less common request types (HEAD, PUT, ...), get-if-modified, etc.
* Maybe extra routines to help with URL encoding, query strings, etc.
* Basic HTTPS support

4) All kinds of more advanced features (most probably all optional add-ons, if they ever are implemented):
* Cookie handling
* Support HTTP/1.1 and chunked encoding
* Streaming data to/from server
* Compressed payloads
* Establishing websockets
* etc

jsuiker:
[...]

PR #263, opened 2016-11-05
HTTP parser intergation

UncleRus:
> This is a parser for HTTP messages written in C. It parses both requests and responses. The parser is designed to be used in performance HTTP applications. It does not make any syscalls nor allocations, it does not buffer data, it can be interrupted at anytime. Depending on your architecture, it only requires about 40 bytes of data per message stream (in a web server that is per connection).

> Features:

> - No dependencies
> - Handles persistent streams (keep-alive).
> - Decodes chunked encoding.
> - Upgrade support
> - Defends against buffer overflow attacks.

Good base for HTTP client and server.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -16,3 +16,6 @@
 [submodule "examples/posix_fs/fs-test"]
 	path = examples/posix_fs/fs-test
 	url = https://github.com/sheinz/fs-test
+[submodule "extras/http-parser/http-parser"]
+	path = extras/http-parser/http-parser
+	url = https://github.com/nodejs/http-parser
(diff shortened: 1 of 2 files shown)


sheinz:
> Good base for HTTP client and server.

I agree.  

@UncleRus Do you plan to continue work in this direction? I mean to bring some http server/client functionality into esp-open-rtos that is based on this parser? It was actually discussed here https://github.com/SuperHouse/esp-open-rtos/issues/133
If it is just this PR without following higher lever library it would be nice to have an example for it as we usually have for each driver/library.

UncleRus:
@sheinz Yes, currently I'm working on http-server that allows flexible requests/handlers routing, like other web frameworks does.
After it I plan to start work on http-client library based on lwip sockets and http-parser.

PR #265, opened 2016-11-05
WiFi scan example

UncleRus:
Just a simple example.


sheinz:
Thanks @UncleRus.
That functionality we have been missing in the examples. Will come in handy for new users.

PR #267, opened 2016-11-11
Update and re-organise the sdk internal definitions.

ourairquality:
This update adds some new sdk internal structures and updates the existing ones. The global definitions have been moved into the include/esplibs/ directory and sorted per library and per source object file. These are needed to support sdk code to be submitted and if this could be landed first then it would help split them up and might avoid some merge ordering problems. If this were to move a slot then it might break the code, but it's a small part of what it is to be submitted and some extra checks have been added to the structure layouts to try to detect edit problems.  Checks have also been added for the lwip netif slots offsets known to be used by the sdk binary code which should help other people using esp-open-rtos as a lwip development platform.

### DIFF
--- a/FreeRTOS/Source/portable/esp8266/portmacro.h
+++ b/FreeRTOS/Source/portable/esp8266/portmacro.h
@@ -144,7 +144,7 @@ void PendSV(enum SVC_ReqType);
    ESPTODO: It may be possible to just read the 'ps' register instead
    of accessing thisvariable.
 */
-extern char sdk_NMIIrqIsOn;
+extern uint8_t sdk_NMIIrqIsOn;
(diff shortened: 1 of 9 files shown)


sheinz:
The work with reverse-engineering SDK internals is tedious and prone to introduce regression.
When I did work on reverse engineering and rewriting SPI flash access the automatic tests helped me a lot. That is why I started work on brining some automatic testing to EOR. So, the basic functionality can be quickly checked with rewritten (open sourced) libraries.
On my side, I will continue to improve testing and help with opening binary libraries.

sheinz:
From the first glance looks good to me.
I will merge it if there's not objections.

PR #268, opened 2016-11-11
Use a separate file for the private SSID config.

ourairquality:
This patch moves the private SSID config to a file not included in the git repo. The current setup is a real foot gun for developer. Non-developers can almost as easily create a separate file as editing the ssid_config.h file. Alternatively a template file could be included. When developers are dealing with many git repo copies and trying to test things, it's just too much of a burden to have to manually config git for the ssid_config.h file, and we need to commit to the master branch and test that so can not deal with this in a local branch. As a developer, don't expect more contributions from me here if this is not fixed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,7 @@ before_install:
 script:
   - cd ${TRAVIS_BUILD_DIR}
   # Remove ssid_config requirement for examples
-  - sed -i "s%#error%//#error%" include/ssid_config.h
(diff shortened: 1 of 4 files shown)


projectgus:
A suggestion to go one further, add private_ssid_config.h to .gitignore so you have to "git add -f" to add it to the index.

sheinz:
This is a good addition. It can be merged as it is.

I agree with @ourairquality that it is burden to configure each repository clone to test some features.
It would be even better to be able to just clone a repo, build an example and flash it straight away without having to modify or create any additional files.

PR #271, opened 2016-11-12
[WIP] Travis running tests

sheinz:
This PR adds additional step in Travis CI. This step is running tests on a real hardware connected to a test server(currently only one).

This will allow to check basic functionality with each PR.

My test server is running 24/7 at my place but in any case if a test server is down the Travis build will not fail. The travis build will only fail if the tests fail.


sheinz:
For some reason Travis can't decrypt the key. With my fork everything works fine.
I tried to encrypt it specially for SuperHose/esp-open-rtos Travis repo but without luck.
Need to investigate it more...

resetnow:
I wonder how many write cycles will the flash chip survive until errors start to appear.

sheinz:
> I wonder how many write cycles will the flash chip survive until errors start to appear. Branded ones like Micron have a life of about 100k cycles, but I think most esp have noname chips.

Yes, I also though about it when testing spiffs. At that time I had several modules constantly running spiffs tests on them. Those tests were constantly overwriting flash. So far I haven't had a single faulty module.
Nevertheless, the modules are cheap, so for testing purposes if one fails we can easily replace it.

By the way, the current testing approach, which is based on @projectgus original work is less cruel to flash, because it only flashes one firmware for all test cases. My first implementation had a separate firmware for each test case.

sheinz:
It appears that I can't provide encrypted key for Travis via a PR from a fork:
>Please note that encrypted files are not available for pull requests from forks.

I will create another PR from the branch on the original repository.

PR #272, opened 2016-11-13
Travis running tests

sheinz:
>This PR adds additional step in Travis CI. This step is running tests on a real hardware connected to a test server(currently only one).

>This will allow to check basic functionality with each PR.

>My test server is running 24/7 at my place but in any case if a test server is down the Travis build will not fail. The travis build will only fail if the tests fail.

Adding encrypted keys for Travis can be done only from the original repository.
This PR is essentially https://github.com/SuperHouse/esp-open-rtos/pull/271


sheinz:
Any comments, objections on this one?

ourairquality:
Looks like a good step forward, really good contribution, thanks.

PR #274, opened 2016-11-15
Flush the uart tx fifo before a restart.

ourairquality:
Otherwise a restart ends up in a wdt halt. This change improves things.

Also comment the point at which the bus clock that drives the uart changes on startup and comment out the change in the uart divisor. This at least allows a consistent uart baud rate during a restart if using the rate 115200. When restarting the bus clock does not change and the rom resets the rate to 115200 so if using a rate of 115200 the rate is consistent and the startup messages can be see without corruption.

### DIFF
--- a/core/app_main.c
+++ b/core/app_main.c
@@ -254,12 +254,25 @@ static void zero_bss(void) {
 
 // .Lfunc006 -- .irom0.text+0x70
 static void init_networking(sdk_phy_info_t *phy_info, uint8_t *mac_addr) {
+    // The call to sdk_register_chipv6_phy appears to change the bus clock,
+    // perhaps from 40MHz to 26MHz, at least it has such an effect on the uart
+    // baud rate. The caller flushes the TX fifos.
     if (sdk_register_chipv6_phy(phy_info)) {
         printf("FATAL: sdk_register_chipv6_phy failed");
         abort();
     }
-    uart_set_baud(0, 74906);
-    uart_set_baud(1, 74906);
+
+    // The boot rom initializes uart0 for a 115200 baud rate but the bus clock
+    // does not appear to be as expected so the initial baud rate is actually
+    // 74906. On a cold boot, to keep the 74906 baud rate the uart0 divisor
+    // would need to changed here to 74906. On a warm boot the bus clock is
+    // expected to have already been set so the boot baud rate is 115200.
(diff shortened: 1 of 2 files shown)


sheinz:
I'm not sure if this is related. If I programatically reset a device after the flashing it stucks after wdt reset:
```
ets Jan  8 2013,rst cause:2, boot mode:(1,6)


ets Jan  8 2013,rst cause:4, boot mode:(1,6)

wdt reset
```
After the first hardware reset the device starts to reset properly. I reset the device by calling `sdk_system_restart();`

projectgus:
> I'm not sure if this is related. If I programatically reset a device after the flashing it stucks after wdt reset:

The reason for this (I believe) is that the GPIO strapping pin register remains set for the UART bootloader, so it resets back to the bootloader. This can be useful in some cases (for example, if your ESP8266 doesn't have an auto-reset circuit and you want to be able to re-program it any time it crashes without toggling GPIO0), but it's also kind of annoying. One of the features I want to do add in esptool.py v2.0 is a programmable "reset after' mode.

ourairquality:
@projectgus Thank you I see the issue after flashing not now - the strapping pins were captured when it cold booted and it does not appear to be possible to clear them. So the workaround for this was to change esptool to hard reset when done.

```
def write_flash(esp, args):
   ...
    print 'Reset into firmware...'
    time.sleep(0.05)
    esp._port.setDTR(False)
    esp._port.setRTS(True)
    time.sleep(0.05)
    esp._port.setRTS(False)
```

ourairquality:
@sheinz Thank you, it does seem to be a post-flash issue, and I am having trouble reproducing the issue taking more care on that point! There may be no compelling reason to make these changes now, but the flushing might still be prudent and has been updated, and the other changes at least give a clear warm boot log when running at 115200 baud.

projectgus:
FWIW, the high 16 bits of the GPIO.IN register contain strapping information:
https://github.com/SuperHouse/esp-open-rtos/blob/master/core/include/esp/gpio_regs.h#L44

and I believe you can write to them to set the bit corresponding to GPIO0, and therefore change the reset mode. I haven't tested this though.

The workaround you have (hard reset) is a good one and will be added to esptool.py (with an option to disable) in v2.0.

projectgus:
(I agree that flushing UART FIFOs before reset is still good practice, though.)

PR #280, opened 2016-11-15
open_esplibs libnet80211: source code for sdk_cnx_node_search.

ourairquality:
Also code for sdk_cnx_sta_leave, although disabled due to accessing static date, but it shows a reference to the netif->flags and the NETIF_FLAG_DHCP flag removed in lwip v2.

### DIFF
--- a/open_esplibs/libnet80211/wl_cnx.c
+++ b/open_esplibs/libnet80211/wl_cnx.c
@@ -7,5 +7,100 @@
 #if OPEN_LIBNET80211_WL_CNX
 // The contents of this file are only built if OPEN_LIBNET80211_WL_CNX is set to true
 
+#include "espressif/esp_misc.h"
+#include "esplibs/libnet80211.h"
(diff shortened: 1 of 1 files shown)


open_esplibs/libnet80211/wl_cnx.c
```diff
@@ -7,5 +7,100 @@
...
+    int end = sdk_g_ic.s._unknown310 + 2;
+
+    // Note this defensive test seems dead code, the value is loaded
+    // as a uint8_t value so adding 2 ensures this test never passes.
+    if (end < 1)
+        return NULL;
+
+    struct sdk_netif_conninfo **conninfo = sdk_g_ic.v.softap_netif_info->conninfo;
+
+    int i = 0;
```
sheinz: They might have written it with a **for** loop but the compiler optimised away the first check as `end` can't be zero. I don't see the reason why not to write such code with a **for** loop.

PR #283, opened 2016-11-15
open_esplibs libwpa: source code for os_xtensa.c

ourairquality:
Interesting support functions that we might want or need to re-implement. Hope WPA does not depend on the time returned by os_get_time as it does not fill the value at all.


resetnow:
Out of curiosity, does original binary use `rand()` as well? Shouldn't it make more sense to use hardware RNG?

ourairquality:
@resetnow Yes, this is just a straight translation of the binary code, faults and all, just a start.

PR #284, opened 2016-11-15
Add BearSSL extra

stschake:
Hey,

this adds the BearSSL library (https://bearssl.org/) to extras/ and adds a client/server TLS example similar to the existing ones for mbedTLS.

The main benefits in this library over mbedTLS are 1) a complete lack of dynamic memory allocation, so there are no free heap surprises as you might encounter with mbedTLS, 2) secure by default choices, the library doesn't support insecure communication setups (such as SSL 3.0) and 3) a significantly smaller flash footprint than mbedTLS. As BSD it's also more liberally licensed.

However, it is still an experimental library (version 0.1). This pull request doesn't modify upstream in any way, so it should be very straightforward to update.

Regards,
Stefan


sheinz:
Hi @stschake,

Thanks for the contribution!

It's always nice to have options to choose from.

UncleRus:
Closed in favor #340

PR #286, opened 2016-11-17
open_esplibs libnet80211: source code for sdk_wifi_softap_start paths.

ourairquality:
The function sdk_wifi_softap_start is one of two paths that allocates a struct netif and also accesses the netif->hwaddr so is needed for lwip development. This function depends on static data and functions so a good bit so other code was needed.

This was a difficult set, challenging to review, but it gets us the allocation of the netif structure needed for LwIP development, plus the remaining uses of the ets_timer, so feedback welcomed.

### DIFF
--- a/open_esplibs/libnet80211/ieee80211_hostap.c
+++ b/open_esplibs/libnet80211/ieee80211_hostap.c
@@ -7,4 +7,312 @@
 #if OPEN_LIBNET80211_HOSTAP
 // The contents of this file are only built if OPEN_LIBNET80211_HOSTAP is set to true
 
+#include <string.h>
+#include "tcpip.h"
+#include "espressif/esp_wifi.h"
+#include "espressif/esp_misc.h"
+#include "etstimer.h"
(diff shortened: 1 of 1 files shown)


resetnow:
By the way, some of decompiled net80211 functions can be found at https://github.com/ernacktob/esp8266_wifi_raw

ourairquality:
@sheinz Testing is limited, it appears to work in all the code I am working with, but some of these functions were a struggle to translate by hand so I would like to re-check them. Very hard to review these. I am exploring some automation to help translate the code, it still needs a lot of manual work, but might give a different view for comparison. Still need to get the wl_cnx.o translated for lwip v2 to be practical, which could take time, so there is no rush on this PR.

ourairquality:
Have been over this many times so hope it is all ok, and it was necessary to bundle it in with the lwip v2 support https://github.com/SuperHouse/esp-open-rtos/pull/389 so closing this PR.

PR #292, opened 2016-11-22
ADS1113/ADS1114/ADS1115 I2C ADC driver + example

UncleRus:
> The ADS1113, ADS1114, and ADS1115 are precision analog-to-digital converters (ADCs) with 16 bits of resolution offered in an ultra-small, leadless QFN-10 package or an MSOP-10 package.

Unfortunately I overvoltaged my ADS1115 during tests, so I hope someone can help me with testing.

### DIFF
--- a/examples/ads1115_test/Makefile
+++ b/examples/ads1115_test/Makefile
@@ -1,0 +1,4 @@
+PROGRAM = ads1115_test
+EXTRA_COMPONENTS = extras/i2c extras/ads111x
+#ESPBAUD = 460800
+include ../../common.mk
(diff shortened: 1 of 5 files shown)


sheinz:
I think we can merge it even if it is untested.
It will not break anything as it is a new driver. If it is in the master branch it has more chances to be used. And if it doesn't work properly somebody might open an issue.

I've ordered this module but it will take a month till I receive it and can test it.

ourairquality:
I agree with being less demanding on code reviews for code in the extras. Writing a driver for the ADS1115 was on my TODO list, so it's great to see this, and I'll also give it some testing. The other one I have on hand is the MCP3424 which I will write a driver for in time unless someone beats me to it. Both look promising for hooking up to electro-chem air quality sensors.

PR #296, opened 2016-11-26
Spi features

Zaltora:
Hello, i wrote some new functions to spi. Inspired from multiple spi drivers i had like this one : https://github.com/MetalPhreak/ESP8266_SPI_Driver.

New funtions:
```
spi_set_command(uint8_t bus,uint8_t bits, uint16_t data);
spi_set_address(uint8_t bus,uint8_t bits, uint32_t data);
spi_set_dummy_bits(uint8_t bus, uint8_t bits, bool pos);
spi_clear_address(uint8_t bus);
spi_clear_command(uint8_t bus);
spi_clear_dummy(uint8_t bus);
void spi_repeat_send_8(uint8_t bus, uint8_t data, int32_t repeats);
void spi_repeat_send_16(uint8_t bus, uint16_t data, int32_t repeats);
void spi_repeat_send_32(uint8_t bus, uint32_t data, int32_t repeats);
```
I tested these functions on the oscilloscope. all work fine but you can do some test to check. 
I had some strange behavior with memcpy so i used this code,
`SPI(bus).W[i] = *dword; ` 
instead of,
`memcpy((void *)&SPI(bus).W[i], dword, SPI_32BIT);`
Maybe need test again ?

It will allow SPI3 protocol for SSD1306 driver and more.

### DIFF
--- a/core/esp_spi.c
+++ b/core/esp_spi.c
@@ -248,33 +248,33 @@ size_t spi_transfer(uint8_t bus, const void *out_data, void *in_data, size_t len
 
 static void _repeat_send(uint8_t bus, uint32_t* dword,int32_t* repeats,spi_word_size_t size)
 {
-	uint8_t i = 0 ;
-	while(*repeats > 0)
-	{
-		uint16_t bytes_to_transfer = __min(*repeats * size , _SPI_BUF_SIZE);
-		_wait(bus);
-		_set_size(bus,bytes_to_transfer);
-		for(i = 0; i < (bytes_to_transfer + 3) / 4;i++)
-			SPI(bus).W[i] = *dword; //need test with memcpy !
-		_start(bus);
-		*repeats -= (bytes_to_transfer / size ) ;
-	}
-	_wait(bus);
+    uint8_t i = 0 ;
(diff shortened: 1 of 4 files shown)


sheinz:
Hi @Zaltora,

Thanks for the contribution.

> I had some strange behavior with memcpy so i used this code,
SPI(bus).W[i] = *dword;
instead of,
memcpy((void *)&SPI(bus).W[i], dword, SPI_32BIT);

The problem with `memcpy` might be that it discards `volatile` qualifier of `SPI(bus).W` which causes that `memw` instructions are not placed before accessing registers.

I noticed tabs in the code, [we don't use tabs in eor](https://github.com/SuperHouse/esp-open-rtos#coding-style).

Overall looks like a useful addition, but I haven't checked the code in details yet.

Zaltora:
Sry about tab, forget to change my auto-formater in eclipse. BSD style come back !

UncleRus:
Very useful functions! Repeating send is needed in most SPI protocols.
Going to merge it tomorrow if no objections.
@Zaltora Thank you, great contribution!

Issue #174, opened 2016-07-27
Swap in a faster memcpy


projectgus:
This is a bit "full circle", but:

* When we swapped in our own newlib for a libc we got this memcpy: https://github.com/projectgus/newlib-xtensa/blob/xtensa-2_2_0-lock/newlib/libc/machine/xtensa/memcpy.S - it does byte reads until a 4 byte aligned address, then word reads, and I think it does byte reads for any trailing bytes as well.

* With #142 we enabled the unaligned load handler for default constant data which was now in flash. So this memcpy got slower in more cases.

* In #163 @sheinz noticed substantial performance drop using the newlib memcpy for spiffs, and he's written a new one. He also points out that the one in Espressif's ROM is pretty fast.

So maybe we should swap out memcpy (and maybe some of the other very simple low level libc functions) to ones which never perform byte reads?

Could do this in one of several ways:
* Use the Espressif ROM binary-only implementations, swap them in at link time.
* Write our own and put them in our codebase, like @sheinz has done with his memcpy implementation. Swap them in at link time.
* Patch newlib-xtensa, given we already have some e-o-r specific modifications in there.

... I don't really mind which one. Leaving this issue open to track/discuss the desire for one of them, though!

sheinz:
> With #142 we enabled the unaligned load handler for default constant data which was now in flash. So this memcpy got slower in more cases.

Are you sure? Have you done any benchmarking? My preliminary testing shows no substantial performance difference between those functions when operating with RAM data buffers:
0x4000df48	memcpy
0x400018b4	ets_memcpy
0x400006c4	xthal_memcpy
newlib's memcpy

Am I correct that unaligned exception handler works only for unaligned flash(ROM) access?

>In #163 @sheinz noticed substantial performance drop using the newlib memcpy for spiffs, and he's written a new one. He also points out that the one in Espressif's ROM is pretty fast.

I couldn't use standard **memcpy** because it discards **volatile** qualifier for registers.
You might misunderstood me. I noticed performance drop between manually copying  data from registers into unaligned buffer with pure C. So, I had to use assembler to do it efficiently.

Doesn't e-o-r currently use [**memcpy** from rom](https://github.com/SuperHouse/esp-open-rtos/blob/master/ld/rom.ld#L62)?

sheinz:
I've done some benchmarking.
The number in the following log is the time in microseconds required to copy 128 bytes 1000 times.

RAM to RAM copy:

    newlib memcpy:
[...]

PR #299, opened 2016-11-28
sysparam fixes, tests, spi flash refactoring

sheinz:
This PR includes:
 * Sysparam fixes for thread safety by @ourairquality 
 * Sysparam unit tests
 * Fix for negative int8 values
 * Sysparam bools without heap memory allocation
 * SPI flash refactoring, so spiffs and sysparam use the same flash access
 * spiflash test for unaligned access
 * switch from hand-written memcpy to the newlib's as discussed in https://github.com/SuperHouse/esp-open-rtos/issues/174 for spiflash

This PR allows https://github.com/SuperHouse/esp-open-rtos/pull/208 to be merged. I tested it and it works.


core/sysparam.c
```diff
@@ -33,11 +31,14 @@
...
 #define SCAN_BUFFER_SIZE 8 // words
 
-/* The size of the temporary buffer used for reading back and verifying data
- * written to flash.  Making this larger will make the write-and-verify
- * operation slightly faster, but will use more heap during writes
+/* The size in words of the buffer used for reading keys when searching for a
+ * match, for reading payloads to check if the value has changed, and reading
+ * back from the flash to verify writes. Will work well if big enough for
+ * commonly used keys, and must be at least one word. Stack allocated so not too
+ * large!
```
ourairquality: Looks like it is statically allocated now, which seems ok, so perhaps this 'stack allocated' comment is now stale.

core/sysparam.c
```diff
@@ -119,50 +122,28 @@ static struct {
...
-    CHECK_FLASH_OP(sdk_spi_flash_write(addr, (void*) data, data_size));
-    return SYSPARAM_OK;
-}
-
-static inline IRAM sysparam_status_t _do_verify(uint32_t addr, const void *data, void *buffer, size_t len) {
-    CHECK_FLASH_OP(sdk_spi_flash_read(addr, buffer, len));
-    if (memcmp(data, buffer, len)) {
-        return SYSPARAM_ERR_IO;
+static sysparam_status_t _write_and_verify(uint32_t addr, const void *data, size_t data_size) {
+    static uint8_t bounce[BOUNCE_BUFFER_SIZE];
```
ourairquality: This looks much cleaner using the improved flash support, thanks. Could we move all the flash read/write paths across to the new code now and remove the sdk versions (in follow up work)?

core/sysparam.c
sheinz: Actually, I added static during debugging and just forgot to remove it... I'm not sure what is actually better for that buffer to be static or be in stack.
I would actually prefer it to be in stack as it was before. As there's another such buffer in function `_compare_payload` which is in stack and it's better to be consistent.

core/sysparam.c
```diff
@@ -295,17 +279,39 @@ static sysparam_status_t _find_entry(struct sysparam_context *ctx, uint16_t matc
...
+
+    CHECK_FLASH_OP(spiflash_read(addr, buffer, buffer_size));
+
+    return SYSPARAM_OK;
+}
+
+static inline sysparam_status_t _compare_payload(struct sysparam_context *ctx, uint8_t *value, size_t size) {
+    debug(3, "compare payload (%d) @ 0x%08x", size, ctx->addr);
+    if (ctx->entry.len != size) return SYSPARAM_NOTFOUND;
+    uint32_t *bounce[BOUNCE_BUFFER_WORDS];
```
sheinz: This seems like a typo. The asterisk is redundant. Luckily pointer and uint32_t are the same size, so it worked without errors.

core/sysparam.c
sheinz: I think yes. It was tested quite intensively within spiffs tests.

kanflo:
I think this one has been dormant for too long. It looks good to me and can be merged if no-one has any different opinion.

PR #301, opened 2016-11-29
More convenient GPIO interrupt handlers

UncleRus:
Then:

```C
void gpio00_interrupt_handler(void) //< Weak function override
{}

void gpio01_interrupt_handler(void) //< Weak function override
{}

...
gpio_set_interrupt(0, GPIO_INTTYPE_EDGE_ANY);
gpio_set_interrupt(1, GPIO_INTTYPE_EDGE_ANY);
```

Now:
```C
void handler(uint8_t gpio_num)
{}

...
gpio_set_interrupt(0, GPIO_INTTYPE_EDGE_ANY, handler);
gpio_set_interrupt(1, GPIO_INTTYPE_EDGE_ANY, handler);
```


sheinz:
Hi @UncleRus,

I would say it's a controversial PR :)

What pros I see in the PR:
 * More easy/usual usage style. (you just pass what you want to be called)
 * No extra overhead on a single GPIO interrupt handler. (the same extra call as with the implementation before)
 * Interrupt handlers run-time reconfiguration. The handler can be changed in run-time.

Cons:
 * Uses extra 16 * 4 = 64 bytes of RAM
 * Mandatory extra call for each interrupt handler for time critical applications. (The previous implementation allows to replace a main interrupt handler, to be called straight away)

UncleRus:
@sheinz PTAL

sheinz:
The changes look good.

But I'm still not totally convinced :)
If not taking into account the syntaxes of handling the interrupts.
What functionality does this PR brings? It allows to dynamically change the individual interrupt handlers for single GPIOs in run-time. But how often such a need arises in a project?
And this functionality costs 64 bytes of RAM regardless of it's usage. Not much, but still...

It would be nice to hear other users of eor so, we could decide what approach is better.

The other idea is to make handling of interrupts configurable during compile-time. So, a user might decide what is best for his project: more clear to read code or RAM optimised but somewhat tricky.

UncleRus:
Dynamic interrupt handling very useful in drivers (e.g. software UART).

So I think that the possibility of a dynamic interrupt handling is definitely worth the 64 bytes of RAM.

resetnow:
I like dynamic interrupt handler setup too.

Minimizing interrupt response time is important for real-time applications. The thing is any ESP8266 firmware is far from realtime since there is a whole lot of code running in the background. Since the actual interrupt handler is still weak, the user can replace it.

By the way, isn't `const gpio_interrupt_handler_t gpio_interrupt_handlers` (as it is now) placed in the const data section? If so, then interrupt handling may produce SPI flash read and in this case this PR could actually improve response time and its value distribution.

sheinz:
@resetnow Good point! Thanks for stepping in.

Then, I'm ok with this PR.

PR #302, opened 2016-12-01
Fixes SPI Bug(s)

Zaltora:
Hi i discovered 2 bug with the new SPI API, one is fixed. I still think the way for resolve the second.
First Bug:
Wrong address was set, just swap instructions order to solve it.

Second Bug:
It is a special case that the API does not take into account.
If CS is software , If Address/command/dummy is set and if data transfered is greater than SPI_BUFFER, it will send a second time (or more) addr/dummy/cmd as data :(.

I have some idea to solve it.
1 - If addr/dummy/cmd set, limit transfert to BUF SIZE  ( not sure if it is a good idea )
2 - add extras instructions to clear and  rearm dummy/addr/cmd.

But maybe an other way to solve it is better, some suggestion ?

### DIFF
--- a/core/esp_spi.c
+++ b/core/esp_spi.c
@@ -224,20 +224,20 @@ uint32_t spi_transfer_32(uint8_t bus, uint32_t data)
 static void _rearm_extras_bit(uint8_t bus, bool arm) {
 
     if(!_minimal_pins[bus]) return ;
-    static uint8_t status = 0 ;
+    static uint8_t status[2] ;
 
     if (arm)
     {
-        if (status & (0x01<<(4*bus))) SPI(bus).USER0 |= (SPI_USER0_ADDR) ;
-        if (status & (0x02<<(4*bus))) SPI(bus).USER0 |= (SPI_USER0_COMMAND) ;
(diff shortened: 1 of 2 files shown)


UncleRus:
IMHO the second way is better, it's more predictable behavior.

Zaltora:
Hi, Second bug fixe and tested.  Maybe need some review or optimization.

core/esp_spi.c
```diff
@@ -221,6 +221,26 @@ uint32_t spi_transfer_32(uint8_t bus, uint32_t data)
     return res;
 }
 
+static void _rearm_extras_bit(uint8_t bus, bool arm) {
+
+    if(!_minimal_pins[bus]) return ;
+    static uint8_t status = 0 ;
```
UncleRus: Single status for two buses?

Zaltora: ok i admit, missing this ^^.  I fix it without using one more byte. but  i dont no if it is better to save one byte or juste add one more like uint8_t status[bus].

UncleRus: IMHO `uint8_t status[bus]` is more readable

Zaltora: Well, i will take readable option then. We can optimise spi in other way if needed.

UncleRus:
@Zaltora Thank you!

PR #303, opened 2016-12-01
Fix ignored offset in rtc_mem access.

otopetrik:
Fix sdk_system_rtc_mem_write and sdk_system_rtc_mem_read,
the functions did not use the user-provided offsets to rtc_mem.
Without using offset, rboot's rtcdata access, and its
temp_rom handling were broken.


sheinz:
Hi @otopetrik,

Thanks you very much!
Just checked the disassembly of the binary libraries. It was a mistake during reverse engineering. The original SDK uses the offset.

PR #304, opened 2016-12-01
Allow changing write function of stdout

### LINKED ISSUE #300: printf does not respect sdk_os_install_putc1
A user-supplied implementation of putc does not work in esp-open-rtos like it does for other ESP8266 programs. (e.g. [http://smallbits.marshall-tribe.net/blog/2016/06/15/esp8266-remote-debug](http://smallbits.marshall-tribe.net/blog/2016/06/15/esp8266-remote-debug) and [http://bbs.espressif.com/viewtopic.php?t=218](http://bbs.espressif.com/viewtopic.php?t=218).)
I believe in core/newlib_syscalls.c, in long _write_r(..) both calls to "uart_putc" should instead be calls to "sdk_os_putc", which calls the user-installed putc implementation. 
This way, it is very simple for the end user to switch the serial port that printf messages should appear from.

It appears that overriding/reimplementing _write_r would also work, but it took me hours to even discover where in the code the issue lay, especially since all the documentation around the ESP SDK indicates that it's very simple to install a different uart_putc implementation - so it's my opinion that the change would be a benefit with no downsides.

otopetrik:
It is required for working console output redirect.

Should fix #300.

### DIFF
--- a/core/include/stdout_redirect.h
+++ b/core/include/stdout_redirect.h
@@ -1,0 +1,24 @@
+/*
+ * Part of esp-open-rtos
+ * Copyright (C) 2016 Oto Petrik <oto.petrik@gmail.com>
+ * BSD Licensed as described in the file LICENSE
+ */
+
+#ifndef _STDOUT_REDIRECT_H_
+#define _STDOUT_REDIRECT_H_
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+typedef long _WriteFunction(struct _reent *r, int fd, const char *ptr, int len );
+
+void set_write_stdout(_WriteFunction *f);
+_WriteFunction *get_write_stdout();
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif /* _STDOUT_REDIRECT_H_ */
--- a/core/include/stdout_redirect.h
+++ b/core/include/stdout_redirect.h
@@ -7,6 +7,8 @@
 #ifndef _STDOUT_REDIRECT_H_
 #define _STDOUT_REDIRECT_H_
 
+#include <sys/reent.h>
+
 #ifdef __cplusplus
 extern "C"
 {
@@ -14,7 +16,26 @@ extern "C"
 
 typedef long _WriteFunction(struct _reent *r, int fd, const char *ptr, int len );
 
+/** Set implementation of write syscall for stdout.
+ *
+ *  Use this function to redirect stdout for further processing (save to file, send over network).
+ *  It may be good idea to save result of `get_write_stdout()` beforehand and call
+ *  it at the end of the supplied function.
+ *
+ *  NOTE: use NULL to reset to default implementation.
+ *
+ *  @param[in] f  New code to handle stdout output
+ *
+ */
 void set_write_stdout(_WriteFunction *f);
+
+/** Get current implementation of write syscall for stdout.
+ *
+ *  Save returned value before calling `set_write_stdout`, it allows for chaining
+ *  multiple independent handlers.
+ *
+ *  @returns current stdout handler
+ */
 _WriteFunction *get_write_stdout();
 
 #ifdef __cplusplus
--- a/core/newlib_syscalls.c
+++ b/core/newlib_syscalls.c
@@ -12,6 +12,7 @@
 #include <xtensa_ops.h>
 #include <esp/uart.h>
 #include <stdlib.h>
+#include <stdout_redirect.h>
 
 extern void *xPortSupervisorStackPointer;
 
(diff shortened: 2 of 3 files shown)


sheinz:
Yes, it looks like an obvious solution.

But, I don't like too much function calls. For a single character of printf there's a chain of calls and now it's getting even longer:
printf->_write_r->sdk_os_putc->default_putc->uart_putc(0)

Currently this chain is two calls shorter:
printf->_write_r->uart_putc(0)

I agree that the fix is needed by there must be a better way to handle this.

otopetrik:
Found another issue with original commit - it failed when using spiffs.
New version redirects on block level, and spiffs does not need to know about stdin/stdout handling.

Sdk_os_putc is not used for block level stdout redirection, not sure if it is used anywhere else...
May be worth removing it to avoid confusion.

otopetrik:
Fixed missing "return" when calling to _read_filesystem_r, reading from SPIFFS now works.

otopetrik:
Thank you for the initial review @sheinz.

Last version fixes the long chain issue - it redirects block instead of single character:
_write_r -> _current_stdout_write_r (pointer-to-function)
It does not break when linked with SPIFFS:
_write_r -> _write_filesystem_r (weak symbol, esp_spiffs.c overrides) -> SPIFFS_write 

Could you please review it ?

extras/spiffs/esp_spiffs.c
```diff
@@ -125,32 +125,15 @@ int32_t esp_spiffs_mount()
...
-    for(int i = 0; i < len; i++) {
-        /* Auto convert CR to CRLF, ignore other LFs (compatible with Espressif SDK behaviour) */
-        if(ptr[i] == '\r')
-            continue;
-        if(ptr[i] == '\n')
-            uart_putc(0, '\r');
-        uart_putc(0, ptr[i]);
-    }
-    return len;
+    return SPIFFS_write(&fs, (spiffs_file)fd, (char*)ptr, len);
```
sheinz: Yes. That makes more sense to handle only filesystem stuff here 👍

core/newlib_syscalls.c
```diff
@@ -41,12 +42,8 @@ IRAM caddr_t _sbrk_r (struct _reent *r, int incr)
 }
 
 /* syscall implementation for stdio write to UART */
-__attribute__((weak)) long _write_r(struct _reent *r, int fd, const char *ptr, int len )
+__attribute__((weak)) long _write_stdout_r(struct _reent *r, int fd, const char *ptr, int len )
```
sheinz: Do we really need it to be `weak`? We already have weak `_write_r` in case user needs to override it completely. And have a run-time selection of stdout write function.
So, I think it might be a static function without underscore prefix.

core/newlib_syscalls.c
```diff
@@ -58,6 +55,37 @@ __attribute__((weak)) long _write_r(struct _reent *r, int fd, const char *ptr, i
     return len;
 }
 
+_WriteFunction *_current_stdout_write_r = &_write_stdout_r;
```
sheinz: This also could be made static without underscore.

core/include/stdout_redirect.h
```diff
@@ -0,0 +1,24 @@
+/*
+ * Part of esp-open-rtos
+ * Copyright (C) 2016 Oto Petrik <oto.petrik@gmail.com>
+ * BSD Licensed as described in the file LICENSE
+ */
+
+#ifndef _STDOUT_REDIRECT_H_
+#define _STDOUT_REDIRECT_H_
```
sheinz: This header uses `_reent` structure so it's better to include `<sys/reent.h>` here.

core/include/stdout_redirect.h
```diff
@@ -0,0 +1,24 @@
...
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif
+
+typedef long _WriteFunction(struct _reent *r, int fd, const char *ptr, int len );
+
+void set_write_stdout(_WriteFunction *f);
```
sheinz: It would be nice to have a comment here that specifies that `f` could be NULL for resetting to the default stdout writer.

core/include/stdout_redirect.h
```diff
@@ -0,0 +1,24 @@
...
+_WriteFunction *get_write_stdout();
```
sheinz: I'm not sure how much use would be of the getter function. But if you think it is needed than I'm ok with it :)

core/newlib_syscalls.c
otopetrik: ok, will fix in next version.

core/include/stdout_redirect.h
otopetrik: thanks, missed that one.

core/include/stdout_redirect.h
otopetrik: It allows multiple implementations of console redirection libraries.
If the redirection library calls the old function (saved from gettter) at the end of its newly installed one, the libraries can stack.
A library that sends console data over a network and a library that logs them to filesystem can work at the same time (and still display data to serial console)

UncleRus:
@sheinz Can this PR be merged? It looks good to me.

sheinz:
@UncleRus Yes, I think so. Looks good for me too.

skcks:
If want to redirect stdout to a file in SPIFFS, what is need to do?any clue will be helpful, thanks

skcks:
is correct  the code?  @otopetrik @sheinz 
`
ssize_t log_writer(struct _reent *r, int fd, const void *ptr, size_t len) {
    int fd_log = open("excinfo.log", O_WRONLY | O_CREAT | O_APPEND, 0);
    if (fd_log < 0) {
        return 0;
    }

    len = write(fd_log, (uint8_t *)ptr, len);
    close(fd_log);

    return len;
}

set_write_stdout(log_writer);

`

PR #319, opened 2016-12-29
add some missing include & fixed display output on sh1104

tuanpmt:
I got some errors when build with esphttpd

### DIFF
--- a/core/app_main.c
+++ b/core/app_main.c
@@ -21,6 +21,8 @@
 #include "esp/spi_regs.h"
 #include "esp/dport_regs.h"
 #include "esp/wdev_regs.h"
+#include "esp/wdt_regs.h"
+#include "esp/rtcmem_regs.h"
 #include "esp/hwrand.h"
 #include "os_version.h"
 
--- a/core/debug_dumps.c
+++ b/core/debug_dumps.c
@@ -19,6 +19,7 @@
 #include "xtensa_ops.h"
 #include "esp/rom.h"
 #include "esp/uart.h"
+#include "esp/dport_regs.h"
 #include "espressif/esp_common.h"
 #include "esplibs/libmain.h"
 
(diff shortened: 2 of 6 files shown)


Zaltora:
What problem you got with SH1106 ?   It is display good with me.  
Maybe you forget to change DISPLAY_WIDTH  to  132  because SH1106 got  132*64 buffer, even if screen is 128*64.
Need maybe update example with this details

tuanpmt:
My display is 128x64, as the datasheet, `sh1106_go_coordinate` execute every page = 128bytes width. I can not using this library without this condition `i%dev->width == 0`

https://github.com/tuanpmt/esp-audio-player/blob/master/extras/player/eq.c#L35

Zaltora:
This fix make sense, because currently go_coordinate is call every 16 bytes. Great fix @tuanpmt 
Just a point, i checked datasheet, the sh1106 got a 132 X 64 bits SRAM.
ref: https://www.elecrow.com/download/SH1106%20datasheet.pdf
That why in `sh1106_go_coordinate`, you got +2 offset. To center for 128X64  true screen format. I didn't know what happens if i use 128  in DISPLAY_WIDTH. If it is work well for you, maybe this part of ram is ignored. I will test that with SPI protocols.

Zaltora:
It is work well. ^^

core/app_main.c
```diff
@@ -21,6 +21,8 @@
 #include "esp/spi_regs.h"
 #include "esp/dport_regs.h"
 #include "esp/wdev_regs.h"
+#include "esp/wdt_regs.h"
+#include "esp/rtcmem_regs.h"
```
ourairquality: Were these includes really needed? It was not giving any warnings last time I built? Might these be included by mistake, some other local changes?

tuanpmt: the `RTCMEM_BACKUP` define inside `rtcmem_regs.h`, `WDT` inside `wdt_regs.h`. It's normal to me, except that I compiled together with esphttpd.

ourairquality: ok, thanks.

PR #321, opened 2017-01-02
I2c optimization and features

Zaltora:
Hi, i do some work on i2c.

Currently done:
Add custom delay to get true 400kHz at 160Mhz and  320KHz at 80Mhz ( cant be faster with 80 MHz for now )
Add status flag check ( usefull when more than one task use i2c ) ( Level 1 API dont start if bus busy) 
Rewrote comment (Update them)
Rework printf debug

My plans (need opinion):
-Selectable frequency. (100kHz / 400kHz )
-Add force option. ( if bus busy, stop com and start a new one  for level1 API)
-Add possibility to write and read 16 bit data.
-Change data from `i2c_slave_read(uint8_t slave_addr, uint8_t data, uint8_t *buf, uint32_t len)` to a data pointer. Function dont send register address when NULL. 
-Optional (if time permit): Add multiple i2c setting support.

What is better: 
Add new functions to include 16 bits possibility or add  data format to current function ?
Write a new "read" function without register data or change him to a pointer  ?   (other idea ?)

### DIFF
--- a/extras/i2c/i2c.h
+++ b/extras/i2c/i2c.h
@@ -103,7 +103,8 @@ bool i2c_status(void);
 //Level 1 API (Don't need functions above)
 
 /**
- * This function will allow you to force a transmission I2C, canceled the current one.
+ * This function will allow you to force a transmission I2C, cancel current transmission.
+ * Warning: Use with precaution. Don't use it if you can avoid it. Usefull for priority transmission.
  * @param state Force the next I2C transmission if true (Use with precaution)
  */
 void i2c_force_bus(bool state);
@@ -116,7 +117,7 @@ void i2c_force_bus(bool state);
  * @param len Number of byte to send
  * @return Non-Zero if error occured
  */
-int i2c_slave_write(uint8_t slave_addr, uint8_t *data, uint8_t *buf, uint32_t len);
+int i2c_slave_write(uint8_t slave_addr, const uint8_t *data, const uint8_t *buf, uint32_t len);
 
 /**
  * Issue a send operation of 'data' register adress, followed by reading 'len' bytes
@@ -127,7 +128,7 @@ int i2c_slave_write(uint8_t slave_addr, uint8_t *data, uint8_t *buf, uint32_t le
  * @param len Number of byte to read
  * @return Non-Zero if error occured
  */
-int i2c_slave_read(uint8_t slave_addr, uint8_t *data, uint8_t *buf, uint32_t len);
+int i2c_slave_read(uint8_t slave_addr, const uint8_t *data, uint8_t *buf, uint32_t len);
 
 #ifdef	__cplusplus
 }
--- a/extras/i2c/i2c.h
+++ b/extras/i2c/i2c.h
@@ -42,6 +42,7 @@ extern "C" {
  *  Some bit can be transmit slower.
  *  Selected frequency fix the speed of a bit transmission
  *  I2C lib take the maximum frequency defined
+ *  Don't change frequency when I2C transaction had begin
  */
 
 #ifdef I2C_FREQUENCY_500K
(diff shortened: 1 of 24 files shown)


resetnow:
I think using volatile variable for delay may produce different actual delay with different optimization levels. Even though it doesn't get any simpler than `addi` and `s32i` in a loop, compiler may still unroll the loop.

Something like this might produce more consistent results:

```C
uint32_t delay;
__asm volatile (
	"loop: movi %0, %1" "\n"
	"addi %0, %0, -1" "\n"
	"bnez %0, loop" "\n"
: "=a" (delay) : "i" (5));
```

You can fine-tune the speed by putting `nop` instructions after the loop.

extras/i2c/i2c.h
```diff
@@ -32,26 +32,71 @@
 extern "C" {
 #endif
 
-// Init bitbanging I2C driver on given pins
+// I2C driver for ESP8266 written for use with esp-open-rtos
+// Based on https://en.wikipedia.org/wiki/I²C#Example_of_bit-banging_the_I.C2.B2C_Master_protocol
+// With calling overhead, we end up at ~320kbit/s
+
+///////Level 0 API
```
resetnow: It's probably better to stick with a uniform comment style everywhere, i.e. `//`

Zaltora: okay, i will fix it.

Zaltora:
thank you for your help @resetnow , i had hesitate to use asm because i got problem when trying to replace fix value 5 by a variable :(.
I think, asm delay can be applied to ws2812 driver too for same reason?

resetnow:
>i got problem when trying to replace fix value 5 by a variable :(.

That's because this snippet is made to explicitly accept immediate arguments. To avoid magic numbers, maybe it makes sense to make 5 a macro. Using a variable is also possible (the same `movi` instruction is now generated by the compiler) :

```C
uint32_t delay = 5;
__asm volatile (
	"1: addi %0, %0, -1" "\n"
	"bnez %0, 1b" "\n"
: "=a" (delay));
```

>asm delay can be applied to ws2812 driver too for same reason

Likely yes.

Zaltora:
Update WIP:
-Added ASM delay.
-Added 100KHz, 400Khz or 500 KHz selection  ( measured ok but its Approximate)
-Added 16 bits data transfert I2C ( tested with ina3221 ok!)

Need feedback for these points:
**Rework transfert:** 
I had just created new api function for 16bits to show  prototype format and know if it is good for all to update 8bits functions like them and for all lib. I think this allows more choice of transfer for the library.
Added Improvement of error handling ( EBUSY and EIO ) : make sens or keep old system ?
```
bool i2c_slave_write(uint8_t slave_addr, uint8_t *buf, uint8_t len);
bool i2c_slave_read(uint8_t slave_addr, uint8_t data, uint8_t *buf, uint32_t len);
```
to
```
int i2c_slave_write(uint8_t slave_addr, uint8_t *data, uint8_t *buf, uint8_t len);
int i2c_slave_read(uint8_t slave_addr, uint8_t *data, uint8_t *buf, uint32_t len);
```

**Rework of flag system**:
For test, i created a new flag variable, Level 0 API dont check it right now. reason: dont include transfert timing when i prevent task change here:
```
static int i2c_bus_test(bool force)
{
    taskENTER_CRITICAL(); // To prevent task swaping after checking flag and before set it!
    bool status = flag ; // get current status
    if(force)
    {
        flag = true ; // force bus on
        taskEXIT_CRITICAL();
        if(status)
           i2c_stop(); //Bus was busy, stop it.
    }
    else
    {
        if (status)
        {
            taskEXIT_CRITICAL();
            debug("busy");
            taskYIELD(); // If bus busy, change task to try finish last com.
            return -EBUSY ;  // If bus busy, dont read
        }
        else
        {
            flag = true ; // Set Bus busy
            taskEXIT_CRITICAL();
        }
    }
    return 0 ;
}
```
I use this prevent to be sure than task don't change between a check and a set.I I dont want include time from " i2c_stop()", that why i created new flag. Everything is ok ? idea to improve this ? or maybe it is useless ?

only 16 bits transfert got the busy protection, i will update everything after feedback.

Zaltora:
I had doing some test with ina3221 example reworked.
One task read flag every 10 ms ;
Second task read value if flag good ;
No EIO error,  bus busy detected

Zaltora:
I will remove 16 bits api after update 8 bits function. Is not really usefull.

extras/i2c/i2c.h
```diff
@@ -27,16 +27,39 @@
...
+
+/*
+ *  Some bit can be transmit slower.
+ *  Selected frequency fix the speed of a bit transmission
+ *  I2C lib take the maximum frequency defined
+ */
+
+#ifdef I2C_FREQUENCY_500K
+#define I2C_CUSTOM_DELAY_160MHZ   6
+#define I2C_CUSTOM_DELAY_80MHZ    1   //Sry, maximum is 320kHz at 80MHz
```
resetnow: Something that could be done to squeeze the last bit of performance and go up to 400 kHz is avoiding `sdk_system_get_cpu_freq` call in the delay function. Instead, the state of cpu frequency can be cached into a `static uint8_t` variable at the beginning of the transaction. This way, there will be no actual `call0` performed at every delay call, only `l32r`, `l32i` and a comparison. The downside is that CPU frequency should not change during i2c transactions — a reasonable requirement IMO. 

I'm not saying you should implement this in this PR — you already did a lot of work. Just for future reference. I'm out of esp modules currently but I'll try to do it myself once I receive a new batch.

Zaltora: It is like a good idea to get performance like this  and store frequency at `i2c_start() `for example but the link will be less secure for a special case (Probability very low, but murphy is here ). 
When you start a transaction, if the task change and second task change frequency ( 80 => 160 ) , when the i2C link come back, delay will be very low and frequency will be high ( approximately  800KHz for 400KHz base link ). Can expect EIO error.  Of course user can prevent this manually with status from i2c bus when he need change frequency or doing some treatments from this error. The component can support high frequency too.

Zaltora:
The new api will look like this :
```
int i2c_slave_write(uint8_t slave_addr, uint8_t *data, uint8_t *buf, uint32_t len, bool force);
int i2c_slave_read(uint8_t slave_addr, uint8_t *data, uint8_t *buf, uint32_t len, bool force);
```
I think that all lib can be update with Level 1 API .  Level 0 API dont implement flag security. Even if we got i2c_status(), Level 1 API is more secure.

extras/i2c/i2c.h
Zaltora: `a reasonable requirement IMO`  Okay then :)

Zaltora:
It is better that each lib do special treatment when I2C bus busy or that level 1 API  retry each time busy line detected ? Can be infinite loop if I2C bus permanently used by other task.

resetnow:
Usually multiple tasks share a semaphore that restricts access to a single task. I don't think this should be implemented inside the library.

Zaltora:
hi, i have done few fixes and test.  I think core I2C work well now. Need tests for some library to confirm that changes on  i2c components lib is good. i will do some test again on i2c frequency. Already tested with example:
-ssd1306  ok
-ds3231 lok
-ina3221 ok
-bmp280 ok

resetnow:
I looked over this yesterday and here are my thoughts:

1. Do we really need to pass register address as a pointer? 90% of devices with i2c I worked follow the pattern slave_address — reg_address — data. For the few of those doing it differently the read/write sequence can be implemented separately.
2. Do we really need to pass force flag with every read/write? It seems to me that forcefully stopping existing communication isn't something you would want to do often.

Zaltora:
Hi @resetnow , thank you for your feedback. Interesting thing.
1- I have choice to create one function because i was thinking that is not so much less used like you said. I use many sensors that need raw reads and some components that use register can be read without specify a register to be faster when reading a lot of data ( when start from 0 or when you know already where is the current register pointer  for example ). I'm not against the option of having two functions because I hesitated to do it. If it is better, i can do it.

2- you are right,  we will not use this often. In my case i want do some priority reading on a specific sensor with a timer.  Just thinking to add 1 function like 
i2c_force_bus();  => This function will force the next transaction level 1 api. you need rearm it or
i2c_force_bus(bool state); => This function will force all i2c until he is setting off

What did you think ?

resetnow:
I'll wait for some feedback from others.

andree182:
Ad 1) - I vote for having the pointer. Just make it (const uint_8*). Same for the write.

andree182:
Is there some blocker regarding this? Or is it simply not going to get merged?

resetnow:
@sheinz @UncleRus what do you think?

Zaltora:
I use this PR in my project. No problem detected with library i use (ds3231, ina3221, bmp280, ...). I have done some test with multi-task I2C without using semaphore. Reading and writing work well. Busy line properly detected.
I have done some test with forced option too. Didn't see problems. ( I mostly read device than write into them)
Some frequency bus need to be adjusted but it is fine.

I got not time right now but i was thinking to add bus system like SPI lib or Softuart and remove all I2C init from some library. It can be done in an other PR.

UncleRus:
@resetnow For me, this PR is good enough.

UncleRus:
Going to merge it if no objections.

PR #324, opened 2017-01-08
HTTP server

lujji:
I've added extras/httpd from LwIP and created an example HTTP server with CGI and SSI handlers to demonstrate basic functionality ([screenshot](https://cloud.githubusercontent.com/assets/9296247/21746461/d4e44380-d53d-11e6-9b8f-cd9609c7924e.png)). Original code is MIT licensed.
The only issue is that build process depends on a perl script, which generates C structures from files inside 'fs' directory.


sheinz:
Hi @lujji,

Thanks for the contribution.
esp-open-rtos have lacked http server support for a long time. It's really cool that finally there's some light in this direction.

PR #330, opened 2017-01-26
bh1750 i2c light sensor driver + pwm cleanup

andree182:
Hi, this is a "preliminary" driver pull request - so that others know the driver already exist and they don't have to rewrite it (though it's trivial).

Since I did change extras/i2c lib a bit, and there's pull request #321 in the queue for similar functionality (but cleaner), I'll rebase this once #321 is merged - basically the commit 0b1197b can be dropped.

### DIFF
--- a/extras/bh1750/bh1750.h
+++ b/extras/bh1750/bh1750.h
@@ -1,0 +1,82 @@
+/*
+ * Driver for BH1750 light sensor
+ *
+ * Part of esp-open-rtos
+ * Copyright (C) 2017 Andrej Krutak <dev@andree.sk>
+ * BSD Licensed as described in the file LICENSE
+ *
+ * ROHM Semiconductor bh1750fvi-e.pdf
+ */
+#ifndef EXTRAS_BH1750_H_
+#define EXTRAS_BH1750_H_
+
+#include <stdint.h>
+#include <stdbool.h>
+#include <time.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/* Possible chip addresses */
+#define BH1750_ADDR_LO 0x23 // ADDR pin floating/low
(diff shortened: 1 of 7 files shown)


extras/bh1750/bh1750.h
```diff
@@ -0,0 +1,65 @@
...
+// Device is automatically set to Power Down after measurement.
+#define BH1750_ONE_TIME_HIGH_RES_MODE  0x20
+
+// Start measurement at 0.5lx resolution. Measurement time is approx 120ms.
+// Device is automatically set to Power Down after measurement.
+#define BH1750_ONE_TIME_HIGH_RES_MODE_2  0x21
+
+// Start measurement at 1lx resolution. Measurement time is approx 120ms.
+// Device is automatically set to Power Down after measurement.
+#define BH1750_ONE_TIME_LOW_RES_MODE  0x23
```
UncleRus: Better to use enums instead of all this defines

UncleRus: Better to declare enums instead of all this macros.
Something like
```C
typedef enum
{
     BH1750_MODE_CONTINUOUS_HIGH_RES = 0x10,
     ...
} bh1750_mode_t;
```

UncleRus:
It would be great if you add simple example.

Thank you for your contribution!

panoti:
Why isn't this pull-request not merged?

andree182:
Waiting for #321 first (to rebase this...), then I'll also add an example..

andree182:
Hi guys. I reworked the patches to match new i2c, added example, and added a small cleanup of PWM. I didn't do the enum thing, since the flags can now be combined (for which enums are a bit dangerous - although here only 1 byte stuff is moved around).

PR #331, opened 2017-01-30
httpd: WebSockets

lujji:
WebSockets implementation on top of httpd. Binary and text modes are supported, opening multiple sockets is possible. Continuation frames are not implemented. Example code demonstrates streaming ADC readings, and controlling LED.


UncleRus:
Thank you, great contribution!
Will merge it tomorrow if no objections.

lujji:
Thanks. It just so happens that I've added large frames support recently, so I'd appreciate if you merge my next commit.

PR #339, opened 2017-02-12
CRC library

Zaltora:
Hi, i have finish the first release of my crc library based on many things found about CRC ( details in my repository). I post it as submodule because the code can be ported for other things. 

- This library can compute crc from 1 to 64 bit.
- Every parameters of crc is present. (Poly,Init,RefIn,RefOut,XorOut )
- Independent crc struct for multiple usage.
- You can select algorithm you want use (4 at this moment. Thank to Sven Reifegerste original lib)
- Sized functions ( 8bits, 16 bits, 32bits and 64 bits) for save RAM usage if needed.
- Pre-generated  constant lookup table for many crc standard 
- Can generate  lookup table  for polynom multiple of 8bits ( 8,16,24,32,..)

It is the first release. Maybe some bugs are present for custom crc. Some code optimisation can be done too.  Someone got idea to add specific freertos headers to a config header ? i was thinking that library want  search a header in user application with a specific include like 
#include "config_crc_file.h"
and user need always create this files when he use the lib to specifies header dependant project and crc config options. Good idea ? 

Optimised functions for specific crc not present. Custom hardware crc functions selection not present (not needed with esp8266). Other features like error management will be added later.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -22,3 +22,6 @@
 [submodule "extras/http-parser/http-parser"]
 	path = extras/http-parser/http-parser
 	url = https://github.com/nodejs/http-parser
+[submodule "extras/crc_generic/crc_lib"]
+	path = extras/crc_generic/crc_lib
+	url = https://github.com/Zaltora/crc_generic_lib.git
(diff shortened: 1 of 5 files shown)


UncleRus:
Very nice contribution, @Zaltora !

> Good idea ?

Totally.

Zaltora:
Hi, Can we reopen this branch ? i doing some update on current API and crc lib. I think this update need to be a part of this branch because it is do some fix ( specially on include )
Or i open a new one ?

UncleRus:
Just open the new one :)

PR #348, opened 2017-03-08
Feature/mdnsresponder

mjahamel:
This is a basic multicast DNS responder which should be useful for locating ESP-based devices on the local LAN that have been assigned addresses by a DHCP server. 
Limitations/bugs: it only advertises a single service on a single port, and it understands compressed DNS queries but does not generate them.


UncleRus:
Nice contribution, thank you!
It would be great if you'll add example app.

PR #352, opened 2017-03-15
Fixed error of writing to SPI registers with memcpy

UncleRus:


jkent:
This seems to have broken 16-bit transfers and I'm guessing 32-bit transfers as well.  spi_transfer_16(1, 0x8000) on scope results in 0x8080.

UncleRus:
I reverted this commit, but without it we have another issue.
Seems `memcpy()` incorrectly working when copying 3 bytes to hardware registers.

May be this is related with unaligned read from SPI flash.

ourairquality:
Where does this leave the state of the code, and any ideas on how to move forward?

The older aligned memcpy versions appear to use word sizes operations and also a `memw` after each access to the spi buffer. Are either or both needed, and would that resolve the problem? Would the fix be to bring back `memcpy_unaligned_src` and `memcpy_unaligned_dst` and add them to the core code now?

Looking at the code, in `spi_transfer` there is a call to `_rearm_extras_bit` in the following that looks like dead code and might this point to a problem too?

```
    for (size_t i = 0; i < blocks; i++)
    {
        size_t offset = i * _SPI_BUF_SIZE;
        _spi_buf_transfer(bus, (const uint8_t *)out_data + offset,
            in_data ? (uint8_t *)in_data + offset : NULL, buf_size, e, word_size);
        if (blocks) _rearm_extras_bit(bus, false);
    }
```

UncleRus:
This is a very strange bug in memcpy(): when copying three (exactly three!) bytes into the register area, for example `SPI(1).W`, the third byte is repeated three times:

```C
const uint8_t data[] = {
	0x01, 0x02, 0x03
};
const size_t data_size = sizeof(data);
...
memset((void *)SPI(BUS).W, 0, sizeof(SPI(BUS).W));
dump("ZERO", (uint8_t *)SPI(BUS).W, data_size);
memcpy((void *)SPI(BUS).W, data, data_size);
dump("SRC", data, data_size);
dump("DST", (uint8_t *)SPI(BUS).W, data_size);
```
Output:
```
ZERO: 0x00 0x00 0x00 
SRC:  0x01 0x02 0x03 
DST:  0x03 0x03 0x03 
```

4 bytes output:
```
ZERO: 0x00 0x00 0x00 0x00
SRC:  0x01 0x02 0x03 0x04
DST:  0x01 0x02 0x03 0x04
```

Just tried `memw` before and/or after copying bytes, and nothing changes.

pfalcon:
I believe that all esp8266 registers are 32-bit and require aligned access, i.e. using memcpy() to write (or read) anything to a register isn't correct as it doesn't provide such guarantees.

resetnow:
Most likely SPI registers don't process "byte enable" signal of system bus (AHB?) at all, therefore, subsequent byte writes still write the full 32 bit word which contains remains of a previous transaction.

PR #355, opened 2017-03-20
Give FreeRTOS Hooks weak linkage

andrewclink:
Sometimes it's nice to have access to the FreeRTOS tick and idle hooks. This pull request enables that by allowing user code to override the default implementations provided in `app_main.c`.

To use this, include a file in your project called `FreeRTOSConfig.h` with (at least) the contents shown in the example. You may need to `make clean` to cause FreeRTOS to be rebuilt. Overrides for `vApplicationIdleHook` and `vApplicationTickHook` can now be provided.

**One Caveat** with this pull request specifically is that I do not yet fully understand the sleep functions provided in the SDK. I believe that if `WIFI_SLEEP_LIGHT` is passed in the idle hook, the wifi radio will sleep if connected (or the call will fail if not connected). Either way, the call is noisy and that makes timing and whether the sleep call is working at all difficult to discern. If it's easier to just pull out the example that's fine too.

### DIFF
--- a/core/app_main.c
+++ b/core/app_main.c
@@ -236,12 +236,12 @@ void IRAM vApplicationStackOverflowHook(TaskHandle_t task, char *task_name) {
 }
 
 // .text+0x3d8
-void IRAM vApplicationIdleHook(void) {
+void __attribute__((weak)) IRAM vApplicationIdleHook(void) {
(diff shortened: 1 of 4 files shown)


UncleRus:
Thank you!

PR #356, opened 2017-03-21
Add a TSL4531 driver and example

bschwind:
Hey I'm back again with another light sensor driver, this time the TSL4531.

[TSL4531 Light Sensor](https://media.digikey.com/pdf/Data%20Sheets/Austriamicrosystems%20PDFs/TSL4531.pdf)

It uses the I2C module from the `extras` directory of this project.

This driver supports setting an integration time from the presets of (100ms, 200ms, and 400ms), as well as a "skip power save" mode which just comes straight from the data sheet. I played around with the "skip power save" mode a bit but didn't see any noticeable differences. Perhaps it draws less current but I'm not sure at the moment.

The driver uses `vTaskDelay` when the lux value is requested from the sensor so it has time to integrate the sensors and get a reading. The sensor remains in a "powered down" mode when not in use.

The cool thing about this sensor is it calculates the lux onboard so we can get it straight from the I2C bus. Very convenient, and takes up very little space.

### DIFF
--- a/examples/tsl4531/Makefile
+++ b/examples/tsl4531/Makefile
@@ -1,0 +1,3 @@
+PROGRAM=tsl4531_example
+EXTRA_COMPONENTS = extras/i2c extras/tsl4531
+include ../../common.mk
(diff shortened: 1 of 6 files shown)


UncleRus:
Hi!
You need to refactor your code according to #321

PR #358, opened 2017-03-23
Use WEB_PORT define instead of hardcoded "80" string

hugows:


resetnow:
@UncleRus could you please take a look at this?

examples/http_get/http_get.c
```diff
@@ -21,7 +21,7 @@
 #include "ssid_config.h"
 
 #define WEB_SERVER "chainxor.org"
-#define WEB_PORT 80
+#define WEB_PORT "80"
 #define WEB_URL "http://chainxor.org/"
```
ourairquality: A `Host` header may be needed too and the port included in that. e.g. `"Host: chainxor.org:8000"` Might be nice to avoid the host name duplication too. So perhaps keep the port as a number and use `snprintf` to build up the headers.

UncleRus:
Agreed with @ourairquality.

PR #369, opened 2017-04-19
Fixes unaligned writes to SPI data register

UncleRus:
Seems it's OK to use memcpy() for unaligned read or aligned write to the register area, but we need a special method for unaligned writes.
See #352.

### DIFF
--- a/core/esp_spi.c
+++ b/core/esp_spi.c
@@ -159,6 +159,22 @@ inline static void _start(uint8_t bus)
     SPI(bus).CMD |= SPI_CMD_USR;
 }
 
+inline static void _store_data(uint8_t bus, const void *data, size_t len)
+{
+    uint8_t words = len / 4;
+    uint8_t tail = len % 4;
+
+    memcpy((void *)SPI(bus).W, data, len - tail);
+
+    if (!tail) return;
+
+    uint32_t last = 0;
+    uint8_t *offs = (uint8_t *)data + len - tail;
+    for (uint8_t i = 0; i < tail; i++)
+        last = last | (offs[i] << (i * 8));
+    SPI(bus).W[words] = last;
+}
+
 inline static uint32_t _swap_bytes(uint32_t value)
 {
     return (value << 24) | ((value << 8) & 0x00ff0000) | ((value >> 8) & 0x0000ff00) | (value >> 24);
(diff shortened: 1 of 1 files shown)


core/esp_spi.c
```diff
@@ -159,6 +159,22 @@ inline static void _start(uint8_t bus)
     SPI(bus).CMD |= SPI_CMD_USR;
 }
 
+inline static void _store_data(uint8_t bus, const void *data, size_t len)
+{
+    uint8_t words = len / 4;
+    uint8_t tail = len % 4;
+
+    memcpy((void *)SPI(bus).W, data, len - tail);
```
ourairquality: Might this still be a problem if the source `data` is unaligned? Might `memcpy` write some leading bytes in that case?

core/esp_spi.c
```diff
@@ -269,7 +283,7 @@ size_t spi_transfer(uint8_t bus, const void *out_data, void *in_data, size_t len
         size_t offset = i * _SPI_BUF_SIZE;
         _spi_buf_transfer(bus, (const uint8_t *)out_data + offset,
             in_data ? (uint8_t *)in_data + offset : NULL, buf_size, e, word_size);
-        if (blocks) _rearm_extras_bit(bus, false);
+       _rearm_extras_bit(bus, false);
```
ourairquality: Could you consider adding a comment to `_rearm_extras_bit` on what it does. It seems to be to toggle off the address in the middle of transfer needing multiple blocks, so the address is only sent at the start?

Is using the full `_SPI_BUF_SIZE` ok? I recall mention of a possible hardware problem using a size of 64 for reads, and it notes that the esp rom code uses a size of 32 bytes, see http://www.esp8266.com/viewtopic.php?f=13&t=10843 Might it be prudent to change the size to 32, or at least 60?

core/esp_spi.c
UncleRus: Tested it with unaligned out_data, unaligned in_data, different sizes, it works.

core/esp_spi.c
UncleRus: > comment to _rearm_extras_bit on what it does
We can ask @Zaltora about it

> Is using the full _SPI_BUF_SIZE ok?
Tested it with 64 bytes, works ok.

UncleRus: About `_rearm_extras_bit(false)`: looking at this comment in spi.h:
```C
/*
 * Possible Data Structure of SPI Transaction
 *
 * [COMMAND]+[ADDRESS]+[DataOUT]+[DUMMYBITS]+[DataIN]
 *
 * [COMMAND]+[ADDRESS]+[DUMMYBITS]+[DataOUT]
 */
```
I believe this code just clearing [COMMAND] + [ADDRESS] bits while transferring  between the [DataOUT] blocks.

core/esp_spi.c
resetnow: The only downside is this code still depends on memcpy implementation.

core/esp_spi.c
Zaltora: `_rearm_extras_bit` function is to toggle off extras bit like address, dummy and command if data to send is higher than buffer.
The problem was only if the user use software CS. He need  manually toggle CS before send again ADDR/DUMMY/CMD . If the SPI device receive more data than buffer, he will receive theses bits as data.
This function is not needed if we are using hardware pin because esp toggle CS automatically.

faryon93:
+1 you saved my day! :)

UncleRus:
Going to merge it

PR #370, opened 2017-04-21
Fix for counter overflow in delays based on sdk_system_get_time()

UncleRus:


extras/sdio/sdio.c
```diff
@@ -141,6 +141,8 @@ static uint16_t crc_ccitt(const uint8_t *data, size_t n)
 #define spi_skip_word()   do { spi_read_byte(); spi_read_byte(); } while(0)
 #define spi_skip_dword()  do { spi_read_byte(); spi_read_byte(); spi_read_byte(); spi_read_byte(); } while(0)
 
+#define timeout_expired(start, len) ((uint32_t)(sdk_system_get_time() - (start)) >= (len))
```
kanflo: Isn't it still possible for this to be problematic if sdk_system_get_time wraps before len elapsed, meaning start > sdk_system_get_time?

UncleRus: No, it's not because of the unsigned math:
```C
uint32_t diff = (uint32_t)10000 - (uint32_t)0xfffffff0;
printf("Difference: %d\n", diff);
```
gives
```
Difference: 10016
```

kanflo: Oh, didn't think of that. I am all for merging this.

PR #372, opened 2017-04-25
8-bit digital-to-analog conversion PCF8591 library

panoti:


panoti:
I have to change "i2c_delay" by sdk_os_delay_us(10) for working. Maby pcf8591 can not work with high frequency.

Zaltora:
Yes, see  datasheet. max frequency is 100kHz.

try change 
 `#define I2C_FREQUENCY_400K true // for test WIP`
to
`#define I2C_FREQUENCY_100K true // for test WIP`

in i2c.h line 39

panoti:
Change into "#define I2C_FREQUENCY_100K true" may be equivalent with "#define I2C_FREQUENCY_400K false" which I have commented in my code. I also changed it but my program still catch some read errors (just a while). So, I use sdk_os_delay_us(10) instead. Are you sure this code
```
#define I2C_CUSTOM_DELAY_160MHZ   100
#define I2C_CUSTOM_DELAY_80MHZ    20
```
is equivalent with `sdk_os_delay_us(10)`?

Zaltora:
No, is not equivalent. With 10us delay, you are at 50kHz  ( 20us per period)
The goal was to reproduce as most as possible i2c clock with 100kHz. But it can be faster (clock signal are not perfect). Maybe that why is not workung
You can try to increase them by adding a new parameter 50KHZ and test:

```
#define I2C_CUSTOM_DELAY_160MHZ   200
#define I2C_CUSTOM_DELAY_80MHZ    40
```
If you get oscilloscope, you can adjust them.

panoti:
yep, so can this branch be merge? I think frequency adjusting should have for each user. I just add new lib and example without editing anything in i2c lib.

Zaltora:
Yes it can be merge.  frequency system and measures for i2c can be done in an other PR

PR #389, opened 2017-06-06
LwIP v2 support

ourairquality:
Getting more confident with this change, and requesting wider testing and feedback. It is working fine here in limited testing, in station and softap modes, with a http server and dns server and the dhcp server and client.

Tracking the upstream lwip master as it is adding some useful changes that help, particularly for pbuf management that helps tracking the esp rx pbufs.

Couldn't get all the dependant code decompiled to source, so had to do some minor patching of the binaries. But double checked all the code that has been converted to source.

Could this ever make it back into this repo or will it always be too much of a step?


UncleRus:
Closing/reopening this PR to trigger Travis CI build.

ourairquality:
@UncleRus Thank you the build is looking much better now. Remains to update the extras/httpd to get this all building, but everything else builds here.

ourairquality:
@lujji I took a quick look at updating your http server, and have had to disable it for now so the build checks pass. Btw lwip now includes httpd in the main source code. The addition of the websocket support makes it more of a challenge for me to update this example. One benefit of getting this updated might be that your code would be based off the lwip master branch and might be a candidate to integrate there. One reason I am so keen to get lwip updated is to try to avoid wasted resources working with the old 1.4 version. Would you be interested in updating your http server example? Would people accept landing the lwip v2 change with this example disabled?

lujji:
@ourairquality Great job on bringing lwip v2 to the project! I'll definitely try porting the example code, in the mean time - feel free to disable it.

lujji:
@ourairquality could you point out where build checks fail? I get 3 warnings when I build the project - is that the problem?

ourairquality:
@lujji Thank you for taking a look. Problems looked worse at first, and glad it was an easy fix. Re-enabled the http_server example and the checks all pass now.

UncleRus:
Awesome job.
Will merge it if no objections.

malachib:
I am very excited about this. mdns is super convenient, and http(s)d included is the missing link.  Hopefully our devices have enough juice to to the https stuff..

malachib:
I am attempting to use the new lwip2 httpd server but experiencing linker issues:

```
/home/esp/esp-open-rtos/examples/project/src/PGFR-14/http_server.c:83: undefined reference to `http_set_cgi_handlers'
/home/esp/esp-open-rtos/examples/project/src/PGFR-14/http_server.c:196: undefined reference to `http_set_ssi_handler'
/home/esp/esp-open-rtos/examples/project/src/PGFR-14/http_server.c:202: undefined reference to `httpd_init'

```

I notice the lwip/src/apps are skipped during compilation in my environment, which would explain the linker errors.  Do I need to manually enable them similarly to an 'extra'?

@ourairquality thank you for your hard work

On a somewhat related note, I have adapted a docker image for easy use your fork of esp-open-rtos.  You can find it here: https://hub.docker.com/r/malachib/esp-open-rtos/ and use tag "latest-lwip2"

Lastly, if you'd rather I put in these requests other than in the official pull request thread, please point me towards your preferred location.

ourairquality:
@malachib Perhaps look at the current example/http_server and then either copy the httpd code into extras/ or add a lwip/lwip/src/apps/httpd/component.mk file and include it as an extra component from where it is. I was able to get it building quickly doing that and it shows an index page at least. I expect most people with an interest are following here, so I suggest keeping the discussion here.

malachib:
OK sounds great

I will start by chucking a component.mk file around.   Maybe we can do one for all of apps?  I'll see how it goes and put my results here.

malachib:
I think I found an elegant solution for this particular challenge.

The LWIP apps appear by design to default to off unless you explicitly enable them, via flags like `LWIP_HTTPD_SSI=1`, etc.

The result of this is the apps don't seem to collide with anything if you always include them in the build.

I've added the line:

`lwip_SRC_DIR += $(LWIP_DIR)apps/*`

to the /lwip/component.mk file and so far things are compiling nicely

ourairquality:
@malachib The `state` and `hwaddr` netif struct slots have been moved to the start before other optional code, so they are less likely to need to be moved again. Would be curious if you can get mdns working. Shall look in to the build change above that you suggest, to include the apps.

malachib:
Here's what I am fiddling with as we speak:

https://gist.github.com/malachib/82fb94734e0b85242231c3c99b7439b2

Things get interesting around line 237.  I did a minor amount of shuffling and brute forced on a few options ( i love callbacks anyway... so bonus there)

I didn't gut the ipv6. it's a gamble. we could brute force them OFF... make merges in the future easier...

ourairquality:
Added a comment in `wdev.c` on the apparently unused buffers, so there appears to be 7980 bytes free at `wDevCtrl + 0x21a4`. Assuming these were just some planned buffers that were never used, and not some workaround for a hw issue, then it should be usable and perhaps this can help push the networking a bit further.

malachib:
You definitely know more than I do about their internal hackeries.  That would make me nervous, but if you're sure about it light it up!  Maybe #ifdef it?

ourairquality:
@malachib There is no source code using this area. Only a comment has been added. I do think it is unused, but the sdk code in that area is very convoluted so I am not yet certain. I am not sure if such a chunk of memory can just be linked into the malloc poll, but I have not looked. Just wanted to point out that there might be a little extra spare ram.

ourairquality:
If you were having issues with a station with a static IP address then please give it another try - it never worked in the past for me either but it does now with some limited testing. We seem to have all the source code for that path now in wpa_main.c so it was possible to rewrite it, and the same code needed changing for lwip v2 anyway and has been cleaned up and a hack removed from the lwip code.

malachib:
In all my own testing, this lwip2 branch has performed admirably.  Additionally, it looks like this build failure may have been a fluke.  If a rebuild passes, I motion we proceed forward and merge in this pull request.

ourairquality:
Might be best to just wait a little longer. I am trying to get the same lwip v2 code running on the current SDK too and another pass over it might clean it up a little more.

malachib:
I respect that.  I've updated the docker images for a ready-to-go build environment.  They work great for me, but try them at your own risk :)  The one tagged "latest-ourairquality" points to your repo and branch of the same name

ourairquality:
Had some luck getting ipv6 running, and the patch at https://github.com/ourairquality/esp-open-rtos has ipv6 enabled. Currently only the http-get example uses ipv6, so there is some work to go cleaning up the code. Not sure how well ipv6 is tested in lwip, dns was broken, and neighbor solicitation was broken (or I did not know how to get it running as is), but these have been patched. Could use help on this, and perhaps having something barely working might give some encouragement.

SaimenSays:
I have some serious problems with actual lwip or low level hardware adaption. After some time tcp_accept() callback gets never called anymore.
Hopefully this is fixed with this update. Does anybody know when this pull request will be "official" released?

ourairquality:
@SaimenSays Sorry it's probably not ready yet, and the patch here is rather stale now. I think good progress has been made and issues are being found and fixed at a good rate, but there are still open issues such as the queuing of pbufs in the TX paths. The upstream lwip code is in general receiving fixes at a good rate, and a good number of issues have been in the upstream code, so perhaps it has a little way to go to the next stable release. If you want to help then please see the version at https://github.com/ourairquality/esp-open-rtos for now.

SaimenSays:
@ourairquality I checked out your branch. Some modifications because of changed function arguments need tobe done. Now it compiles and seems to work on first sight.

One issue I figured out already before, is the delayed freeing of heap. On each new connection there are 168Bytes more heap used, also if TCP connection is closed. It takes sometimes up to ones second until RAM gets released and I can see previous value with 'sdk_system_get_free_heap_size()'.
It looks little amount, but in one minute many connections can be opened and we know ESP lacks in RAM size.
Any idea how this can happen? Do you know if it is an issue of RTOS memory handling or is there a longer timeout in tcp task?

SaimenSays:
@ourairquality I checked out your branch. Some modifications because of changed function arguments need tobe done. Now it compiles and seems to work on first sight.

One issue I figured out already before, is the delayed freeing of heap. On each new connection there are 168Bytes more heap used, also if TCP connection is closed. It takes sometimes up to ones second until RAM gets released and I can see previous value with 'sdk_system_get_free_heap_size()'.
It looks little amount, but in one minute many connections can be opened and we know ESP lacks in RAM size.
Any idea how this happens? Does anybody know if it is an issue of RTOS memory handling or is there a longer timeout in tcp task?

ourairquality:
@SaimenSays It might be the TCP time-wait state. I had to rework the wificfg http server to try to have the client close the connection first to avoid the server being left in a time-wait, and that change helped a lot. Is that an option in your application, so avoid the esp8266 closing the TCP connection until after the client has? Alternatively perhaps you can lower this wait time. Perhaps try enabling LWIP_STATS_DISPLAY and calling stats_display() in lwip, see if it gives some clues.

ourairquality:
This patch has been updated to the latest code developed. Patch https://github.com/SuperHouse/esp-open-rtos/pull/430 had to be bundled in.

For practical reasons there will be no further development of this particular patch. I can't just keep rebasing as people using my fork need the history and having to deal with a rebased stack of patches is frustrating for them. So further work will build on this. It might still be possible to pick and bundle some follow up patches, but I expect it will become difficult due to conflicts.

This is as good as I have been able to get it in the time available.

malachib:
I appreciate all your efforts @ourairquality .  Your lwip2 code was running great for me

flannelhead:
@malachib These patches still exist and now reside in a different PR :) https://github.com/SuperHouse/esp-open-rtos/pull/437

PR #394, opened 2017-06-13
Add dirty port of irmp library.

hetii:
Hi :)

I wrote a port of irmp library that allow parse or send IR commands.
Here is a project site: https://www.mikrocontroller.net/articles/IRMP_-_english
Currently only receiving is implemented but I should add sending in day or two.

What I need to know is do we keep this library as much as its in original page or adopt it just for esp-open-rtos purpose?

General let me know if you want it and how and then I will try to prepare it. 

Best Regards.


kanflo:
The sources in this PR has an incompatible license (GPLv2) for this project.

PR #397, opened 2017-06-20
MQTT ignores too many handlers registered

Governa:
On MQTTClient.c function mqtt_subscribe, if more than MQTT_MAX_MESSAGE_HANDLERS are registered, the variable rc receives the value of grantedQoS.

If grantedQoS = 0, and all handlers are already occupied, it goes on and returns 0 to the caller.


UncleRus:
Thank you!

PR #401, opened 2017-07-03
Mqtt timer not sleeping

Governa:
Some timing functions of the ESP82668's MQTT client port seems to be using incorrect values. That was causing tasks to busy wait for a packet inside the cycle function, increasing power usage and not letting lower priority tasks execute.


UncleRus:
Thank you!

PR #403, opened 2017-07-04
Update mbed TLS to 2.5.1

ourairquality:
This is the latest stable version, so perhaps we should consider updating to this.


malachib:
Yes please!

PR #412, opened 2017-07-12
FreeRTOS 9.0.1 (dev)

ourairquality:
This is a development branch for now but it must be getting close. The patch has been derived from the new repo https://github.com/ourairquality/freertos which tracks the upstream code making it easier to keep up to date. That repo also has the sdk revisions in case that helps understand what might need to be looked into.

One feature in v9 that I had missed is integrated support for the new reent code. This is not currently used, but might be something to consider.


ourairquality:
Enabled the configUSE_NEWLIB_REENTRANT option in FreeRTOS. It works at least in part with the more recent newlib version I use. This would seem to be essential as we can't have multiple threads working on a common reent structure, sharing the same errno etc. Have not explored all the implications yet. One area to explore is the standard streams, and I immediately noticed the extra buffering with debug output not being flushed as usual.

ourairquality:
Current newlib adds the option --enable-newlib-global-stdio-streams which was reported to save memory and to address some other issues so might be something we want.

ourairquality:
Updated, and again the last of the work on this patch, updates will build on this.

PR #413, opened 2017-07-13
libc: update to a recent newlib version.

ourairquality:
Updates to the current newlib master version, which is 2.5.0 plus some other changes that we may well need. This is compiled from the source code at https://github.com/ourairquality/newlib from the esp-open-rtos branch, and there are instructions there for compiling from the source.


ourairquality:
I think the lwip code is breaking this and a small change is needed with a more recent newlib, but that is in a different repo, so the lwip v2 PR will probably need to land first.

PR #415, opened 2017-07-14
Remove code from libgcc that is provided by the ROM.

ourairquality:
This saves about 2k of iram. Perhaps there is another way, but this was necessary here. If someone has a nicer solution then please let us know. The imported libgcc.a was taken from a current build via the esp-open-sdk.


ourairquality:
Extended this to also remove a few of the libc functions that are in the rom, saving an extra 0.5k of iram. A libc/libc.remove file has been added to list the objects to be remove, rather than actually excluding them from the supplied newlib libc.

UncleRus:
2K of IRAM! That's great!

PR #416, opened 2017-07-20
Use hwrand for the sdk uses of rand, in particular with the NMI.

ourairquality:
The fiq NMI calls rand() from lmac:lmac.a:sdk_lmacTxFrame and the NMI must not touch the newlib reent structure or enter critical regions etc, so just use the simple and safe hwrand implementation as a
substitute.


kanflo:
Looks good to me. The Jenkins slave is missing pyserial :)

PR #418, opened 2017-07-20
Upnp example emulating a Wemo switch

apiel:
I would like to share an example to create Upnp server. In this example I emulate a WeMo switch plug (wifi switch) recognizable by Amazon echo dot. It was not really easy to find a way to use a udp multicast, until I found a great example https://github.com/kanflo/uhej-example

### DIFF
--- a/examples/upnp/Makefile
+++ b/examples/upnp/Makefile
@@ -1,0 +1,5 @@
+PROGRAM=upnp_test
+OTA=1
+EXTRA_COMPONENTS=extras/rboot-ota
+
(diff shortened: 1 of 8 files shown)


kanflo:
Great example, glad you found uHej of use. It's a pity travis is b0rken but since you are not changing anything in the common code I am giving this +1.

apiel:
Yes, I am wondering why it didn't pass the Travis test... by the way thanks for your udp multicast example, without it, I would had never manage to make this upnp example ;-)

PR #419, opened 2017-07-21
Include python3-serial

ourairquality:
Seem that including python-serial no longer works, so try python3-serial?


ourairquality:
Unfortunately this change has not resolved the problem.

PR #421, opened 2017-07-25
dhcpserver: optionally add router and DNS server fields to DHCP reply

flannelhead:
This change allows the user of DHCP server to advertise the soft AP as a router and a DNS server. This behaviour is useful when setting up a captive DNS server.

The options are disabled by default.

### DIFF
--- a/extras/dhcpserver/dhcpserver.c
+++ b/extras/dhcpserver/dhcpserver.c
@@ -14,17 +14,20 @@
  * BSD Licensed as described in the file LICENSE
  */
 #include <string.h>
+#include <strings.h>
 
 #include <FreeRTOS.h>
 #include <task.h>
 #include <lwip/netif.h>
 #include <lwip/api.h>
+#include "esplibs/libmain.h"
 
 /* Grow the size of the lwip dhcp_msg struct's options field, as LWIP
    defaults to a 68 octet options field for its DHCP client, and most
    full-sized clients send us more than this. */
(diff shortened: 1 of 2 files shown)


ourairquality:
fwiw the lwip v2 patches bundle in similar support which is used by wificfg, and there it was useful to have these be dynamic settings that could be changed at runtime rather than making them compile time options. The wificfg also has a very simple DNS that seems adequate - perhaps that could be split out as another alternative.

flannelhead:
@ourairquality, thanks for the heads-up! I wasn't aware of your own branch (for others reading this, it's [here](https://github.com/ourairquality/esp-open-rtos)). The `dhcpserver` changes there seem to be indeed more useful than this patch due to the runtime configurability. I'll be happy to close this one if/when these changes enter upstream.

The [DNS implementation](https://github.com/ourairquality/esp-open-rtos/blob/ourairquality/extras/wificfg/wificfg.c#L1545) also seems simple but sufficient. I'll try to use that in favour of my `captdns` patch and report back. Perhaps the DNS responder task could be even exposed as a separate module.

flannelhead:
Alright, so I backported the changes to `dhcpserver` from @ourairquality's fork - this allows configuring the router and DNS option fields at runtime.

flannelhead:
Closing and reopening to trigger rebuild.

flannelhead:
Closing, resolved by https://github.com/SuperHouse/esp-open-rtos/pull/437

PR #423, opened 2017-07-26
Set by default ESP8266 I2C pin for bmp280 example

apiel:
I propose to use the I2C pin settings of the ESP8266 since it is much more common than ESP8285.


kkornerz:
ESP8266 ESP-01 module has only GPIO0 and GPIO2 available, so current default values seem reasonable.

apiel:
Ok, I didn't know that. Thank you for the feedback. Actually I was assuming this by looking to the ESP-arduino since there are using GPIO4 and 5...

PR #426, opened 2017-07-29
DHCP Allows selecting the DHCP server's network interface

Governa:
Simple change allowing the selection of network interface on server start.


kanflo:
Looks good to me.

extras/dhcpserver/include/dhcpserver.h
```diff
@@ -26,7 +26,19 @@ extern "C" {
...
+
+/* Start DHCP server
+
+   See @dhcpserver_start_on_if.
+
+   Selects the first netif in netif_list as the server_if
+ */
+static inline void dhcpserver_start(const ip_addr_t *first_client_addr, uint8_t max_leases)
+{
+	dhcpserver_start_on_if(first_client_addr, max_leases, NULL);
```
ourairquality: I suggest just changing `dhcpserver_start` to take the `netif` and patching all the callers to pass in the appropriate netif which they can get using `sdk_system_get_netif`. Perhaps even place the `netif` as the first argument. That seems better than reaching into the lwip `netif_list` and taking the first element from that list and separated the dhcp server a little more from the internals of lwip.

Governa: Ok, I can do that. But is it ok to break the API like this?

UncleRus:

ourairquality:
Commit 5ee6c548 needs to be reverted.

PR #428, opened 2017-07-31
sysparam: reserve one more flash sector before placing the sysparams

ourairquality:
This is to help work with recent SDKs that add a RF cal sector by
default in the fifth last sector - just so the sysparam sectors do not
jump around when using different SDK versions.

### DIFF
--- a/core/app_main.c
+++ b/core/app_main.c
@@ -214,8 +214,8 @@ void IRAM sdk_user_start(void) {
     memcpy(&sdk_g_ic.s, buf32, sizeof(struct sdk_g_ic_saved_st));
 
     // By default, put the sysparam region just below the config sectors at the
-    // top of the flash space
-    sysparam_addr = flash_size - (4 + DEFAULT_SYSPARAM_SECTORS) * sdk_flashchip.sector_size;
+    // top of the flash space, and allowing one extra sector spare.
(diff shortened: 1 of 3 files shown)


kanflo:
Should there not be a check for sysparams in the fourth sector and if found, move it to the fifth, for backwards compatibility with current sysparam nodes?

ourairquality:
@kanflo People will just have to rework their flash layout. Take this patch as a heads up to think ahead now. For example, I had been using all the prior sectors for data storage, so losing a sector broke the end of that storage anyway.

The flash sectors can be saved and moved with esptool, or examples/sysparam_editor/ can be used to dump them before this change and set them again. 

I was not aware this was widely used, seems to still have problems, but if someone has significant deployments then they might just have to add another module to transition the sysparams (and any prior usage of the flash), or just keep a local back-out patch, but that could be an optional extra component.

Delaying the change is likely to make this a bigger problem in future if ever moving to a recent sdk. That seems to require flashing an extra params sector anyway, but if the sysparams do not have to move then that might be just a little easier.

ourairquality:
@kanflo Sorry it does appear that it's ok, as sysparam_init can search forward for the start of its sectors, and app_main gives the flash_size as the top address to search, so it appear it will work in transition. I still think this would be a useful change, so people know that the fifth last sector may be used in future and plan for it.

PR #430, opened 2017-08-02
upnp example: strip down the lwipopts.h file

ourairquality:
@apiel Would this change suit, to make is cleaner to transition to lwip v2? A patch to get your upnp example compiling with lwip v2 is also now bundled in that patch at https://github.com/ourairquality/esp-open-rtos and if you happen to be able to actually test that it would be good to know.


apiel:
At first look it seem to be no problem, but right now I don't have time to try out... I will try as soon as possible, hopefully this weekend and give you some feedback ;-)

apiel:
You need to define LWIP_RAND

```
#define LWIP_IGMP 1
#include <esp/hwrand.h>
#define LWIP_RAND hwrand

/* Use the defaults for everything else */
#include_next <lwipopts.h>
```

ourairquality:
@apiel Ok, thank you. Updated, but I've not included the hwrand change as it either seems misplaced and should be changed in lwip, or we need to also modify lwip to check if it is already defined to avoid compile warnings. Fwiw the lwip v2 patch switches over to hwrand anyway - it's also getting difficult to juggle all the patches, e.g. would this land first or at all?

Was there a good reason that upnp needed hwrand, apart from it being preferred? Was there some thread safety issue with the newlib rand reent struct etc?

PR #431, opened 2017-08-02
i2c: remove wip 400k define

ourairquality:
Looks like this was test wip code that should not have been included. Want to be able to set the i2c per-project rather than having to patch the driver.

@Zaltora What do you think about this change?

Did not check if some examples depended on a higher speed?

### DIFF
--- a/extras/i2c/i2c.h
+++ b/extras/i2c/i2c.h
@@ -36,8 +36,6 @@ extern "C" {
 #endif
 
 
-#define I2C_FREQUENCY_400K true  // for test WIP
-
 /*
  *  Some bit can be transmit slower.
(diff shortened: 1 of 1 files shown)


Zaltora:
I wanted later to add a simple way to select the frequency. I did not have time to do that.
I wanted to add an extra abstraction to create as the SPI, bus system and add the frequency configuration.
I'm not sure that all components work at low frequency but I'm sure some components do not work at high frequency.

resetnow:
@UncleRus should we merge this? It's essentially a line of code that ended up in the repo by accident. Afterwards, we can see how it goes with @Zaltora's PR

Issue #429, opened 2017-08-01
i2c + MMA7455L - Input/Output ERROR


AdahilMuniz:
Hi, I have a problem related to the i2c. My application needs to take acceleration from a MMA7455L. So, first, I need to configure the sensor writing in the register 0x16. But i'm receiving a I/O error. Can someone help me?

Module: NodeMCU V1.0 || ESP-12E
Sensor: MMA7455L

Code:

`#include "i2c/i2c.h"

#define SCL (0)
#define SDA (2)

void configure();

void user_init()
{
    gpio_enable(BT, GPIO_INPUT);
    uart_set_baud(0, 115200);
    printf("SDK version:%s\n\n", sdk_system_get_sdk_version());
    i2c_init(SCL, SDA);

    configure();
}

void configure(){

    uint8_t addr = 0x1D;
    const uint8_t reg = 0x16;
    const uint8_t cmd = 0x01;

    int error = i2c_slave_write(addr, &reg, &cmd, 1);
    if(error!=0){
        printf("ERROR\n");
        if(error == -EBUSY){
            printf("Device Busy!\n");
        }
        if(error == -EIO){
            printf("Input/Output Error!\n");
        }
    }

}`

andrewclink:
Unfortunately, I've had some sketchy behavior with I2c. Double and triple check your connections and pull-ups, and consider modifying the i2c bit delay to slow down communications.

A logic analyzer (or DSO or maybe even analog scope) would go a long ways toward debugging this.

resetnow:
I would also suggest changing pullups to 4.7kOhm and maybe trying a lower frequency (if you are using 400k)

AdahilMuniz:
@andrewclink, I used another code(Arduino) with the same connections that I have now and the applicantion worked. So, I think that the connections are right . I will try to change the delay, but how can I do this? Thank you for your help.

AdahilMuniz:
@resetnow how can I change the pullups and the frequency? Thank you for help!

resetnow:
>how can I change the pullups 

If you haven't already, you need to physically connect resistors to I2C lines (if your board doesn't have them already).

>how can I change the pullups and the frequency

Currently there is a line of code that ended up in the repo by accident. You can remove `#define I2C_FREQUENCY_400K` from i2c.h

AdahilMuniz:
[...]

PR #432, opened 2017-08-02
I2C bus upgrade

Zaltora:
I started to integrate a bus system for the i2c library.
It is in response to #429 and #431 .
With this upgrage, we can have several i2c bus and the desired frequency to each.

i dont get an esp8266 to testing. I changed the ds3231 sample to try compilation. All other libraries must be modified. I would do it later.
You will probably need to test and adjust the delay values to match the correct frequencies
`const static uint8_t i2c_freq_array[3][2] = { {100,20}, {10,1}, {6,1} } ;
`
Don't merge it now ^^

### DIFF
--- a/examples/ad770x/main.c
+++ b/examples/ad770x/main.c
@@ -8,7 +8,6 @@
 #include <esp/uart.h>
 #include <espressif/esp_common.h>
 #include <stdio.h>
-#include <i2c/i2c.h>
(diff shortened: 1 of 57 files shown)


Zaltora:
I have updated all the libraries using the I2C.
I can't test hardware but all example compile fine.

UncleRus:

Issue #151, opened 2016-06-27
implementation issue with LWIP and FreeRTOS


DutchFox2:
I'm encountering problems with LWIP in esp-open-rtos.
Easiest steps to reproduce:
- start with the http_get example from the esp-open-rtos examples directory,
- modify the URL, so it will download a somewhat larger file (100 kB should do),
- in the receive loop, do not print the full received data, but just a '.' or something,
- put a vTaskDelay within the receive loop (100-1000 msec)

This last step is crucial. It mimics an application where the received data can only be processed slowly (in my case this will be a slow serial link to an even slower secondary microcontroller). The TCP protocol should handle this gracefully, because the TCP Rx Window will/should only be updated when there is enough free memory available to allow for further reception.

Observed results: the download starts normally, but will hang after a while (ca 10 sec). The 'hang' will be short in the beginning, but subsequent hangs will take longer and longer, until the delays are getting so long that the webserver finally gives up and stops serving the rest of the download (we are talking _minutes_ already now).
In a Wireshark log, lots of requests for retransmits can be seen, which is an indication that a serious number of packets is getting lost in the ESP.

According to the [LWIP-user mailing list](https://lists.gnu.org/archive/html/lwip-users/2016-04/msg00019.html), the most likely cause is a faulty vendor implementation: threading issues, failing to fetch _all_ received packets after a NIC interrupt, etc.
I have tried to pinpoint this problem myself, but without success. It might even be possible that things are already going wrong in the closed-source libraries from Espressif.
Compared to the code I mentioned in the LWIP-user mailing list, I have already stripped down the situation to the almost unmodified http_get example (see above), which still displays the faulty behavior.

If I try an equally simple example from esp-open-**sdk** (also modified to include the delay in the receive loop), with exactly the same download URL, the download succeeds without problems! Same compiler toolchain, same webserver, same hardware. This leads me to the conclusion that the problem must be somewhere within esp-open-rtos.

It would be great if somebody could try to reproduce this issue.

projectgus:
Hi,

Thanks for reporting this. I can reproduce it, and I think I might have seen it ages ago when investigating #20. Back then I put it down to having to disable interrupts (the "delay" was caused by writing to flash), but it seems like that's not really a requirement to trigger this.

Playing around, a workaround seems to be shrinking the TCP receive window size. So if you add to your lwip/lwipopts.h:

```
#define TCP_WND (TCP_MSS * 2)
```

[...]

PR #437, opened 2017-08-14
Merge the ourairquality fork

ourairquality:
This is a combined patch set, which has been necessary due to the dependencies.


resetnow:
Since that's a lot of stuff to merge and it's possible it won't be merged at all, I think we should ask maintainers (@kanflo, @projectgus) on what conditions this can be commited to master. Perhaps we could collect feedback from 2-3 people using e-o-r and I assume @ourairquality has a number of devices running constantly.

I have a couple of devices running MQTT clients 24/7 and I can test this PR with them.

flannelhead:
@resetnow That's a good idea. I can volunteer as a tester and give some feedback. I've recently adopted these patches in [my robotics project](https://github.com/flannelhead/espway/tree/open-rtos) which uses quite intensive networking via websockets and also takes advantage of the RTOS features.

Right now I'm investigating an issue where the TCP thread apparently gets stuck somewhere, but that might be a bug in my code. I'll get back when I've got that issue resolved and give the patches some testing.

ourairquality:
@flannelhead Enabling the tWIP_TCPIP_CORE_LOCKING_INPUT is causing problems for some projects, e.g. mdns, so please try this with the appended patch.

I am still interested in feedback and consider improvements, but people in my community need the revision history so I need to stop all the rebasing and this patch set is probably the cleanest it will ever be.

kanflo:
I think there's a lot of improvements in this PR. Hopefully I will find the time this weekend to start reviewing and testing. Regarding LWIP2 support, I see you point that submodule to your own repo @ourairquality. I am afraid I have not kept up with the progress of LWIP2 but is there any reason not to use the SuperHouse master for LWIP2?

ourairquality:
@kanflo Thank you. It is not possible to point to the SuperHouse lwip repo because it has not changed in a long time. Also my lwip repo is a fork off the main lwip repo `git://git.savannah.nongnu.org/lwip.git` with just a few changes and I have been working to minimize that set of changes and work to get fixes into the upstream code. I suggest you fork off lwip and create a branch for esp-open-rtos specific changes as I have done, you could just cherry pick the esp-open-rtos branch.

My hope is that by closing the gap between the esp-open-rtos lwip and the upstream code that it will help bring these communities closer. So when we find problems they are discussed and pushed upstream, and that perhaps some people from the wider lwip community will consider using esp-open-rtos.

ourairquality:
In case it helps get other people interested in this, and helps get it across the line, **IPV6 can be enabled**, it is not the default and not well tested but simple examples work for me, just define `LWIP_IPV6` in your project.

resetnow:
If that helps, I think we could update SuperHouse LWIP2 fork and coordinate the merge across different projects.

ourairquality:
@resetnow If the SuperHouse master branch can be updated to the upstream lwip git repo then that might work, otherwise I suggest just opening a fresh repo to track the upstream lwip with a branch for the few remaining patches specific to esp-open-rtos. If the barrier to having people tracking the upstream code can be lowered then that would be good.

I also created a repo for newlib to update that, and one for FreeRTOS, but it seemed easier to just bundle in a specific version of these.

recursify:
@ourairquality - #151 has been open for a while, any thoughts on if this PR will help address it?  If so I'm happy to test it out.

ourairquality:
@recursify This is a big update to lwip so it may well help. If you could re-test then that would be good to know. Might the lwip issue https://savannah.nongnu.org/bugs/?50837 be related to this problem, it talks about a client downloading and with delays? With this patch set it should be much easier for you to cherry-pick that proposed change and to see if it helps.

flannelhead:
I've been testing these changes now for a while. Things seem to be working fine otherwise, but I'm having a problem with [httpd](https://github.com/SuperHouse/esp-open-rtos/tree/master/extras/httpd) WebSocket communication hanging up after a while. I'll try to come up with a minimal repro application during this week, so perhaps we'll get this fixed. I remember not having this issue though when I used the LwIP 2 PR code in late July, though.

It would be nice to get the WebSocket support in `httpd` introduced by [lujji](https://github.com/lujji/esp-httpd) merged into the upstream LwIP `httpd` code. I see that code has got some improvements in upstream so it would make sense not to maintain an outdated version in `esp-open-rtos`. That would probably be quite an effort but I'm willing to lend a helping hand if I can spare some time for that.

ourairquality:
@flannelhead @recursify The latest upstream lwip version has been merged and appears to include some improvements that would help avoid hung connections. If you could find an example to reproduce the problem then that would help.

flannelhead:
@ourairquality It seems my problem is related somehow to `TCP_QUEUE_OOSEQ`. It's enabled by default in this PR and it isn't enabled in current master. When use this patch and set `TCP_QUEUE_OOSEQ=0`, the hanging doesn't happen anymore... I wonder what should I do next to debug this further. For one, I'm not sure if it's something wrong with my code, `httpd` or LWIP.

ourairquality:
@flannelhead Thank you for narrowing this down. Could you please try adding the following to `esp_interface.c:ethernetif_input()` just before the call `netif->input(p, netif)`. It will make a copy of the rx pool buffer (which are limited) and free that pool buffer immediately, and if that fixes the problem then it is likely that that rx pool is becoming exhausted, otherwise it is more likely a problem in lwip and the logic is likely in tcp_in.c.

```
    {
        struct pbuf *q = pbuf_clone(PBUF_RAW, PBUF_RAM, p);
        pbuf_free(p);
        if (q == NULL) {
            return;
        }
        p = q;
    }
```

Making a copy might not be the best solution as it increase peak memory usage, and it might still oom if there oos buffers are not well bound. There is code in lwip to limit the size of the oos buffers but it appears to be per-pcb and perhaps your app is building them up across multiple pcbs and a global limit might need to be added too. But firstly try narrowing it down using the code above.

flannelhead:
Thanks for the suggestion. This addition indeed seems to fix the hangups. To the small extent I understand the issue, it is reasonable to expect that the buffer would be exhausted - the application is doing signal processing at 1 kHz at a higher priority than the TCP thread but should be also receiving WebSocket frames 10 Hz.

In my application there is only one pcb involved at a time.

Please tell me if I can provide any further useful information. I'm not yet very deep into the LWIP internals myself.

ourairquality:
@flannelhead There is the TCP_OOSEQ_MAX_PBUFS option, that defaults to just 2, that appear to be intended to avoid this problem per-pcb. Could you confirm further that the rx pool is becoming exhausted, perhaps by adding a counter for how many lwip holds and increasing this counter in `ethernetif_input()` and decreasing it in `pbuf.c: pbuf_free()` around the call to `sdk_system_pp_recycle_rx_pkt()` which returns these to the internal pool. If such a counter is going over 4 then there is a problem, and perhaps you can track it down in tcp_in.c.

ourairquality:
@flannelhead I can now reproduce this problem with oos buffers. One problem was that I made a mess of the lwipopts TCP_OOSEQ_MAX_BYTES and TCP_OOSEQ_MAX_PBUFS and these were not being defined, but even with those set it's still not running very smoothly. Let me explore the problem further.

flannelhead:
@ourairquality Ok, thanks for digging deeper into this. I did some experiments myself yesterday but didn't get very far. I'm still available for testing so let me know when there's something to test.

ourairquality:
@flannelhead If you are prepared to retest then could you please try the updated patch set. The management of TCP OOSEQ buffers has been improved, and I found that the TCP SACK support also helps and enabled that too. The OOSEQ support alone, without SACK, performed really badly and I could not find a fault there. Filed an upstream bug for feedback see https://savannah.nongnu.org/bugs/index.php?51925 Perhaps OOSEQ support will just be disabled by default, but with SACK it seem significantly better, so feedback is welcomed.

flannelhead:
Hi,

That looks like some great work to be honest! At least it seems now more reliable, but still I managed to get it to hang by e.g. reloading a couple times the page which opens the websocket connection. I should finish the minimal repro which I've been making though... Which kind of test are you currently using for development?

ourairquality:
@flannelhead Found some more suspect errors in lwip. It was dropping pcbs due to errors maintaining the list. Filed a buf upstream for feedback https://savannah.nongnu.org/bugs/index.php?51930 This probably caused a lost of memory and importantly any rx pool buffers associated with the pcb which would bring the rx path to a halt. It appears to have only occurred when running multiple interfaces.

Apart from that, I might need more feedback on your problem as I can not yet reproduce it now. I had adapted the http_get example to download a large file over and over. Some code has been added to `ethernetif_input()` to simulate packet loss which you might want to try, and that had reproduced the problem here but it seems to keep running now. Could you please try a more conservative target in `ooseq_max_pbufs()`, lower the target number of ooseq buffers from 3 to 2, and see it that helps? Could you also print the level of `pp_rx_pool_usage` in `ethernetif_input()` to see that it does not get too hight, not over 4?

flannelhead:
@ourairquality Yes, let me test it for a couple of days and I'll report back to you. Thank you for your work so far, I hope to be able to help.

lwip/esp_interface.c
```diff
@@ -152,30 +152,51 @@ void pp_recycle_rx_pbuf(struct pbuf *p)
...
+#define COPY_PP_RX_PBUFS 0
+
 /* Return the number of ooseq bytes that can be retained given the current
  * size 'n'. */
 size_t ooseq_max_bytes(size_t n)
 {
+#if COPY_PP_RX_PBUFS
     size_t free = xPortGetFreeHeapSize();
-    size_t target = (free - 30000) + n;
+    ssize_t target = ((ssize_t)free - 30000) + n;
```
resetnow: Out of curiosity, is 30000 just an approximate arbitrary number or is it tied to an allocation somewhere?

ourairquality:
@resetnow Thank you. The intention was to set a practical limit, just forget to change that back after testing the limit.

flannelhead:
@ourairquality I don't know what I was doing wrong before, but now it seems I don't get those hangs anymore. The websocket connection seems to persist now, and I didn't manage to hang it despite trying like I did before (refreshing the page on the mobile phone etc.) Page refresh seems to have sometimes a small delay though, probably a timer then frees some buffers.

I will still do the checks you suggested to see there indeed isn't anything fishy going on.

flannelhead:
After some more testing, it seems that reverting the two latest commits 893f715 and 730a0f4 the problem appears again and I can reproduce the hanging. So, it seems the hanging problem is finally fixed by these latest changes.

recursify:
Okay, took a while to upgrade to the latest LWIP / FreeRTOS, but I'm now getting frequent errors with `mbedtls_ssl_read` returning 0x004c (MBEDTLS_ERR_NET_RECV_FAILED).  Usually it fails after a few successful reads/writes over a secure websocket connection.  Any pointers on how to debug this further?

ourairquality:
@recursify I would take a look at the packets with wireshark to see if there are any clues there, see if there are any problems visible at the tcp/ip layer. Lwip also has some debug output that can be enabled, see lwipopts.h, and I would setup a test for which that log can be recorded. I don't use mbedtls and are not going to be working on problem reports at that layer.

flannelhead:
FWIW, I've been now testing this changeset for a good while. After the fixes to lwIP done by @ourairquality, it's been working very well in my project.

It seems to me that it would probably be a good idea to merge these changes soon - after all, there are bugs in lwIP 1.4.1 in master as well, and lwIP 2 in this branch would fix some of them. The code would also get more exposure after merging and any potential bugs could be found out. But of course, all of this is up to the maintainers :)

kanflo:
I have tested with my Esparducam project and so far I only see issues with read(...) as mentioned in #445. Eg. examples/terminal buffer the entire line. I will dig further into this.

ourairquality:
@kanflo There have been some changes to the newlib output buffering for the standard output streams and perhaps that is causing some confusion, and perhaps that path needs some attention. Newlib added an option to make these streams global rather than per-thread, but that option is not used, but it is there now in the update. Personally per-thread streams seems good, to limit the thread conflict a little (to the extent that they write to different buffers before flushing these). I don't think the actual output to the uart was ever thread safe and that needs some attention. Often these streams are used for debug dumps, and it is often important that this work when the image can be in a bad state, so a robust best effort solution for this thread safety issue might be best (rather than calling into FreeRTOS).

The newlib master branch has also moved to a 64 bit time_t and does anyone have thoughts on that?

kanflo:
Thanks for the pointer @ourairquality, a time saver! I will look into fixing the input buffering. Concurrency protection on the UART would be nice but can be left out for the time being, I think.

Moving time_t to 64 bit is +2 for me :)

ourairquality:
@kanflo The initialization of the streams is in newlib/libc/stdio/findfp.c The initialization of stdout has not actually changed, but previously the reent structure was not even being switched on task switches so all tasks were using the same stream, really broken. With that fixed the stdout line buffering becomes visible. I am still not aware of an input buffering problem.

Newlib has been update to the master with the 64 bit time_t, so at least it is the latest code to work on.

kosmonaffft:
Hello all...
Are there any chances that this patch will be integrated soon?
Or i should use ourairquality fork right now to test them?

UncleRus:
Going to merge this one if no objections

flannelhead:
One note: I just noticed `esp_interface.c` in this branch fails to build if `TCP_QUEUE_OOSEQ=0`. `TCP_QUEUE_OOSEQ` is now on by default but it would be good if the code would build without it.

PR #439, opened 2017-08-22
Fix SPI3 protocol with SSD1306

Zaltora:
SPI3 protocol with SSD1306 work now  ( test done)

### DIFF
--- a/extras/ssd1306/ssd1306.c
+++ b/extras/ssd1306/ssd1306.c
@@ -99,7 +99,7 @@ static int inline i2c_send(const ssd1306_t *dev, uint8_t reg, uint8_t* data, uin
 /* Issue a command to SSD1306 device
  * I2C proto format:
  * |S|Slave Address|W|ACK|0x00|Command|Ack|P|
- * 
+ *
  * in case of two-bytes command here will be Data byte
  * right after the command byte.
  */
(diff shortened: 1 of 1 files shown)


ourairquality:
Do you know what this fixed? It appears to just unroll calls to `spi_transfer` and `spi_transfer_8` to send one byte at a time but why would that have been expected to fix a problem? Someone else would be expect to look at that and optimize that change away. Were those functions broken, or is it a timing issue, or is it just unknown in which case wouldn't it have been better to leave the original code but disabled with the working substitutes and a code comment. And where is the code review before merging?

Zaltora:
SPI3 SSD1306 protocol problem is that you need send 1 control bit for each byte. This bit replace D/C pin.  Before this change, transfert was this :
1bit + 8 bits + 8bits + 8bits + ....   
The command bit is send with the full buffer.
That was totally wrong. SSD1306 didnt work at all in this config.
To send this bit every time, i need transfert 1 byte in SPI buffer.
Now it is:
1bit + 8 bits + 1bit + 8bits + ....

Sorry for miss some comment on it. A friend do some test with me. Before, the screen didn't work at all. Now it is ok.  SH1106 got the same problem but no misstake for this one.

PR #440, opened 2017-08-29
Add a hook for a user defined exception handler

flannelhead:
If the software crashes, it is essential in some applications like robotics to e.g. shut down motors for safety. This PR adds an API for the user to add their own exception handler.


Zaltora:
It is possible to add an example ? ( To let the user know about this new feature and how to use it )

PR #444, opened 2017-09-14
Driver for Sensirion SHT3x temperature and humidity sensor

gschorcht:
I have written a driver for Sensirion SHT3x temperature and humidity sensors connected via I2C and some examples how to use it. Please add it to the main repository.


Zaltora:
Nice library and very complete :). Few questions/Improvements.
1)
I see you use multiple function for debug :
```
#ifdef SHT3x_DEBUG
#define DEBUG_PRINTF0(s)       printf(s);
#define DEBUG_PRINTF1(s,a)     printf(s,a);
#define DEBUG_PRINTF2(s,a,b)   printf(s,a,b);
#define DEBUG_PRINTF3(s,a,b,c) printf(s,a,b,c);
#else
#define DEBUG_PRINTF0(s)
#define DEBUG_PRINTF1(s,a)
#define DEBUG_PRINTF2(s,a,b)
#define DEBUG_PRINTF3(s,a,b,c)
#endif
```
Can be better to use this ? 
```
#ifdef SHT3x_DEBUG
#define DEBUG_PRINTF(s, ...) printf("%s: " s "\n", "SHT3x", ## __VA_ARGS__)
#else
#define DEBUG_PRINTF(s, ...)
#endif
```
2) For each i2c transmission, you use 
taskENTER_CRITICAL(); and taskEXIT_CRITICAL();
I read somewhere with FreeRTOS that not good to block task for a long time.
I think you can remove them. If it was to prevent conflict with multiple task, I2C bus got a "busy system" that prevent that error.

3) In example with 2 sensor, you use GPIO16. This pin is slow.  you can change it for a better one if you want :)

Regards

gschorcht:
I changed the library as well as the example according to your suggestions.

gschorcht:
This version of the library works now and is a good starting point. I will improve error handling next days. Furthermore, I will extend functionality a bit, e.g., the possibility to switch on/off the built-in heater of the sensor.

gschorcht:
I closed the pull request for the moment because I'm thinking about a redesign of the drive before it is merged in the master branch. The redesign should make the driver interface clearer and easier to use. I'm not sure yet, a possible change could be to remove the implicit mean value computation during each measurement cycle and to make it the responsibility of the user, if this is required.

Since the driver was still not merged into the master branch there shouldn't be any compatibility problems.

I will readd the driver and open a new pull request once the redesign has been finished.

PR #449, opened 2017-09-30
Added support for RGBW NeoPixels

eboskma:
This adds support for RGBW NeoPixels with the SK6812 controller.


examples/ws2812_i2s/ws2812_i2s_colour_loop.c
```diff
@@ -69,15 +71,23 @@ static void demo(void *pvParameters)
...
 {
     uart_set_baud(0, 115200);
+    struct sdk_station_config config = {
+        .ssid = "Loading...",
+        .password = "morays59924_howitzer",
+    };
+
+    /* required to call wifi_set_opmode before station_set_config */
+    sdk_wifi_set_opmode(STATION_MODE);
+    sdk_wifi_station_set_config(&config);
```
UncleRus: I don't think that WiFi-related things are necessary here

UncleRus:
I believe that example should be left untouched (RGB mode by default)

examples/ws2812_i2s/ws2812_i2s_colour_loop.c
eboskma: Oops, that is not meant to be there...

eboskma:
I've updated the example to use RGB pixels.

ourairquality:
I see a few people that seem to have not base patches off the current master branch. This can be corrected with a sequence such as: `git checkout master` `git fetch upstream` `git merge upstream/master` `git push` `git checkout <mybanch>` `git rebase master` `git push -f <mybranch>`. If your situation is more complex and you need to move just a few patches from some working branch then look at `git rebase --onto master ...`

UncleRus:
Squashed & merged

PR #450, opened 2017-10-01
Driver for Bosch BME680 sensor added

gschorcht:
I have written a driver for the brand new and famous product from Bosch Sensortronic's, the BME680 sensor. It integrates temperature, pressure, humidity and gas sensors with high accuracy within a single unit. It is a very low-power device developed specifically for mobile applications and wearables where size and low power consumption are key requirements. 

Please add the library to the master branch of the repository.

p.s. Because I had to learn using github (I used CVS for more than 20 years :-( ) I made some mistakes, e.g., I did not create separate branches for my developments. So I uploaded SHT3x driver developed some weeks ago to the master branch from which I drew now a branch for uploading this BME680 driver. So I think uploaded files of the SHT3x driver that are part of an open pull request are probably also part of this pull request :-(


ourairquality:
@gschorcht Have you learnt about `git rebase -i master` yet? That makes it easy to squash multiple patches into just one, for example to clean it up to get it ready to submit. After that it will be necessary to do a force push `git push -f bme680-drive`. Also take a look a `git cherry-pick`, useful for moving some patches across to a new branch to split them and then `git rebase -i` can be used to removed them from the prior branch. You can use that to make a branch for the sht3x patches and get them off your master branch, then reset your master branch back to the upstream master.

gschorcht:
I closed the pull request for the moment because I'm thinking about a redesign of the drive before it is merged in the master branch. The redesign should make the driver interface clearer and easier to use. I'm not sure yet, a possible change could be to remove the implicit mean value computation during each measurement cycle and to make it the responsibility of the user, if this is required.

Since the driver was still not merged into the master branch there shouldn't be any compatibility problems.

I will readd the driver and open a new pull request once the redesign has been finished.

PR #451, opened 2017-10-02
Pull new stuffs from origin repo.

ccbruce0812:


extras/dhcpserver/dhcpserver.c
```diff
@@ -150,6 +173,8 @@ static void dhcpserver_task(void *pxParameter)
             printf("DHCP Server Error: Failed to receive DHCP packet. err=%d\r\n", err);
             continue;
         }
+		
+		taskENTER_CRITICAL();
```
ourairquality: Looks like too much is being done in the critical regions. Probably need to use a mutex instead.

lwip/include/arch/cc.h
```diff
@@ -101,6 +101,8 @@ typedef int sys_prot_t;
 #define LWIP_PLATFORM_HTONS(_n)  ((u16_t)((((_n) & 0xff) << 8) | (((_n) >> 8) & 0xff)))
 #define LWIP_PLATFORM_HTONL(_n)  ((u32_t)( (((_n) & 0xff) << 24) | (((_n) & 0xff00) << 8) | (((_n) >> 8)  & 0xff00) | (((_n) >> 24) & 0xff) ))
 
+#ifndef LWIP_RAND
 #define LWIP_RAND()                         hwrand()
+#endif
```
ourairquality: This seems fine, but is the patch to upstream lwip also really needed?

Out of curiosity is hwrand not suitable?

.gitmodules
```diff
@@ -1,6 +1,6 @@
 [submodule "lwip/lwip"]
 	path = lwip/lwip
-	url = https://github.com/ourairquality/lwip.git
+	url = https://github.com/ccbruce0812/lwip.git
```
ourairquality: Not sure about switching across to your own git repo just to get in a modification to upstream code - that is not workable, we all need to work together and try to get changes submitted upstream where possible. You have not even explained why the change is necessary?

kanflo:
Hi @ccbruce0812. When raising a pull request, please be descriptive of what you want to merge and keep one feature to each PR. There are some good stuff in this PR but also some that will never get merged, such as the change in lwip/lwip. To be nit picky, I also spotted some tab indentation.

PR #453, opened 2017-10-03
Driver for Sensirion SHT3x temperature and humidity sensor readded

gschorcht:
I have written a driver and some examples for Sensirion SHT3x temperature and humidity sensors connected via I2C.

This is a redesigned version of the driver. The driver was already added temporarily some weeks ago. By pull request #444 I asked for merge into the master branch. However, due to some difficulties with separate development branches and a redesign of the drivers interface, I deleted the files and closed pull request #444 before the merge happened.

Now, the interface redesign has been finished and I am asking again for a merge into the master.


extras/sht3x/sht3x.c
```diff
@@ -0,0 +1,385 @@
...
+
+  values.temperature_c = ((((data[0] * 256.0) + data[1]) * 175) / 65535.0) - 45;
+  values.temperature_f = ((((data[0] * 256.0) + data[1]) * 347) / 65535.0) - 49;
+  values.humidity      = ((((data[3] * 256.0) + data[4]) * 100) / 65535.0);
+  
+  sht3x_sensors[id].values = values;
+}
+
+
+static void sht3x_background_task (void *pvParameters)
```
ourairquality: Personally I think this is the wrong level of abstraction for a device driver, and would move the task into the example code. There are often more than one device on an i2c bus, and the tasks created here can not coordinate these so it so limited that it seems out of place here. Any substantive use will need a separate task anyway, so suggest just keeping the code needed by such tasks here and moving the rest into the examples. Then you could eliminate the callback and task notification and multiple sensor abstractions from the code here.

gschorcht: @ourairquality 
Many thanks for your review. Your suggestion would make the life much more easier for me since I would have to document much less. But, moving the code from the background task into a user task would mean that the user is completly responsible to implement the protocol and the timings of the sensor. From my point of view, the user task should not take care about this. I would expect from a driver that I just call a function to get the values without thinking about details like protocol timings and so on. 

The driver can be used exactly in that way. Let the background task doing measurements periodically and call simply function sht3x_get_values whenever you need the values. You have not to take care about anything else.

I don't understand exactly your concern with multiple I2C device at the same bus. As I understand, the i2c driver takes care about concurrency situations.

ourairquality: The i2c driver does not coordinate multiple tasks, so the caller is responsible, and as noted if this sht3x driver were to be used with any other devices on the bus this driver would not work anyway.

On the bmp180, I don't think that was a good abstraction either, it has the same problem, but it is layered. I don't use the higher level abstractions there either, but it has the `bmp180_measure` interface that does all the work and can be used.

For the timing, why not just block while measuring. The thread task does this anyway, so the polling measurement could be moved to a separate function, and the rest moved out.

gschorcht: Ok, at least the i2c driver returns an busy error if another i2c device is communicating. I could improve the access to the i2c bus in that way, that the driver handles the busy situation and waits for a certain time. I think, this could be solved.

Yes, bmp180 driver has a single function for measurements. But, it starts the measurement and then uses sdk_os_delay_us to wait up to 30 ms before it reads the sensor values. I don't think that it is a good idea to block the system for 30 ms. Event though vTaskDelay could be used in such a function, the driver has no idea what task is delayed. For example, if the call of this single measurement function comes from a software timer callback, the timer service task would be delayed for 30 ms, ..., a bad idea.

Exactly that was the motivation to think about the solution with background task that is doing the measurement without disturbing other tasks or the system itself.

I wrote a driver for BME680 in same way, you saw it already in pull request #450. This sensor has a much more complex timings. The measurement cycle can vary from some milliseconds up to 4.5 seconds.

Ok, one possibility would be to separate the low level and the high level functions into seperate layers, but leaving the user alone with complex measurement procedures, ... I'm not sure whether its a good idea.

ourairquality: I does not seem reliable enough to me to just depend on the i2c driver falling over if there are thread conflicts. It appears to be an attempt to use a protocol designed to handle conflict on the wire to manage thread conflicts at the driver level. I suspect there could be thread conflicts internally, such as access to the I/O registers etc that also need to be managed. I am not sure, but would not use code that did not coordinate thread access in the i2c driver. This might be done per-bus.

The delay in the bmp180 driver does not block the system, it can be pre-empted, and as you note vTaskDelay would have been a better choice there.

Your own code does not use a callback, it's all inline in the thread body, and that would be the usage. If people want to build higher level abstractions then I would leave that up to them in their code, and there are so many options for that.

I have not looked your bmp680 code yet, but the existing bmp280 code has the function `bmp280_read_fixed` that does all the work there too, no need for all the higher level baggage.

These are just my observations. Personally your driver would be unusable for me for the reasons noted, but perhaps it suits someone else.

gschorcht: @ourairquality 
What about an low level interface like this one? 
```
void user_task(void *pvParameters)
{
    sht3x_values_t* results;
    int32_t duration;

    while (1) 
    {
        duration = sht3x_start_measurement (sensor);

        vTaskDelay (duration/portTICK_PERIOD_MS);

        if ((results = sht3x_get_results (sensor)))
            ... use the values

        vTaskDelay ((1000 - duration) / portTICK_PERIOD_MS);
    }
}
```
Function **sht3x_start_measurement** triggers the sensor to start the measurement and returns an estimated duration. **sht3x_get_results** reads raw sensor data and returns computed sensor values. I think, this would be easy and controllable enough.

extras/sht3x/README.md
```diff
@@ -0,0 +1,256 @@
+# Driver for **SHT3x** digital **temperature and humidity sensor**
+
+This driver is written for usage with the ESP8266 and FreeRTOS using the I2C interface driver ([esp-open-rtos](https://github.com/SuperHouse/esp-open-rtos) and [esp-open-rtos-driver-i2c](https://github.com/kanflo/esp-open-rtos-driver-i2c)).
```
ourairquality: It does not appear necessary to reference the esp-open-rtos i2c driver source code like this, and this driver is not even using the kanflo version referenced.

examples/sht3x/README.md
```diff
@@ -0,0 +1,48 @@
+# SHT3x Driver Examples
+
+These examples references two addtional drivers [i2c](https://github.com/kanflo/esp-open-rtos-driver-i2c) and [sht3x](https://github.com/gschorcht/esp-open-rtos/extras/sht3x), which are provided in the `../../extras` folder.
+
```
ourairquality: These referenced seem more confusing than useful. You are welcome to include some acknowledgement, by if that is the purpose then I would just do that rather than referencing repos like this.

examples/sht3x/sht3x_two_sensors/sht3x_two_sensors.c
```diff
@@ -0,0 +1,137 @@
...
+    sht3x_init_driver();
+    
+    // Create sensors
+    sht3x_init_sensor (&sensor1, I2C_1_BUS, SHT3x_ADDR_1);
+    sht3x_init_sensor (&sensor2, I2C_2_BUS, SHT3x_ADDR_2);
+    
+    // Create a task for each sensor
+    xTaskCreate(user_task_sensor1, "user_task_sensor1", 256, NULL, 2, 0);
+    xTaskCreate(user_task_sensor2, "user_task_sensor2", 256, NULL, 2, 0);
+
```
ourairquality: This case looks ok as these sensors are on separate IO pins.

examples/sht3x/sht3x_one_sensor/sht3x_one_sensor.c
```diff
@@ -0,0 +1,133 @@
...
+
+    // Init SHT3x driver (parameter is the number of SHT3x sensors used)
+    sht3x_init_driver();
+    
+    // Create sensors
+    sht3x_init_sensor (&sensor, I2C_BUS, SHT3x_ADDR_1);
+    
+    // Create a task for each sensor
+    xTaskCreate(user_task, "user_task", 256, NULL, 2, 0);
+
```
ourairquality: This does not look like good advice because if there are multiple sensors then there will be multiple tasks banding on the same i2c bus without and synchronisation.

extras/sht3x/sht3x.c
```diff
@@ -0,0 +1,342 @@
...
+        dev->meas_first = true;
+        return SHT3x_MEASURE_DURATION[dev->repeatability];   // in ms
+    }
+
+    return -1;  // on error
+}
+
+
+// returns 0 when finished, -1 on error, remaining measurement time otherwise
+int32_t sht3x_is_measuring (sht3x_sensor_t* dev)
```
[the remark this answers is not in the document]
gschorcht: This is something I also found in bmp280 driver and is needed for the busy waiting approach.

ourairquality:
Could you please rebase this and squash it into a single commit because a lot has change, lot been deleted etc. So `git rebase -i master`, and then (except for the first commit) change the first letter to `s`, and the `git push -f ... sht3x-driver`. Back it up first, just in case, if new to this.

gschorcht:
@ourairquality 
The version I commited on Oct 3rd were that one which had the wrong abstraction level from your point of view. Based on your suggestions, I changed it completly and checked in these changes on Oct 5th. I had to delete unsuitable examples. This version is therefore obsolete.

The version from Oct 5th has been reviewed by you, many thanks again. The changes I checked in yesterday do not consider your suggestion you made in this review. I would like to discuss some of them a bit before I start to change again a lot. Some of them I do not agree completely.

My problem with git is, that I don't have git and a checked out version locally. I just use the web front end for uploads. I'm not sure how to make an existing local directory to an checked out version of a repository without getting conflicts and preserves my local changes.

extras/sht3x/sht3x.c
```diff
@@ -0,0 +1,342 @@
...
+      
+    // init sensor data structure
+    dev->bus  = bus;
+    dev->addr = addr;
+    dev->mode = single_shot;
+    dev->repeatability = high;
+    dev->meas_started = false;
+    dev->meas_start_time = 0;
+    
+    dev->active = true;
```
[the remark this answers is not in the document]
gschorcht: I have changed it as following: 

The function **_sht3x_init_sensor_** creates the data structure of type **_sht3x_sensor_t_** dynamically and returns the pointer to this data structure on success. In case of error, the function frees the data structure and returns NULL. That is, resources are only allocated when the sensor device is present.

examples/sht3x/README.md
gschorcht: Removed it.

extras/sht3x/sht3x.c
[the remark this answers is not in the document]
gschorcht: I left it in the code because it is not harmful.

examples/sht3x/sht3x_one_sensor/sht3x_one_sensor.c
gschorcht: I changed the comment so that the creation of a seperate task for each sensor is not an advice anymore. The user can decide how to deal with it.

extras/sht3x/sht3x.c
```diff
@@ -0,0 +1,342 @@
...
+         debug_dev ("periodic data available", __FUNCTION__, dev);
+         dev->meas_first = false;
+         return &dev->results;
+    }
+             
+    return NULL;
+}
+
+
+bool sht3x_stop_measurement (sht3x_sensor_t* dev)
```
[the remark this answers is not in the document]
gschorcht: Removed it.

extras/sht3x/sht3x.c
```diff
@@ -0,0 +1,342 @@
...
+    for (int i=0; i < len; i++)
+        printf("%02x ", data[i]);
+    printf("\n");
+#   endif
+
+    return true;
+}
+
+
+static bool sht3x_compute_values (sht3x_sensor_t* dev, uint8_t* data)
```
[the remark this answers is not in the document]
gschorcht: Made function **_sht3x_compute_values_** public and created a new public function **_sht3x_get_raw_data_**. The user can now use either function **_sht3x_get_raw_data_** or function **_sht3x_get_results_** to fetch the results.

Function **__sht3x_get_raw_data_** fetches only the raw sensor data in 16-decimal format, checks the CRC checksums and stores them in an byte array of type **_sht3x_raw_data_t_**. The user task then can use them directly or can call function **_sht3x_compute_values_** to compute floating point sensor values from them. 

Function **_sht3x_get_results_** combines function *sht3x_read_raw_data* and function 
 *sht3x_compute_values_* to get the latest sensor values. This is the preferred approach to get sensor values by the user task.

extras/sht3x/README.md
gschorcht: Removed it.

extras/sht3x/sht3x.c
```diff
@@ -0,0 +1,342 @@
...
+        elapsed_time = (system_time - start_time) / 1000;  // in ms
+
+    if (elapsed_time >= SHT3x_MEASURE_DURATION[dev->repeatability])
+        return 0;
+    else
+        return SHT3x_MEASURE_DURATION[dev->repeatability] - elapsed_time;
+}
+
+
+sht3x_values_t* sht3x_get_results (sht3x_sensor_t* dev)
```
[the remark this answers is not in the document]
gschorcht: The my comment for the new function **_sht3x_get_raw_data_**.

gschorcht:
@ourairquality 
I have changed the driver interface according to most of your suggestions. Furthermore, I rebased the master as  asked for. I hope that everything is ok now for merging the driver into the master.

ourairquality:
Thank you for the changes. It looks usable in my own app now, so that makes me happy. I have some sht3x on hand and hope to find time to test this soon. There are some white space issues, lines end in white space, a small nit that could be cleaned up later.

extras/sht3x/sht3x.c
```diff
@@ -0,0 +1,443 @@
...
+    if (!dev->meas_first)
+        return 0;
+        
+    uint32_t start_time  = dev->meas_start_time ;     // in us
+	uint32_t system_time = sdk_system_get_time();   // in us
+	uint32_t elapsed_time;
+
+    if (system_time < start_time)
+        // in case of timer overflow
+        elapsed_time = (UINT32_MAX - start_time + system_time) / 1000;  // in ms
```
ourairquality: This special handling for the timer overflow is not necessary. Simply subtracting two uint32 values will give the difference even if the timer has wrapped.

extras/sht3x/sht3x.c
```diff
@@ -0,0 +1,443 @@
...
+    debug_dev ("status=%02x", __FUNCTION__, dev, *status);
+    return true;
+}
+
+
+
+static uint8_t* crc8_table;             // lookup table with precomputed crc values
+static bool crc8_first_time = true;     // indicator whether table has still to be created
+
+static void generate_crc8_table()
```
ourairquality: Could using RAM here for a table avoided here, there is not much! So either put the table in flash, or code a non-table version of crc8.

extras/sht3x/sht3x.c
gschorcht: Special handling removed.

extras/sht3x/sht3x.c
gschorcht: Made lookup table static to be held in flash

extras/sht3x/sht3x.c
```diff
@@ -0,0 +1,419 @@
...
+#include "espressif/esp_common.h"
+#include "espressif/sdk_private.h"
+
+#define SHT3x_STATUS_CMD               0xF32D
+#define SHT3x_CLEAR_STATUS_CMD         0x3041
+#define SHT3x_RESET_CMD                0x30A2
+#define SHT3x_FETCH_DATA_CMD           0xE000
+#define SHT3x_HEATER_OFF_CMD           0x3066
+
+uint16_t SHT3x_MEASURE_CMD[6][3] = { {0x2c06,0x2c0d,0x2c10},    // [SINGLE_SHOT][H,M,L]
```
ourairquality: Could make this a `const uint16_t ...` so it is placed in flash.

extras/sht3x/sht3x.c
```diff
@@ -0,0 +1,419 @@
...
+
+    *status = data[0] << 8 | data[1];
+    debug_dev ("status=%02x", __FUNCTION__, dev, *status);
+    return true;
+}
+
+
+
+static uint8_t crc8_table[256];         // lookup table with precomputed crc values
+static bool crc8_first_time = true;     // indicator whether table has still to be created
```
ourairquality: Would need to make this `const` to have it placed in flash. I suggest using a non-table implementation. A quick look suggests the following might work, or be close:

```
static uint8_t sht3x_crc8(uint8_t data[], uint8_t num_bytes)
{
    uint8_t crc = 0xff;
    uint8_t i;

    for (i = 0; i < num_bytes; i++) {
        crc ^= data[i];
        for (uint8_t bit = 8; bit > 0; bit++) {
            if (crc & 0x80)
                crc = (crc << 1) ^ 0x31;
            else
                crc = (crc << 1);
        }
    }
    return crc;
}
```

extras/sht3x/sht3x.c
```diff
@@ -0,0 +1,419 @@
...
+    }
+    
+    // computation is necessary in periodic mode only for first measurement
+    // in single shot mode every measurment is the first one
+    if (!dev->meas_first)
+        return 0;
+        
+    uint32_t elapsed_time;
+
+    elapsed_time = (sdk_system_get_time() - dev->meas_start_time) / 1000;  // in ms
```
ourairquality: Just a thought, but might `xTaskGetTickCount()` be better used for timing here. The measurement durations might then be in ticks too, and then the integer division could be avoided. Also the number of ticks to wait would be returned and could be used by the caller to pass directly to `vTaskDelay()` without another division.

gschorcht: Good idea, makes really sense.

extras/sht3x/sht3x.c
gschorcht: Yes, implementation is almost the same. I used the lookup table version because it is less time consuming. Ok, if it is used only one time per second or less, computation time doesn't matter.

I will change it.

extras/sht3x/sht3x.c
gschorcht: Ok, I will change. Obviously, your fighting for each byte of RAM ;-)

extras/sht3x/sht3x.c
```diff
@@ -0,0 +1,403 @@
...
+                                           {0x2737,0x2721,0x272a} };  // [PERIODIC_05][H,M,L]
+
+#define max(a,b)    (a) > (b) ? a : b
+
+// tick counts [High, Medium, Low]
+// minimum is one tick if portTICK_PERIOD_MS is greater than 20 or 30 ms
+const int32_t SHT3x_MEASURE_DURATION[3] = { max(30/portTICK_PERIOD_MS,1),
+                                            max(20/portTICK_PERIOD_MS,1),
+                                            max(20/portTICK_PERIOD_MS,1) };
+
```
ourairquality: This might still round down to a period that is too short. What about the following and then remove the `max` macro.

```
#define TIME_TO_TICK_CEILING(ms) (((ms) + (portTICK_PERIOD_MS - 1)) / portTICK_PERIOD_MS)
```

ourairquality: Sorry, the rtos tick period can be short in rare cases, on a cpu freq change it can halve for one tick, so perhaps this should be:

```
#define TIME_TO_TICKS_CEILING(ms) (((ms) + (portTICK_PERIOD_MS / 2) + (portTICK_PERIOD_MS - 1)) / portTICK_PERIOD_MS)
```

gschorcht: Hm, quite complicated.

I had supposed that your first formula should work well because it simply moves all values that are not divisible by portTICK_PERIOD_MS to a value that is higher than the next divisible value. For example, for a portTICK_PERIOD_MS of 10 ms, a duration of 32 ms is shifted to 41 ms and results in 4 ticks, while 30 ms are shifted to 39 ms and 3 ticks are obtained.

Their second formula leads to tick values that are too high for half of the values in normal cases. For example, for a portTICK_PERIOD_MS of 10 ms, a duration of 30 ms gives 44 ms and results in 4 ticks.

Since the durations given in the lookup table are too high, I would use your first formula. For example, according to the data sheet, a measurement with high repeatability (first value in the table) takes 13.5 ms. When I used 20 ms as the next multiple of portTICK_PERIOD_MS, I observed sometimes errors because the resulting waiting time was too short.

extras/sht3x/sht3x.c
```diff
@@ -0,0 +1,401 @@
...
+                                           {0x2236,0x2220,0x222b},    // [PERIODIC_2 ][H,M,L]
+                                           {0x2234,0x2322,0x2329},    // [PERIODIC_4 ][H,M,L]
+                                           {0x2737,0x2721,0x272a} };  // [PERIODIC_10][H,M,L]
+
+// tick counts [High, Medium, Low]
+// minimum is one tick if portTICK_PERIOD_MS is greater than 20 or 30 ms
+const int32_t SHT3x_MEASURE_DURATION[3] = { (30 + portTICK_PERIOD_MS-1)/portTICK_PERIOD_MS,
+                                            (20 + portTICK_PERIOD_MS-1)/portTICK_PERIOD_MS,
+                                            (20 + portTICK_PERIOD_MS-1)/portTICK_PERIOD_MS };
+
```
ourairquality: Sorry to be dithering on this. Trying to work it out for myself. Tried `vTaskDelay(1)` and on a small set get times from 1531us to 9014us, and `vTaskDelay(2)` get times from 11873us to 19343us.

The datasheet has the timings 4ms, 6ms, and 15ms. Might it be best to actually use those values in the code. Looking at the test results about for a 2 tick delay it is clear that this is not adequate to ensure a 15ms delay, so it is not a surprise that was failing.

The first tick must just run to the next tick, which might be much shorter than a full tick, and in this case it should probably be assumed it could be zero time, so add one tick period for that (not less one milli-second).

Also in rare cases a tick period might be halved, as the clock might run at twice the expected speed due to a cpu freq change, so to be sure add half a tick period extra.

```
#define TIME_TO_TICKS(ms) (((ms) + portTICK_PERIOD_MS + portTICK_PERIOD_MS / 2) / portTICK_PERIOD_MS)
```

Could you use a macro to avoid duplication, and the times from the datasheet:

```
const int32_t SHT3x_MEASURE_DURATION[3] = { TIME_TO_TICKS(15), TIME_TO_TICKS(6), TIME_TO_TICKS(4) };
```

gschorcht: Thank you for all your effort. Tried your formula and observed the following. For high repeatability (15 ms duration), I got the following results:

Duration=3 ticks
mean(vTaskDelay(3)) = 28.145160
min(vTaskDelay(3)) = 21.272000

Seems not to be a problem.

But for low repeatability (4 ms duration), I got:

Duration = 1 tick      (ok, according to your fomula it results from 19 ms / 10 ms)
mean(vTaskDelay) = 8.214056
min(vTaskDelay) = 2.735000 :-(

To be sure, that waiting time is long enough, evene if the shortest possible tick happens, the tick count should be 2 for low and medium repeatability and 3 for high repeatability.

ourairquality: Sorry, try again.

```
#define TIME_TO_TICKS(ms) (1 + ((ms) + portTICK_PERIOD_MS / 2 + (portTICK_PERIOD_MS - 1)) / portTICK_PERIOD_MS)
```

This gives 2 ticks for 4ms, 3 ticks for 6ms, and 3 ticks for 15ms. The extra tick might be needed for 6ms because it might be unlucky and one of the ticks might be halved to 5ms, so at the extreme it might wait only 5ms. What do you think?

gschorcht: I will try :-)

gschorcht: What about the following approach? 

Function **_sht3x_start_measurement_** doesn't return the number of of ticks required to wait for the results. Instead, the user task just calls a new function for waiting, e.g., **_sht3x_wait_for_data_** like following:

```
sht3x_start_measurement (sensor, single_shot);
sht3x_wait_for_data (sensor);        
sht3x_get_results (sensor, &values);
```

Function **_sht3x_wait_for_data_** clould deal with ticks that were too short and could look like following or similar:
```
int32_t SHT3x_MEASURE_DURATION[3] = { 15, 6, 4 };
...
bool sht3x_wait_for_data (sht3x_sensor_t* dev)
{
    ... // some error handling

    vTaskDelay((SHT3x_MEASURE_DURATION[dev->repeatability] + portTICK_PERIOD_MS-1) / portTICK_PERIOD_MS);

    while (sdk_system_get_time()/1000 - dev->meas_start_time < SHT3x_MEASURE_DURATION[dev->repeatability])
        vTaskDelay (1);
    ...
}
```
The lookup table contains the values in ms. Function **_sht3x_wait_for_data_** computes the tick number required and uses **_vTaskDelay_** for waiting. If the ticks were too short, it simply waits one further tick.

examples/sht3x/README.md
kanflo: @gschorcht, you just experienced one of those somewhat frustrating open source moments when one tries to follow practice already in place only to find that it made people realise the original practice was not a good one. Happens to all of us ;)

extras/sht3x/sht3x.c
gschorcht: @ourairquality
What is your opinion on the above approach of using a specialized waiting function?

Another possible approach would be to return the ticks by the **sht3x_start_measurement** function, but to give **sht3x_get_results** an additional boolean parameter which allows the user to tell the function to wait for the results. With that approach the user would have the choice either to realize waiting by himself or to invoke **sht3x_get_results** immediately after **sht3x_start_measurement** and to let the function **sht3x_get_results** wait:
```
sht3x_start_measurement (sensor, single_shot);
sht3x_get_results (sensor, &values, true);   // third parameter tells the function to wait for the results
```
If function **sht3x_get_results** should not wait and results are not available, it fails.

ourairquality: The time to complete a measurement is a constant based on the configuration so it does seem better to not return it from `sht3x_start_measurement` every time. There seem no real need for `sht3x_start_measurement` to handle any already running measurement, that seems a coding error.

Some callers might be able to pre-compute a schedule taking advantage of the constant measurement period so exposing this seems better in general. Some callers might still want to know the measurement time and to call back themselves so perhaps add a function to return the measurement period that is only valid after initialization.

Most callers probably do not care if they wait 3 ticks ot 6 ticks. Might be over optimizing here. I would just bump up the number of ticks to be always safe rather than defining a function to wait more accurately. If we want to add such a wait function then it probably has more general utility.

I would remove `sht3x_is_measuring`. It seems too overloaded. The caller should manage this.

I would remove `sht3x_get_raw_data`. It also seems too overload, handling single shot and periodic. In single shot mode it is trivial for the caller to call `sht3x_read_data` and then call a function to check the crc. For the periodic mode perhaps add back a function to fetch and read the data.

I would remove `sht3x_get_results`. It also seems too overload, and it does not help when the raw values are needed. I would not overload it even more by having it optionally wait.

If after the above clean up it seems useful to add back some higher level functions that combine these then that might be useful but I would keep it minimal.

These are just suggestions. I think the code as-is could be merged at least as a starting point.

gschorcht: @ourairquality 
Here are the results of tests with your last suggested ms2ticks macro

repeatability | ticks | required | min | mean | max | read failures
-------------- | ------ | -----  | ----- | ------- | ------ | ----------------------
low | 2 | 4 ms | 13.65 ms | 18.99 ms | 23.86 ms | 0
medium | 3 | 6 ms | 24.35 ms | 28.99 ms | 33.97 ms | 0
high | 3 | 15 ms | 24.35 ms | 28.99 ms | 33.97 ms | 0

3 ticks are not necessary for medium repeatability. I think, the following macro should do the best
```
#define TIME_TO_TICKS(ms) (1 + ((ms) + (portTICK_PERIOD_MS - 1)) / portTICK_PERIOD_MS)
```
It maps the ms value to the next value divisible by portTICK_PERIOD_MS and adds one additional tick for security. That is, one tick could be 0 ms in worst case. It results in 2 ticks for 4 ms an 6 ms (which seems to be enough) and 3 ticks for 15 ms. The tests results were:

repeatability | ticks | required | min | mean | max | read failures
-------------- | ------ | -----  | ----- | ------- | ------ | ----------------------
low | 2 | 4 ms | 13.65 ms | 18.99 ms | 23.86 ms | 0
medium | 2 | 6 ms | 13.65 ms | 18.99 ms | 23.86 ms | 0
high | 3 | 15 ms | 24.35 ms | 28.99 ms | 33.97 ms | 0

ourairquality: Have a look at `sdk__xt_timer_int()` in `open_esplibs/libmain/os_cpu_a.c`. This is the rtos tick handler. Note it adjusts the next trigger count based on the current cpu freq. If the cpu freq changes soon after then the tick period can be longer or shorter than expected. These cpu freq changes are not frequent, perhaps just when computing some hashes etc. This was why I suggested accounting for one tick being halved in period, so the 'medium' with only two ticks might be `0 + 5` ms and thus less than 6. Perhaps this is all being over optimized, does any app really care if it waits an extra tick to be safe?

gschorcht: @ourairquality 
Many thanks for your comprehensive review and your suggestions. I will think about some changes.

Concerning the ticks you are probably right that it doesn't matter for most applications whether they wait 2 or 3 ticks. Just for the sake of interest, is it really possible that a tick will take only 0 ms? I would expect that even if the CPU frequency increases, the previously calculated tick trigger count should not lead to a tick of 0 ms.

gschorcht: @ourairquality 
I am really very greatful for your suggestions and like to discuss them with you.

First I would like to explain a bit why you probably feel that some of the functions are overloaded. Most of your points result from the fact that I have learned all the time
- always realize interfaces that make them as easy as possible to use,
- always hide any details and just provide the user only with the results he is interested on
- always program fault tolerant, and
- never make any assumption that the user would be careful.

That is probably the reason, why I like abstraction and fault tolerance which leads of course to some overloading. I have played a bit with the sensor to see how fault tolerant it is against wrong timing. 

On function ```sht3x_start_measurement```: It is possible to remove the check whether a measurement is already started. The sensor doesn't react on multiple start measurement commands. Concerning the duration in ticks as return value, your are principally right. Even though it is a constant, I would still leave it as it is since the user task can use it but it isn't a must. For that users that don't want to implement a fixed schedule and need some timing information it is very useful.

On function ```sht3x_is_measuring```: I agree, it can be removed. In that case the user would have always to user ```vTaskDelay``` to wait which shouldn't be a problem. However, I would like to keep it for internal use since functions like ```sht3x_get_raw_data``` or ```sht3_get_results``` the needs to check whether measurement is still running. The trial to read out sensor data before they are available lead to an i2c communication error and the sensor blocks completely until power down. Unfortunately, the sensor has no possibility to check whether data are ready to read other than clock stretching that does not work correctly in current i2c version. Thus, this check has to be realized by timers.

On ```sht3x_get_raw_data```: I would like to leave this function as is. Reading the data and checking the CRC is identical for single shot and periodic mode. The only difference is that in periodic mode a fetching command has to be sent to the sensor before data are read. But, this is a technical detail that should be hidden by the function. Therefore, having only one function for both modes that provide exactly the same results makes sense from my point of view. Furthermore, I would like to remove the CRC data from the raw data provided to the user. I don't see any reason why the user task should need the CRCs. These are only to secure the communication on the i2c interface.

On ```sht3x_get_results```: This function is the higher level function you ask for. It combines reading the raw data and converting them into real values. Or do you mean that there should be one single function that comprises starting the measurement, waiting, reading raw data, and converting them to real values?

ourairquality: @gschorcht We obviously see things differently. I don't think that is a problem. We would probably write different drivers that would appeal to different people. These are all 'extras' and there is room for different approaches. So I would encourage you to do it as you see it meeting some use cases, and do it true to that, and well.

Regarding the near zero period tick, this occurs because the time from the start of the delay to the next tick defines the first tick period and the next tick might occur just after requesting a delay in which case that first tick period becomes almost zero.

I might have even defined the wait times as run time constants.

gschorcht:
@ourairquality 
Finally, I have squashed my last changes based on some of your suggestions and all small changes before into a single commit. Now, I would freeze the code for merging. I hope, you can live with it, it seems to be a compromise of your suggestions and my original code.

extras/sht3x/sht3x.c
```diff
@@ -0,0 +1,430 @@
...
+    dev->meas_start_time = sdk_system_get_time ();
+    dev->meas_started = true;
+    dev->meas_first = true;
+
+    return true;
+}
+
+
+uint8_t sht3x_get_measurement_duration (sht3x_sensor_t* dev)
```
ourairquality: I would just pass in the 'repeatability' argument. That way it is clear to the caller that this depends only on the 'repeatability' argument and is a constant that they can hoist and that they need not lock the i2c bus for this call.

extras/sht3x/sht3x.c
```diff
@@ -0,0 +1,430 @@
...
+        dev->error_code |= SHT3x_WRONG_CRC_HUMIDITY;
+        return false;
+    }
+
+    return true;
+}
+
+
+bool sht3x_compute_values (sht3x_raw_data_t raw_data, sht3x_values_t* values)
+{
```
ourairquality: Could just pass in two pointers to floats and remote sht3x_values_t.

extras/sht3x/sht3x.h
```diff
@@ -0,0 +1,250 @@
...
+} sht3x_mode_t;
+    
+    
+/**
+ * @brief   possible repeatability modes
+ */
+typedef enum {
+    high = 0,
+    medium,
+    low
```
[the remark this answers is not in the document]
gschorcht: Good idea.

extras/sht3x/sht3x.c
```diff
@@ -0,0 +1,430 @@
...
+        free(dev);
+        return NULL;       
+    }
+    
+    debug_dev ("sensor initialized", __FUNCTION__, dev);
+    return dev;
+}
+
+
+bool sht3x_start_measurement (sht3x_sensor_t* dev, sht3x_mode_t mode)
```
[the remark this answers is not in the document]
gschorcht: @ourairquality 
Good idea, and indeed I already thought about it before. But I wasn't sure, whether introducing repeatability parameter wouldn't confuse most users. The user would have to specify the parameter every time (and of course I would have to document it :-/ ) and probably most users would ask what the hell this parameter means or for which purpose they have to specify it every time?

The best measurement results are provided at high repeatability. Probably, this is what most users want and therefore seems to be a good default option The only disadvantage of high repeatability is that the duration is one tick longer than for low repeatability and of course a higher power consumption. But I guess, it is the same as with the discussion whether it is a problem to wait one tick more or less to wait long enough. 

So, I decided by design to have high repeatability as default and not to confuse the user. Experts could set the repeatability member before starting the measurement. But, I am really not sure.

Would it be ok for you, to leave it as it is? Just for comfort.

extras/sht3x/sht3x.c
gschorcht: @ourairquality 
I understand your intention. Unfortunately, my understanding of this function was more like "Give me the duration for the current measurement configuration of the sensor" and not "What would the duration be if the sensor would measure with a certain configuration". Exactly that was the reason why function ```sht3x_start_measurement``` initially returned the duration.

I know that you just only try to optimize everything that is constant. But, the function call costs almost nothing. It is just the call and a table lookup. Indeed it could be realized as macro.

Furthermore, this question is related to the question whether to have a repeatability parameter in ```sht3x_start_measurement``` or not. If ```sht3x_start_measurement``` would have this parameter, the user already knows something about repeatability and it would be ok. Without this parameter, the function would be confusing.

At the moment it is a bit frustrating to change again and again the interface and to change the documentation every time. Latter part is the most time consuming part. Therefore I would like to come to an interface code freeze. The code behind the interface can changed anytime later, but the interface must never change.

extras/sht3x/sht3x.h
```diff
@@ -0,0 +1,250 @@
...
+ */
+typedef uint8_t sht3x_raw_data_t [SHT3x_RAW_DATA_SIZE];
+
+/**
+ * @brief	sensor values type
+ */
+typedef struct {
+    float   temperature;    // temperature in degree Celcius
+    float   humidity;       // humidity in percent
+} sht3x_values_t;
```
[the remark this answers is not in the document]
gschorcht: @ourairquality See above.

extras/sht3x/sht3x.c
gschorcht: @ourairquality 
What if the user gets the duration for low repeatability and then starts a measurement at high repeatability? I know, this would be a coding error. But the driver is not fault tolerant. Therefore, my idea was that the user gets always the duration the sensor need for the current configuration.

gschorcht: @ourairquality 
In my first approach I didn't plan to give the user the chance to select the repeatability. I planned to perform measurement always with high repeatability.

ourairquality: I don't mind if the driver lands as-is. I don't want to impose my views on this design and that is not necessary when it is an extra component that people can optionally use. We probably just have different views on coding, my preference is minimal abstraction and for run time checks to be a compile time option.

gschorcht:
I changed all things according to your suggestions even though I am not really lucky with all of them. I hope it's ok now for merge.

ourairquality:
This is not my project and I do not decide what lands or not. I took a particular interest in this driver because I will be supporting the device. I had been trying to communicate that, that you should design it as you wish. Sorry I don't want to be seen as blocking such work and do not want to comment further on this.

gschorcht:
@ourairquality 
I am really sorry that your are feeling to be the blocking, this wasn't my intention. I would like to thank you again for all the review work and your suggestions. I'm really very grateful. I appreciated it to discussed the different approaches with you. For me it is always very important to get a view from another perspective than my own. And I think the driver is now in a good state were it can be used by different users.

I hope that the driver as it is also somewhat usefull for you even though you would have designed it slightly different.

As you know, I like abstraction and to keep the user away from details. Therefore, I added one single function ```sht3x_measure``` that does the complete measurement in single shot mode with high repeatability. I believe that is the most easiest way to use the driver for users that don't want to know to many details how the sensor works.

kanflo:
While I have no sensor to test with I'd like to thank both of you @gschorcht and @ourairquality for your hard work on this PR. I am impressed with the level of documentation on this "extras"

Issue #445, opened 2017-09-18
read(...) and write(...) no longer works with LWIP sockets


kanflo:
I noticed read(...) and write(...) no longer can operate on LWIP sockets. Does anyone know if is this intended or accidental? It used to work about a year ago with the following snippet working as a web server. Today write(...) sends the HTTP response to stdout (the uart) and read(...) returns nothing (probably as the part is idle).

```
struct sockaddr_in server_addr, client_addr;
int server_sock, client_sock;
socklen_t sin_size;
bzero(&server_addr, sizeof(struct sockaddr_in));
server_addr.sin_family = AF_INET;
server_addr.sin_addr.s_addr = INADDR_ANY;
server_addr.sin_port = htons(80);

server_sock = socket(AF_INET, SOCK_STREAM, 0);
bind(server_sock, (struct sockaddr *)(&server_addr), sizeof(struct sockaddr));
listen(server_sock, 5);
sin_size = sizeof(client_addr);
for (;;) {
    if ((client_sock = accept(server_sock, (struct sockaddr *) &client_addr, &sin_size)) < 0) {
        printf("accept fail\n");
        continue;
    }
    printf("Client from %s:%d\n", inet_ntoa(client_addr.sin_addr), htons(client_addr.sin_port));
    int opt = 50;
    lwip_setsockopt(client_sock, SOL_SOCKET, SO_RCVTIMEO, &opt, sizeof(int));
    while ((recbytes = read(client_sock, buffer, BUF_SIZE)) > 0) {
        printf("%c", recbytes);
        buffer[recbytes] = 0;
    }
    sprintf(buffer, "HTTP/1.1 200 OK\nContent-Type: image/jpeg; charset=utf-8\nConnection: close\n\n");
    write(client_sock, buffer, strlen(buffer));
    close(client_sock);
}
```

ourairquality:
I have not seen this problem with the lwip v2 branch. There SO_RCVTIMEO takes a `struct timeval`, and might that be a problem in your code. For some example code see the wificfg web server.

kanflo:
Thanks, I found this when sorting out some code rot needed for me to test your PR but I will skip digging into this and go to testing. Also, thanks for pointing out the mistake regarding SO_RCVTIMEO.

kanflo:
[...]

PR #457, opened 2017-10-08
lwip: define LWIP_POSIX_SOCKETS_IO_NAMES as zero

ourairquality:
This patch gets the newlib standard stream descriptors playing well with the lwip socket descriptors and the spiffs file descriptors. The LWIP_SOCKET_OFFSET is now defined to be 3, rather than zero, to avoid clashing with the standard stream descriptors, and the SPIFFS_FILEHDL_OFFSET is bumped up to start after the lwip descriptors.

**This is not yet ready to merge.** I need to seek help from upstream lwip on the error that was disabled to allow this to build, why LWIP_SOCKET_OFFSET can not be used with an external FD_SET. A quick inspection did not find many problems, so I am probably missing something. But perhaps this will allow progress and more testing and feedback.

See also https://github.com/SuperHouse/esp-open-rtos/issues/445

### DIFF
--- a/core/newlib_syscalls.c
+++ b/core/newlib_syscalls.c
@@ -14,6 +14,21 @@
 #include <stdlib.h>
 #include <stdout_redirect.h>
 #include <sys/time.h>
+#include <lwip/sockets.h>
+
+/*
+ * The file descriptor index space is allocated in blocks. The first block of 3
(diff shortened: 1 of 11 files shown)


ourairquality:
The lwip change has now been accepted upstream and merged, so this might now be considered for merging.

kanflo:
Great!

PR #422, opened 2017-07-25
Add captdns to extras

flannelhead:
Responds to any DNS request with the access point IP. To use this, you also need to enable router and DNS server option fields in dhcpserver (https://github.com/SuperHouse/esp-open-rtos/pull/421). Should I create an example of how to configure this?

Ported from Sprite_tm's libesphttpd, upstream at http://git.spritesserver.nl/libesphttpd.git/

PR #458, opened 2017-10-09
Add libesphttpd to extras and supporting example

nochkin:
This is a fork based on the original work https://github.com/Spritetm/libesphttpd
It was modified to fit for esp-open-rtos.

Things currently work:
- HTTP web server
- Static files using heatshrink and gzip compression
- Templates
- CGI dynamic content
- WebSocket
- All built-in tests passed

"Bonus" features:
- WiFi configuration using web interface
- OTA update over http using a simple curl command (implemented using extras/rboot-ota)

It includes an example to show the features.


ourairquality:
Might adding the package zlib1g-dev to the .travis.yml file fix the build error, the missing `zlib.h`?

examples/esphttpd/cgi-test.c
```diff
@@ -0,0 +1,90 @@
...
+#include "cgi-test.h"
+
+typedef struct {
+	int len;
+	int sendPos;
+} TestbedState;
+
+
+int ICACHE_FLASH_ATTR cgiTestbed(HttpdConnData *connData) {
+	char buff[1024];
```
ourairquality: Looks very interesting. One nit the ICACHE_FLASH_ATTR might not be needed for esp-open-rtos as functions are linked to the flash by default.

Any chance that you could rework https://github.com/nochkin/libesphttpd so that the master branch follows the upstream, and create a separate branch for the esp-open-rtos modifications. That would make it much easier to follow upstream changes and to see what has changed.

nochkin: I have #define to null it. It's in platform.h file. This is done in hopes that I will push my changes to upstream some day.
It's currently forked from the master if that's what you mean.

nochkin: Yes, it needs zlib to build linux binaries for compression of static html/css/js files.
Can I add the zlib package or the maintainer has to do it?

UncleRus:
```
make[2]: Entering directory `/home/travis/build/SuperHouse/esp-open-rtos/extras/libesphttpd/libesphttpd/espfs/mkespfsimage'
ccache xtensa-lx106-elf-gcc -I../../lib/heatshrink -I../../include/libesphttpd -I.. -std=gnu99 -DESPFS_GZIP -DESPFS_HEATSHRINK   -c -o main.o main.c
main.c:26:18: fatal error: zlib.h: No such file or directory
 #include <zlib.h>
                  ^
compilation terminated.
```
**xtensa-lx106-elf-gcc**
Wrong compiler for mkespfsimage

nochkin:
Yes, I noticed that it's trying to use xtensa instead of local gcc. Trying to work it out.
I just call "make -C" in the component.mk, but I guess I need to pass some variable to force it. Maybe just doing CC=cc, but not sure how portable is it.

nochkin:
@kanflo, kudos to @Spritetm mostly :-)
I did not try the CaptDNS, that's why I did not include it in the list.
I'll check it out maybe later today and see if I could debug/fix it.

kanflo:
@nochkin,  FYI ```extras/wificfg``` include a fully functional (for me) captive DNS.

nochkin:
@kanflo, you are right. Now since you told me about it, I remember I saw it.
Frankly, I don't use this feature in my own projects, that's why I completely forgot about it.
I'm sure the CaptDNS works fine in libesphttpd initially. I'll see if it's easy to adapt/fix it, this way we'll have an alternative with http server embedded.
Thank you for pointing it out for me.

kanflo:
It seems ```recvfrom```never returns in ```dns_task``` in captdns,c. Trying to figure out why.

examples/esphttpd/user_main.c
```diff
@@ -0,0 +1,208 @@
...
+                .password = AP_PSK,
+                .max_connection = 3,
+                .beacon_interval = 100,
+            };
+            sdk_wifi_softap_set_config(&ap_config);
+
+            ip_addr_t first_client_ip;
+            IP4_ADDR(&first_client_ip, 172, 16, 0, 2);
+            dhcpserver_start(&first_client_ip, 4);
+            break;
```
kanflo: These are needed:
```
dhcpserver_set_dns(&ap_ip.ip);
dhcpserver_set_router(&ap_ip.ip);
```

nochkin: Agree, this is a good addition, but it makes sense for some specific needs which are not covered in the example.
The example does not utilize any routing or NS resolution, that's why it does not set it.
Unless there is some kind of client which goes crazy when DNS and GW are not set, but I never seen it.

After all, this is just a minimal viable product, more or less required to get started.
Here is a good example of what I mean: https://github.com/SuperHouse/esp-open-rtos/blob/master/examples/access_point/access_point.c

ourairquality: @nochkin fwiw I seem to recall Android not liking the link without a dns and/or router here - can't recall all the details.

nochkin: @ourairquality, that's interesting. I suspect it was old version or some other issue.
I remember I tried on Android 7 and 8 with no problems. I did not try anything older than that, but I think I could still locate some older device to test.

flannelhead: I recently also tested this very same captive DNS code (https://github.com/SuperHouse/esp-open-rtos/pull/422) and it seemed that indeed setting the DNS and/or router field is required for Android 6.

nochkin: @flannelhead, thank you for providing more information on that. I updated my example to reflect that.

UncleRus:

PR #461, opened 2017-10-10
Fix example broken by DHCP update

UncleRus:


extras/wificfg/wificfg.c
```diff
@@ -1995,7 +1995,7 @@ void wificfg_init(uint32_t port, const wificfg_dispatch *dispatch)
                 if (wifi_ap_dns < 0 || wifi_ap_dns > 1)
                     wifi_ap_dns = 1;
 
-                dhcpserver_start(&first_client_ip, wifi_ap_dhcp_leases);
+                dhcpserver_start(sdk_system_get_netif(SOFTAP_IF), &first_client_ip, wifi_ap_dhcp_leases);
```
ourairquality: Unfortunately this is not going to work because when this runs the interfaces have not even been created.

If you see the comment removed from the dhcpserver "netif_list isn't assigned until after user_init completes, which is why we do it inside the task".

Best to revert all these changes soon before other changes are piled on top of them.

Does anyone even need to run this dhcp server on the station interface and not also the AP interface?? If so then we probably need some flags to configure this early that are read later when the dhcp server task runs.

PR #469, opened 2017-10-16
Driver for Bosch Sensortec BME680 added

gschorcht:
I have written a driver for the brand new Bosch Sensortec BME680. BME680 is an ulta-low-power environmental sensor that integrates temperature, pressure, humidity and gas sensors in only one unit.

Please add it to the master branch.

### DIFF
--- a/examples/bme680/bme680_heating_profiles/bme680_heating_profiles.c
+++ b/examples/bme680/bme680_heating_profiles/bme680_heating_profiles.c
@@ -1,0 +1,126 @@
+/**
+ * Simple example with one sensor connected to I2C bus 0 and a sequence of
+ * heating profiles. The heating profile is changed with each cycle.
+ *
+ * Harware configuration:
+ *
+ *   I2C   +-------------------------+     +----------+
+ *         | ESP8266  Bus 0          |     | BME680   |
+ *         |          GPIO 5 (SCL)   ------> SCL      |
+ *         |          GPIO 4 (SDA)   ------- SDA      |
+ *         +-------------------------+     +----------+
+ */
+ 
+#include "espressif/esp_common.h"
+#include "esp/uart.h"
+
+#include "FreeRTOS.h"
+#include "task.h"
+
+// include communication interface driver
+#include "i2c/i2c.h"
+
+// include BME680 driver
+#include "bme680/bme680.h"
+
+// define I2C interface for BME680 sensors
+#define I2C_BUS         0
+#define I2C_SCL_PIN     GPIO_ID_PIN((5))
+#define I2C_SDA_PIN     GPIO_ID_PIN((4))
+
+static bme680_sensor_t* sensor;
+
+/*
+ * User task that triggers measurements of sensor every seconds. It uses 
+ * function *vTaskDelay* to wait for measurement results and changes the
+ * heating profile in each cycle.
+ */
+void user_task(void *pvParameters)
+{
+    bme680_values_float_t values;
+
+    TickType_t last_wakeup = xTaskGetTickCount();
+    
+    uint32_t count = 0;
+    
+    while (1) 
+    {
+        if (count++ < 60)
+            // disable gas measurement for cycle counter < 60
+            bme680_use_heater_profile (sensor, BME680_HEATER_NOT_USED);
+        else
+            // change heating profile in each cycle
+            switch (count % 5)
+            {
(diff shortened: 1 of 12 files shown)


ourairquality:
Could I ask if you know if the BME680 is readily available yet on small breakout boards etc? I would like to transition from the bme280 to the bme680 to get the voc sensor data too, and currently get them on small breakout board such as https://www.aliexpress.com/item/3In1-BME280-GY-BME280-Digital-Sensor-SPI-I2C-Humidity-Temperature-and-Barometric-Pressure-Sensor-Module-1/32817812439.html

gschorcht:
The only small breakout board I found til now is that from http://www.watterott.com/index.php?page=product&info=4329. Unfortunately, it is an online shop in Germany and shipping costs to foreign countries are quite expensive. I know, that a lot of people are already waiting for an Adafruit productt. But, there is no official information when it becomes available or whether it becomes available at all.

extras/bme680/bme680.h
```diff
@@ -0,0 +1,351 @@
...
+ * function *bme680_get_results_float* to fetch the measurement results.
+ *
+ * The measurement duration strongly depends on which measurements in the THPG
+ * measurement cycle are performed and which configuration parameters were set.
+ * It can vary from 1 RTOS (10 ms) tick up to 4500 RTOS ticks (4.5 seconds).
+ *
+ * @param   dev   pointer to the sensor device data structure
+ * @return        measurement duration given in RTOS ticks or -1 on error
+ */
+int32_t bme680_force_measurement (bme680_sensor_t* dev);
```
kanflo: Could I suggest returning a time in milliseconds or would 1ms be far too long?

gschorcht: I'm a bit unsure. Initially, this function returned the duration in milliseconds. It was the same in the SHT3x driver.

During the review of SHT3x driver, @ourairquality had the good idea to return the measurement duration simply in ticks so that they can be used directly as parameter of function vTaskDelay to wait. We had to realize that vTaskDelay(measurement_duration_in_ms+(portTICK_PERIOD_MS-1))/portTICK_PERIOD_MS) would result in a waiting time that is too short. Reason is that the first tick and in worst cases the first 2 ticks are shorter than portTICK_PERIOD_MS. To avoid problems at user task, we thought it is better to return a number of ticks that is high enough in any case.

gschorcht: At the moment, I'm separating the function to start the measurement and a function to get the measurement duration. This results from the discussion with @ourairquality about the SHT3x driver. 

Since the measurement duration depends only on the configuration of the measurement cycle, it is not necessary to compute the duration every time a measurement is started. It seems to be enough to configure the sensor, to get the measurement duration for this configuration and then to start the measurement multiple times. Measurement duration has to be only recalculated when the measurement configuration is changed. That is, as long as the measurement configuration isn't changed, the measurement duration can be considered as constant. So I'm changing the interface in that way:

```
bme680_set_... // set measurement configuration
uint32_t duration = bme680_get_measurement_duration(...);

while (1)
{
    if (bme680_force_measurement(...)) 
    {
        vTaskDelay(duration);
        bme_680_get_results_float(...);
    }
    ...
}
```

examples/bme680/bme680_heating_profiles/bme680_heating_profiles.c
```diff
@@ -0,0 +1,126 @@
...
+
+// include communication interface driver
+#include "i2c/i2c.h"
+
+// include BME680 driver
+#include "bme680/bme680.h"
+
+// define I2C interface for BME680 sensors
+#define I2C_BUS         0
+#define I2C_SCL_PIN     GPIO_ID_PIN((5))
```
UncleRus: There is no need to use GPIO_ID_PIN(...) macro here

gschorcht: Thank you for your revire. Removed GPIO_ID_PIN(...) macro.

extras/bme680/bme680.h
kanflo: I see you have good reasons to return the duration in ticks which is fine by me.

kanflo:
This PR seems ready for merging if no-one disagrees.

gschorcht:
Travis CI build failed due to timout :-(

PR #474, opened 2017-10-21
Fix a printf issue.

dora38:
When I pass a multi-line string to printf, only the first line is printed after PR #437 .
This fixes it.


ourairquality:
Thank you, this was introduced in https://github.com/SuperHouse/esp-open-rtos/pull/457 sorry. There is a similar issue in `_read_r` and would you like to roll in a fix for that too?

dora38:
Yes. #457 . sorry for misleading.

And yes. _read_r should be fixed. Thanks!

kanflo:
Ooh, shame on us and kudos to you for catching it. Thanks!

Issue #471, opened 2017-10-17
i2c clock stretching problem


gschorcht:
The implementation limits the maximum duration of clock stretching to only 10 times the duration of function **read_sda**. The I2C specification does not specify any timeout conditions for clock stretching. Each slave can hold SCL down as long as it needs. The master has to wait until SCL goes high before it can continue.

I have a CCS811 sensor here, that produces clock stretchings up to 60 us. With CLOCK_STRETCH=10 the time the master waits before timeout is much too less and the I2C communication fails. Increasing CLOCK_STRETCH to 200 works for this device but might still produce problems with other devices. 

Please increase the CLOCK_STRETCH constant a lot, remove it completly, or make it configurable.

kanflo:
Configurable would be the best option I think.

PR #475, opened 2017-10-24
Configurable I2C clock stretching

UncleRus:
Fix for #471

### DIFF
--- a/extras/i2c/i2c.h
+++ b/extras/i2c/i2c.h
@@ -40,9 +40,11 @@ extern "C" {
 /**
  * Define i2c bus max number
  */
-#ifndef MAX_I2C_BUS
-    #define MAX_I2C_BUS 2
+#ifndef I2C_MAX_BUS
+    #define I2C_MAX_BUS 2
 #endif
+
+#define I2C_DEFAULT_CLK_STRETCH (10)
 
 typedef enum
 {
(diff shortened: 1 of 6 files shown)


extras/i2c/i2c.h
```diff
@@ -41,9 +41,11 @@ extern "C" {
  * Define i2c bus max number
  */
 #ifndef MAX_I2C_BUS
-    #define MAX_I2C_BUS 2
+    #define I2C_MAX_BUS 2
```
ourairquality: Would want to change the `#ifndef MAX_I2C_BUS` too here.

Also this patch will need to be edited to remove the lwip changes. If working with modules is a challenge then just save the diffs to a file and remove all the changes from this branch then re-apply and make a forced push. Then just don't commit the lwip directory.

UncleRus: Thank you for your review!

UncleRus:
Going to merge this one if no objections.
@gschorcht any comments?

gschorcht:
@UncleRus 
Many thanks for your changes and all the work you have done in conjunction with them. Perhaps it would be better to leave function ```i2c_init``` as it is and to add a new function ```i2c_set_clock_stretching (bus, clock_stretch_counter)```?

```
i2c_init(I2C_BUS, I2C_SCL_PIN, I2C_SDA_PIN, I2C_FREQ_100K);
i2c_set_clock_stretching (I2C_BUS, MY_CLK_STRETCH);
```

Advantage would be that existing drivers and examples which are using i2c wouldn't have to be changed. Furthermore, clock stretching counter could be changed during runtime if neccessary.

PR #478, opened 2017-10-27
Fix a typo in ssd1306 driver that prevents compiling with c++ compiler

Petezah:
Simple fix.  Discovered while working on a project and thought others would find it useful.


UncleRus:
Nice catch!

PR #481, opened 2017-11-02
Fix mdnsresponder compilation issues

maximkulkin:


maximkulkin:
Hello?

extras/mdnsresponder/mdnsresponder.c
```diff
@@ -480,7 +482,7 @@ void mdns_add_TXT(const char* rKey, u32_t ttl, const char* txStr)
     }
 }
 
-void mdns_add_A(const char* rKey, u32_t ttl, struct ip_addr addr)
+void mdns_add_A(const char* rKey, u32_t ttl, ip_addr_t addr)
```
ourairquality: Perhaps these should be ip4_addr_t, and the other uses modified. So that it builds when IPv6 is enabled? Could you even test with IPv6 enabled.

There do not appear to be any examples using this, which is probably why it was not noticed. There is mdns support in lwip too, that seems to work but it uses a lot of stack space.

maximkulkin: Well, I agree that it is nice to get things right from the first try, I do not have neither expertise on various ways to build project nor willing to dive into intricacies of protocols. Let everybody scratch their own itch. This thing works for me. If there will be a person who uses IPv6 and this does not work, let him further improve this code.

For me it was outrageous that the code didn't even compile, I spent some time figuring how to fix it and decided to share.

ourairquality:
I suggest merging this as-is, and I'll follow up with a minor clean up patch.

PR #485, opened 2017-11-18
pwm fix 

Zaltora:
-Special duty state 0 and UINT16_MAX   crash when you change frequency  fix
-IRAM handler improvements
-Replace printf by debug system

I tested with a motor.
Can someone confirm to me this:  
If i put handler on  IRAM, it is improvement performance.  pwm interupt is called often.
No need to load function on flash with SPI transaction?

### DIFF
--- a/core/esp_timer.c
+++ b/core/esp_timer.c
@@ -83,6 +83,11 @@ int timer_set_frequency(const timer_frc_t frc, uint32_t freq)
     uint32_t counts = 0;
     timer_clkdiv_t div = timer_freq_to_div(freq);
 
+    if(freq == 0) //can't divide by 0
+    {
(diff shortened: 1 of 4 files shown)


Zaltora:
recent update fix a crash too if you use pwm_start() when duty is 0 or UINT16_MAX.

Zaltora:
Reworked pwm to work with a better behavior.
Only `pwm_start()`  and `pwm_stop()` can enable/disable the signal send to motor, even with special case.
Set duty and frequency  will not start the motor but if the motor is already on, duty and frequency will be dynamically set.

Zaltora:
The API work fine.
Good to merge if no objections.

PR #486, opened 2017-11-19
Fixed Repeated-start in I2C

ekalyvio:
The i2c_stop should not be there.
We have to keep the bus started and then issue again the i2c_start in order for the function to handle correctly the repeated-start.

### DIFF
--- a/extras/i2c/i2c.c
+++ b/extras/i2c/i2c.c
@@ -344,8 +344,6 @@ int i2c_slave_read(uint8_t bus, uint8_t slave_addr, const uint8_t *data, uint8_t
             goto error;
         if (!i2c_write(bus, *data))
             goto error;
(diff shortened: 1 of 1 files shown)


ekalyvio:
Bellow is a picture of how it is currently handled (the repeated-start). Observe in the yellow circle the **stop** condition and in the **blue** the start one.
inkedesp_li
This is wrong.

Bellow is the correct one:
inkedarduino_li
Observe the SDA going high while SCL is low. Then SCL goes high. And Finally, SDA goes low (before SCL) and later SCL follows.

Quick further info on repeated start [contains here](https://learn.sparkfun.com/tutorials/i2c)

UncleRus:
Great work, @ekalyvio !

PR #500, opened 2017-12-01
FreeRTOS v10.0.0

ourairquality:
FreeRTOS is now with Amazon and the license is more compatible. Here's an announcement
https://aws.amazon.com/blogs/opensource/announcing-freertos-kernel-v10/

Github repo:
https://github.com/aws/amazon-freertos

The code has not changed much. There are still a few issues to be followed up and addressed before this is ready to merge, e.g. the coroutine support is not include.


ourairquality:
There is a new feature in this release, stream buffers and message buffers.

ourairquality:
This could be consider for merging now. Based it on the svn repo rather than the amazona aws github repo as this seemed more complete and does not move files around.

PR #502, opened 2017-12-02
ws2812 i2s quick fix

Zaltora:
I found a problem when we do multiple following i2S_update.
I put a minimum control delay to prevent update do not work on led.
If someone get clue on i2s registers settings to avoid this. help welcome.


Zaltora:
i close this one, It is buggy. I don't know how fix it properly right now.

PR #483, opened 2017-11-14
I2C clock timing improvements - fix of #480

gschorcht:
Suggestion for improving clock timing a lot, at least for 80 MHz system clock and 80 kHz and 100 kHz I2C SCL clock.

PR #503, opened 2017-12-05
I2C optimization

flannelhead:
In an effort to make the I2C driver operate faster, I noticed a few optimization opportunities in the code. This code goes over 1 MHz at 160 MHz clock speed and I suppose ~800 kHz would be achievable with 80 MHz clock speed.

After the refactoring to multiple coexisting I2C bus instances, the code was making a lot of lookups in the `i2c_bus_descriptor` structure during the transaction. Especially in `i2c_delay()`, that caused quite a bit of overhead. Therefore the delay is now looked up only once in `i2c_start()`.

Furthermore, I noticed that in some occasions `gcc`with `-O2` was optimizing `i2c_delay()` aggressively. Namely, the counter wasn't even assigned on every inline occurrence of `i2c_delay()` in the code. This resulted in an infinite loop due to the counter wrapping over. The `delay` variable is now ~~declared `volatile` to force `gcc` to load the delay value to the corresponding register every time~~ properly indicated as an output variable in the inline assembly block, which prevents `gcc` from doing these harmful optimizations.

Also, `gpio_write()` and `gpio_read()` have some overhead since they are checking if the GPIO number is 16 (and thus they would have to use the RTC registers). I decided to change it so that GPIO16 isn't allowed (it is a bit special anyway), and the "normal" GPIO registers are manipulated directly. Thus this is a breaking change, but I hope anyone hasn't been using GPIO16 for I2C. As an optimization, the SDA and SCL pin masks are precalculated in `i2c_start()`.

`read_sda()` and `read_scl()` are changed such that they're not changing the pin state to HIGH when they're called. That is now done separately by `set_sda()` and `set_scl()`.

I've already tested the driver with [LSM6DS3](http://www.st.com/content/ccc/resource/technical/document/datasheet/a3/f5/4f/ae/8e/44/41/d7/DM00133076.pdf/files/DM00133076.pdf/jcr:content/translations/en.DM00133076.pdf) but of course it would benefit from testing by other people, too. The delay values will need retuning.

There are even more optimization opportunities. Mainly, I would like to assign the delay and the GPIO masks to global variables to save yet a few more cycles. Concurrent access by different bus instances would be protected by a mutex which is locked by `i2c_start()` and released by `i2c_stop()`. This would also improve the timing of transactions.

Just before I go further and perform the tuning (and possibly the optimization mentioned above), I would like to hear some opinions and discussion about this PR. I'll be also happy to introduce only a subset of this if all the changes are not desired.

This PR also incorporates the changes from PR https://github.com/SuperHouse/esp-open-rtos/pull/483.

### DIFF
--- a/extras/i2c/i2c.c
+++ b/extras/i2c/i2c.c
@@ -43,8 +43,8 @@
 // Following array contain delay values for different frequencies
 // Warning: 1 is minimal, that mean at 80MHz clock, frequency max is 320kHz
 const static uint8_t i2c_freq_array[][2] = {
-    [I2C_FREQ_80K]  = {255, 35},
-    [I2C_FREQ_100K] = {100, 20},
+    [I2C_FREQ_80K]  = {255, 80},
+    [I2C_FREQ_100K] = {100, 55},
     [I2C_FREQ_400K] = {10, 1},
     [I2C_FREQ_500K] = {6, 1}
 };
@@ -150,7 +150,7 @@ static inline bool read_sda(uint8_t bus)
 {
     gpio_write(i2c_bus[bus].g_sda_pin, 1);
     // TODO: Without this delay we get arbitration lost in i2c_stop
-    i2c_delay(bus);
+    // i2c_delay(bus); 
     return gpio_read(i2c_bus[bus].g_sda_pin); // Clock high, valid ACK
 }
 
@@ -203,6 +203,9 @@ bool i2c_stop(uint8_t bus)
     // Stop bit setup time, minimum 4us
     i2c_delay(bus);
     // SCL is high, set SDA from 0 to 1
+    read_sda(bus); 
+    // additional delay before testing SDA value to avoid wrong state
+    i2c_delay(bus); 
     if (read_sda(bus) == 0) {
         debug("arbitration lost in i2c_stop from bus %u", bus);
     }
--- a/extras/i2c/i2c.c
+++ b/extras/i2c/i2c.c
@@ -25,7 +25,6 @@
 #include "i2c.h"
 
 #include <esp8266.h>
-#include <espressif/esp_misc.h> // sdk_os_delay_us
 #include <espressif/esp_system.h>
 #include <FreeRTOS.h>
 #include <task.h>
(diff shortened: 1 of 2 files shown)


Zaltora:
I agree for GPIO16, this pin not got the same behavior and can cause some problem. Need to be controlled  when bus is initialized and return error if pin 16 is set.
To be compatible with people than use GPIO16, i suggest to use a define like
`#define GPIO_REGISTERS_CONTROL true`  to swap between register/ gpio API.
. I will test your change with my i2c devices (ds3231, bme280, ina3221, ... )

flannelhead:
@Zaltora Thanks for the comments, please report back about your test results :) Please note that the delays are not yet tuned, so the clock speed might not be accurate. I agree about using a `#define` to allow falling back to enabling GPIO16.

Does anybody have comments about locking the bus during transactions? The current implementation allows for multiple concurrent bus instances (presumably on different pins). However running multiple simultaneous I2C transactions doesn't seem to make much sense to me since the timing is done entirely by software - thus the timing would become inaccurate if one switches context between two I2C transactions running at equal priorities.

Hence I would like to add some locking by a mutex to prevent multiple I2C transactions from running concurrently on different buses - that would also allow for some further optimization as mentioned above. This would both improve the timing of individual transactions and also reduce the amount of overhead. @Zaltora, IIRC you did the multi-bus refactor of the I2C driver. How are you using it? Specifically, are you running multiple simultaneous I2C transactions on different buses?

Zaltora:
Well i didnt get  oscilloscope. If i guess your refactoring is faster, i will take 400Khz setting and append value to approach 300KhZ.
For locking bus, level 1 Api got a flag to signal if bus is busy. Can be better to use global static flag instead of a flag busy by bus. only one i2c transaction will be allowed.
Actually, i use one bus and semaphore API to protect  each transaction between task.  i dont no if it is better to lets user or i2c lib manage mutex but if you do it, i think only level 1 api need this protection. Anyway each library of esp-open-rtos uses level 1 api.
the i2c lvl 1 api got a function "force"  to force a i2c transaction to be shutdown if an other need to be started. It is usefull if a device is priority.  Actually, when a transaction is shutdown, "i2c_stop" detect it and return error.

flannelhead:
I'd now consider this PR ready for merging.

I've now added a define `I2C_USE_GPIO16` which will enable the `gpio_write` and `gpio_read` calls instead of direct register manipulation. The consequences are documented in the comments.

This is about as fast and accurate as I can get it. I've also now done a rough tuning of the delay values. I didn't see much value in implementing synchronization / mutexes in the library - it is good as it is.

PR #505, opened 2017-12-07
mDNS-responder: ipv6 support and fixes

ourairquality:
* Adds ipv6 support. Compiles with ipv6 enabled and then accepts ipv6 connections and answers AAAA questions.

* Fixes a few overflows of the reply buffer. Reduce the default reply buffer size, name it MDNS_RESPONDER_REPLY_SIZE, and allow overriding.

* Fix mdns_add_TXT.

* Prefer malloc to large stack buffers, to control stack sizes, and try to malloc only the buffer size needed where known in advance.

* Determine the IP addresses when responding, eliminating the update function  and the update task.

* Allow use in StationAP mode too, but bind to the station netif.

* Fix to compile without the debug output.

* Slightly better integration with lwip.

* Some code style changes, not comprehensive.

### DIFF
--- a/extras/mdnsresponder/mdnsresponder.c
+++ b/extras/mdnsresponder/mdnsresponder.c
@@ -1,9 +1,10 @@
 /*
  * Basic multicast DNS responder
  * 
- * Advertises the IP address, port, and characteristics of a service to other devices using multicast DNS on the same LAN,
- * so they can find devices with addresses dynamically allocated by DHCP. See avahi, Bonjour, etc
- * See RFC6762, RFC6763
+ * Advertises the IP address, port, and characteristics of a service to other
+ * devices using multicast DNS on the same LAN, so they can find devices with
+ * addresses dynamically allocated by DHCP. See avahi, Bonjour, etc. See
+ * RFC6762, RFC6763
  *
  * This sample code is in the public domain.
  *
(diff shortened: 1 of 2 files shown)


ourairquality:
@maximkulkin This update fixes a good number of bugs, and might reduce memory usage a little. Testing and feedback welcomed.

maximkulkin:
I've tried this new code in my project and TXT records does not seem to be correct. I'm going to debug it.

ourairquality:
The TXT records were completely broken, but an attempt was made to get them working and a simple example was working. It helped debugging to use wireshark to inspect the mdns packets - it gives a breakdown of the messages and even detected some that were broken. If you find an example that is broken then please post it and I'll take a look too.

maximkulkin:
Ok, I guess the problem was on my side. Still don't get why there is no API to add *multiple* entries to TXT record. Guess I need to add it. Here is API from esp-idf: http://esp-idf.readthedocs.io/en/latest/api-reference/protocols/mdns.html?highlight=mdns#_CPPv220mdns_service_txt_setP13mdns_server_tPKcPKc7uint8_tPPKc

ourairquality:
For a more comprehensive implementation then perhaps it would be a better investment to work on the lwip mdns responder - it will have a wider community to work on it. Problem is that is uses a lot of stack space, allocates large buffers on the stack.

maximkulkin:
I tested this patch more and decided to keep old implementation. mDNS service does not work at all in my setup and I think it has something to do with initializing it before obtaining IP address (old implementation had a task to wait/check for that). I guess, it is not handled at all in new code.

I also checked lwip and while having a better API, it a) still requires you to initialize service AFTER IP was obtained; b) (more importantly) it requires a lot of configuration through preprocessor defines which kind of inconvenient in case you develop a component. I really do not want to provide a long documentation on which CFLAGS to provide in order to use my component from an application, it should be just component inclusion. But in that case it is tricky to set options without shadowing some user provided options.

Long story short, I decided to fork old mdnsresponder code and enhance it to my needs.

maximkulkin:
Ok, I take it back: I personally can leave with this patch. It's OK to merge.

ourairquality:
The lwip mdns responder appeared to work well with lwip, it can work with the LWIP_NETIF_EXT_STATUS_CALLBACK option, so handle configuration changes to the interface, and thus work through a change in the IP. If these options are commonly used then lets consider enabling the support by default. Perhaps we could also enable the lwip dns client support for mdns too by default so code plays well with other mdns devices. Fwiw the mdns responder is not currently working with the softap interface, the multicast frames are not getting out, it's probably a trivial matter to fix if we had the source code but need to explore how best to deal with it as-is.

PR #511, opened 2017-12-08
Newlib: implement locks

ourairquality:
Looks like implementing mutex locks for newlib will help solve a few issues. The current use of critical regions for locking causes stutters or worse, such as getchar blocking with interrupts disabled.

Newlib has some statically allocated and initialized locks. It has been patched to make them globally visible so they can be initialized at startup.

The malloc lock is used early, and there is a runtime check to handle this.

So far only one use of a newlib lock in a critical region has been spotted, a use of the malloc lock in the ppTask, and this has been addressed with a binary patch to libpp.a.

This will need more testing, and feedback is welcomed.


meowthink:
Works like a charm!

UncleRus:
Tested it with sysparam_editor and other examples. Works great!
Going to merge this PR.

PR #513, opened 2017-12-09
Makefile: add flag to disable C++ components

flannelhead:
When experimenting with toolchains, I found it difficult to build a C++-enabled toolchain with crosstool-NG. Therefore I propose to add a flag which disables building `*.cpp` files. The only C++ file in `core` is `cplusplus_operators.cpp` which is not an essential part if one isn't programming in C++. Disabling C++ files in Makefile enables building `esp-open-rtos` with a C-only toolchain.


flannelhead:
I think this can be closed. I managed to get my cross toolchain compiled with C++ support, and most people will use the one built by `esp-open-sdk` anyway, and it has C++ support.

PR #522, opened 2017-12-15
extras/new_pwm: multi-channel soft PWM with example

nochkin:
This is a fork of [ESP8266_new_pwm](https://github.com/StefanBruens/ESP8266_new_pwm).
The main difference with the included "extras/pwm" is that the new_pwm supports multi-channels PWM and 100% duty mode. Also, it's a drop-in replacement of the Espressif's pwm.c, but not compatible with "extras/pwm" which is included with esp-open-rtos.

I used it in the past with non-OS SDK and had positive experience.
I created a fork to modify it to work under esp-open-rtos.
The correspondent example is included.


nochkin:
Actually, just realized that I have to take licensing into consideration. I'm kinda new into contribution into open source community.
If someone else who is more knowledgable about the topic, please chime in. The original license of the code is GPL-2. I hope it still can be merged into esp-open-rtos, but wanted to double check.

UncleRus:
> I hope it still can be merged into esp-open-rtos

Unfortunately this is not possible. GPL is incompatible with MIT.

nochkin:
Thanks for clarifications.
Besides MIT, what other popular licenses are compatible when contributing to esp-open-rtos? BSD, Apache, something else?
Sorry, I just found so many conflicting and confusing information when researching.

skcks:
call pmw_init will stop printf run, how to fix this problem?

Issue #510, opened 2017-12-08
heap and memory question ?


Zaltora:
-official esp8266 RTOS sdk use heap_5.c memory management for freertos. esp-open-rtos too or another model ?
- "malloc" and "free"  will allocate memory from the task heap (inside) ? or oustide of the task ?
- if a task "malloc" memory, other task can manipulate this memory ? 
- i got a situation : one task "malloc" memory, send memory ptr with a queue to another task and this task after treatments "free" this memory. the memory is "free" truly ?

I got some memory problem.

Zaltora:
if i do this:
`#define "MY TEXT HERE"`
and use it in my function, the text is on ROM memory or RAM ?
when i do this: 
`const char* text = "MY TEXT HERE"`
The text is in ROM ?
I call multiple function with text and think the compiler will stock these text in ROM. its good ?
if not, can i force put them in ROM ?

Zaltora:
`#define configTOTAL_HEAP_SIZE		( ( size_t ) ( 32 * 1024 ) ) `

Max heap can be change ? i remenber, that an update of linker from @ourairquality  save a lot of RAM. (Maybe not the same RAM)

Zaltora:
Something interesting, the official RTOS SDK got ~70KB heap
SDK version:1.5.0-dev(a8c4880)
Heap :69984

Zaltora:
I think that espressif use Instruction RAM/FLASH cache ram. get more heap. 
[esp8266 memory map](https://github.com/esp8266/esp8266-wiki/wiki/Memory-Map)
When we look last commit of them:
> 1. Change cache size from 32KB to 16KB, reverse this 16KB region as heap;
Someone know how to disable cache for this region: 0x40108000 ? if we can do this,  it will allow 16KB heap more. (if we not change other things).

Zaltora:
After compare DPORT register for both SDK:
esp-open-rtos:
`DPORT SPI_CACHE_RAM         3FF00024      0000001A`
esp8266 rtos sdk:
`DPORT SPI_CACHE_RAM         3FF00024      0000000A`

Maybe we just need to setup this register to get more 16Kb more heap.

ourairquality:
There are some clues in the include files:
```
/* Details for SPI_CACHE_RAM register */
#define DPORT_SPI_CACHE_RAM_BANK1  BIT(3)
#define DPORT_SPI_CACHE_RAM_BANK0  BIT(4)
```

[...]

PR #523, opened 2017-12-16
Reclaim a chunk of unused dram in the wdev bss area, 8000 bytes.

ourairquality:
Some people appear desperate for more dram, and there is 8000 bytes unused. This patch attempts to reclaim that region, to put it onto the malloc free list where is can be used. This is not ideal as it is a fragmented region, but it is dram so faster and more flexible than the 16k+ or iram that might also be added see https://github.com/SuperHouse/esp-open-rtos/issues/510 Some testing and feedback would be welcomed, perhaps also if someone could look over the wdev disassembly and see if they concur about the position and size of this region.

### DIFF
--- a/core/app_main.c
+++ b/core/app_main.c
@@ -135,6 +135,8 @@ static void IRAM default_putc(char c) {
 }
 
 void init_newlib_locks(void);
+extern uint8_t sdk_wDevCtrl[];
+void nano_malloc_insert_chunk(void *start, size_t size);
 
 // .text+0x258
 void IRAM sdk_user_start(void) {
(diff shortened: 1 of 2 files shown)


Zaltora:
Some people like me ^^. I got program to analyze memory from 3FFE8000h to 40000000h. And see a lot of zeroes in bss area. I think you speak about it ? Yes, it is fragmented. I don't no why.
Nice addition

Zaltora:
@ourairquality ,
I was looking for the iram section. I didn't know well libc but it is possible to get special malloc/free function for 16K iram section ? Don't need to add it as heap already, just for user usage (If Freertos permit it).

ourairquality:
We might have a problem with the travis build testing, it now reports exceeding the maximum time for jobs. Would anyone know if this is a limit that can be increased, or do we need to start being more selective about the examples that are built?

@Zaltora The same approach as used here can be used to insert the unused iram as a malloc chunk too. However it's not ready yet. This PR seems to be a useful one to merge, just makes more dram usable. Whereas moving a lot of data to iram might have a performance trade off, so might need to be a compile time option. I have something working on some simple tests and for example I see the following on the wificfg example:

```
Free dram heap     48624 bytes
Free iram heap    6744 bytes
Free heap    55368 bytes
```

Zaltora:
You don't use heap_5 from FreeRTOS to do this. But directly the lower layer with libc and malloc ?
if we can do this directly other layer. Why use FreeRTOS heap management ? What is the difference?

Question about compilation option, possible case can be :
- IRAM  DISABLE ( like now 16K + 16K chunk IRAM FLASH CACHE)
- IRAM  ENABLE USER (user got a special malloc_iram(s)  and know limitations of this, freertos use dram)
- IRAM  ENABLE RTOS (the iram can be use by user, queue, stream buffer, ....)
What did you think ? it is possible ? it is good idea to got option like this ? ( maybe already your plan)

A lot of question ^^ Subject very interesting.

SaimenSays:
Using ourairquality fork for some days. No issues recognized up to now and less out of RAM crashes.

@ourairquality Thanks for that nice improvement!

PR #529, opened 2017-12-27
newlib: add an option to combine some locks.

ourairquality:
The locks are using a bit of the limited ram. It's probably fine to combine some of these to use the same mutex, and this patch does so for the set initialized early. The file locks will still be separate, and dynamically created, so a thread blocking on them will not deadlock all uses of newlib that need a lock. See https://github.com/SuperHouse/esp-open-rtos/issues/510#issuecomment-354001234


UncleRus:
Thank you!

Zaltora:
Tested, 1k heap more with this patch :)

PR #536, opened 2018-01-04
Flash size as megabytes

ckerkhoff:
When building I noticed the warning, "WARNING: Flash size arguments in megabits like '16m' are deprecated." This should take care of that.

### DIFF
--- a/examples/i2s_audio/Makefile
+++ b/examples/i2s_audio/Makefile
@@ -1,6 +1,6 @@
 PROGRAM=i2s_audio_example
 EXTRA_COMPONENTS = extras/spiffs extras/i2s_dma
-FLASH_SIZE = 32
+FLASH_SIZE = 4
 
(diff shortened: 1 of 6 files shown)


ckerkhoff:
It looks like the CI build server is using a esptool version older than Aug 28 2016, which is when they started accepting megabytes for the flash size argument.

nochkin:
The esptool is a part of [esp-open-sdk](https://github.com/pfalcon/esp-open-sdk) which is being built for esp-open-rtos:
https://github.com/SuperHouse/esp-open-rtos/blob/master/utils/travis_build/install_toolchain.sh
In order to support your patch, [esp-open-sdk](https://github.com/pfalcon/esp-open-sdk) needs esptool updated.

P.S.: I'm using external esptool which supports the appropriate arguments. The same way as you do probably.

PR #542, opened 2018-01-18
Minor changes in CCS811 driver

gschorcht:
- error messages added to one example in case of initilization error
- some notes added to README.md 
- changed debug output in the driver


gschorcht:
Failed due to timeout

PR #553, opened 2018-01-22
Http client OTA

AndreaGreco:
Add extra for lightweight HTTP client ota upgrade
With relative example


UncleRus:
Thank you!

oremic:
I used this library for the basic sonoff with the library maximkulkin / esp-wifi-config.
But on restart I lose the parameters saved in sysparameter.
Flashsize = 8m
how can i solve thanks

PR #556, opened 2018-01-25
Add makefile flash hooks

andrewclink:
Every time I update I find that I'm editing common.mk to fix my terminal program. With this PR, I can just stick functions in local.mk (which is in .gitignore) like so:

    define will_flash
        @echo About to flash, fix terminal program, etc.
    endef

    define did_flash
        @echo Flash finished, clean up.
    endef

It might even be possible to determine the exit status of esptool to pass as an argument, but it's not necessary for me.


andrewclink:
I'm using GNU Make 3.81 which ships with macOS High Sierra. Obviously this needs to be tested on other versions of Make.

PR #563, opened 2018-02-07
Improve extras/sntp -- Issue #562

### LINKED ISSUE #562: sntp_fun.c incorrecty computes fractional seconds
Examination of the results of `sntp_get_rtc_time(&us)` in operation shows both values over 1,000,000 microseconds as well as negative values. When using the SNTP module, this breaks other time-related utilities as well.

The issue appears to be `sntp_fun.c:84` which appears to fail to properly rescale the result by 4096. It appears that changing that line to 
```
		*us = ((base * cal) % (1000000U<<12)) >>12;
```
resolved the range issue and a quick test shows an advance of ~0.10 seconds with a task delay of 100 ms. 

Patch coming after I review the algorithm and confirm the correctness of the patch.

jeffsf:
Primary change is correcting the missing scaling of the microseconds calculation 
Resolves #562 

Other changes address:
* Capture RTC count before any logging
* No "obligatory" printf() logging
* Change logged message to show:
  * adjustment about to be made in seconds
  * time since last SNTP update
  * ratio of the two as ppm
  * "cal" value (raw as int, as before)
* Initialize "cal" value from sdk_system_rtc_clock_cali_proc() 

I'm more than happy to squash this, or split into mutiple commits

### DIFF
--- a/extras/sntp/component.mk
+++ b/extras/sntp/component.mk
@@ -5,4 +5,9 @@ INC_DIRS += $(sntp_ROOT)
 # args for passing into compile rule generation
 sntp_SRC_DIR =  $(sntp_ROOT)
 
+# For SNTP logging, either supply own SNTP_LOGD
+# or define SNTP_LOGD_WITH_PRINTF (see sntp_fun.c)
+
+# sntp_CFLAGS = $(CFLAGS) -DSNTP_LOGD_WITH_PRINTF
+
 $(eval $(call component_compile_rules,sntp))
--- a/extras/sntp/sntp_fun.c
+++ b/extras/sntp/sntp_fun.c
@@ -13,6 +13,11 @@
 #include <esp/rtc_regs.h>
 #include <sntp.h>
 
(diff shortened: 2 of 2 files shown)


jeffsf:
As I "peel the onion" on the SNTP implementation, it looks like there are other issues to resolve as well. I'm planning on rebasing this into two commits
* Resolve calculation error
* Improve logging
so that various portions could be cherry-picked, if appropriate.

jeffsf:
86fa1ad represents the compiler flag

I tried rebasing into a functional-correctness commit and a debug-output commit and wasn't completely successful. I could give it another try, or just go ahead and squash this entire thing if you choose to accept it.

The new LWIP drop of #554 seems to contain a near rewrite of the LWIP SNTP code that now seems to:
* Consider RTT in setting time
* Handle broadcast NTP
* Accept NTP server from DHCP

Rather than reinvent the wheel on the obsolete LWIP core of extras/sntp, I'm going to continue development off a merge between `master` and #554.

PR #571, opened 2018-02-16
Newlib: update to version 3.0.0

ourairquality:
* Also include some code from libc/unix such as basename.


jeffsf:
It's certainly a welcome change to see newlib has dropped support for ancient C and has readable function prototypes. Looks like they've bumped `time_t` up to 64 bits as well. Thanks again @ourairquality -- looking forward to this update.

PR #582, opened 2018-02-27
i2c: revise timing calculation

flannelhead:
Look-up tables were used for determining the delay loop counts before.
Based on these hand-tuned values, the loop overhead was estimated for
each option -- 80 and 160 MHz, fast and slow GPIO access. Instead of the
great number of tunable parameters one now only has to tune the overhead
values if the code is changed.

Functions were added to the API which allow setting an arbitrary
frequency. API backward compatibility is retained.

### DIFF
--- a/extras/i2c/i2c.c
+++ b/extras/i2c/i2c.c
@@ -37,29 +37,18 @@
 #define debug(fmt, ...)
 #endif
 
-// The following array contains delay values for different frequencies.
-// These were tuned to match the specified SCL frequency on average.
-// The tuning was done using GCC 5.2.0 with -O2 optimization.
-const static uint8_t i2c_freq_array[][2] = {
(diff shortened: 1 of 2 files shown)


Zaltora:
You tested if frequency signals are good  for each option and frequency ?
Nice update. I guess, this soft i2c library will be the best of the esp8266 ^^

flannelhead:
@Zaltora: I tested with some configurations. Probably I should still check all the combinations (80MHz/register access, 80MHz/GPIO API, 160MHz/register access and 160MHz/GPIO API).

However, the formula that was introduced here produces the very same delay values that were hand-tuned and hardcoded before (if not even more precise!) The linear delay formula and coefficients are based on those hand-tuned values but the number of parameters is reduced.

UncleRus:
Thank you!

Issue #564, opened 2018-02-11
Task memory leaking after 0.0.1-134-gb83c262


linuxforum5:
The vTaskDelete() is not freeing all unused memory.
This is the sample code:
[heaptest.c.txt](https://github.com/SuperHouse/esp-open-rtos/files/1714552/heaptest.c.txt)

The free heap is constant (43692) in version 0.0.1-134-gb83c262.
[screen.log.0.0.1-134-gb83c262.txt](https://github.com/SuperHouse/esp-open-rtos/files/1714553/screen.log.0.0.1-134-gb83c262.txt)

Memory leaks in version 0.0.1-163-g546cc47 and in further versions.
[screen.log.0.0.1-260-g5f8b3d4.txt](https://github.com/SuperHouse/esp-open-rtos/files/1714554/screen.log.0.0.1-260-g5f8b3d4.txt)

ourairquality:
Thank you for the report. A first guess would be to look at the newlib integration into FreeRTOS. On a task being deleted it should call back into newlib to free up the newlib resources, and FreeRTOS calls _reclaim_reent() in libc/reent/reent.c

It might be that some of the I/O resource are not being freed. Some of these might be allocated in newlib/libc/stdio/findfp.c and there is a clean up function too.

I doubt there was anything in the commit you noticed that caused this apart from simply enabling the FreeRTOS integration with newlib which was essential. We will just have to work through it to get it freeing the resources properly.

Have you tried simply closing the stdin stdout and stderr FILEs before deleting the task to see if that solves this problem, or reduces the amount leaked?

Newlib also has an option to use global streams, and that might be another option to explore.

PR #583, opened 2018-02-28
newlib: rebuild with the global stdio streams enabled.

ourairquality:
This change means that there is only one set of global stdin, stdout, and stderr FILE streams shared by all the threads. This reduces memory usage and avoids having to close these streams before threads exit. These streams still have a lock to synchronise access.

This was explored to address https://github.com/SuperHouse/esp-open-rtos/issues/564 but it seems to save a useful bit of memory too. Guess the stdio file stream memory usage adds up with a good number of threads.

This required adding support to newlib for this global stdio streams option when using the reent small option. Shall submit the change upstream for feedback there.

Testing and feedback welcomed.


Zaltora:
I got no problem using this PR in my project. it allow to me 3.5k more heap.

ourairquality:
The newlib patch was accepted upstream. There were only some minor documentation improvements. This patch seems a good 'next step', ahead of the icache/iram usage patch.

PR #587, opened 2018-03-02
newlib: skip locking when within the NMI Irq.

ourairquality:
NMI Irq error paths call into printf which will use the newlib locks but these can not be used within the NMI, a task switch can not occur here. This case would be a terminal error path that is attempting to write some debug message in the process, so just bail out of the locking in this case. As a warning a ':' character is emitted and this will typically prefix lines emitted in this context - it would be an error for this path to be take in normal operation.

### DIFF
--- a/core/newlib_syscalls.c
+++ b/core/newlib_syscalls.c
@@ -303,6 +303,10 @@ void _lock_acquire(_lock_t *lock) {
 
 void _lock_acquire_recursive(_lock_t *lock) {
     if (locks_initialized) {
+        if (sdk_NMIIrqIsOn) {
+            uart_putc(0, ':');
+            return;
+        }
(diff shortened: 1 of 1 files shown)


core/newlib_syscalls.c
```diff
@@ -303,6 +303,10 @@ void _lock_acquire(_lock_t *lock) {
 
 void _lock_acquire_recursive(_lock_t *lock) {
     if (locks_initialized) {
+        if (sdk_NMIIrqIsOn) {
+            uart_putc(0, ':');
```
UncleRus: Is this `uart_putc` call really necessary?

ourairquality: Thought was that if it gets here it is an fatal error path, so emit some diagnostic info. A restart is going to be verbose anyway. Welcome to omit them.

Issue #586, opened 2018-03-02
Add ability to customize stop/parity bits for UART


tkremeyer:
As far as I know, currently there is no possibility to customize the stop bit count and parity bit type for UART communication. In [esp/uart.h](https://github.com/SuperHouse/esp-open-rtos/blob/master/core/include/esp/uart.h) it is only possible to set a custom baud rate.
Since a lot of projects (as my DMX project) require 1.5 or 2 stopbits and/or parity bits, customizing this value would enable those projects to use esp-open-rtos.

PR #590, opened 2018-03-11
UART: Add ability to configure stopbits and parity bit

tkremeyer:
Some functions and typedefs to customize stopbit count and parity bit type for UART.  
Includes an example about how to use the functions.
(Related Issue: #586 )


UncleRus:
Thank you!

PR #591, opened 2018-03-12
extras/dhcpserver - Fix memory leak on stop

### LINKED ISSUE #589: dhcp server memory leak
Hi guys
There is a memory leak in the function dhcpserver_stop()
Before freeing the server_state_t * state pointer, the netconn and leases pointers should be deallocated.

        vTaskDelete(dhcpserver_task_handle);
        netconn_delete(state->nc);/****ADD THIS LINE****/
        free(state->leases);/****ADD THIS LINE****/
        free(state);
        dhcpserver_task_handle = NULL; 

I saw this issue when I was connected to the STATION AP mode of the ESP, then configured it in station mode and then took it back again to the STATION AP mode. After this, I was not able anymore to connect to it from my phone, I had the mesage:  station: ec:d0:9f:03:c7:b6 join, AID = 1, but never got an IP addr. To do so I had to restart the chip which is not what I wanted.

maximkulkin:
Free up all allocated resources on DHCP server stop.

Fixes #589

### DIFF
--- a/extras/dhcpserver/dhcpserver.c
+++ b/extras/dhcpserver/dhcpserver.c
@@ -99,6 +99,7 @@ void dhcpserver_start(const ip4_addr_t *first_client_addr, uint8_t max_leases)
         dhcpserver_stop();
 
     state = malloc(sizeof(server_state_t));
+    memset(state, 0, sizeof(*state));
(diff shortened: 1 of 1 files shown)


UncleRus:
Thank you!

PR #596, opened 2018-03-28
reset stdout redirection for exception dumps

phkehl:
Reset stdout to the default direct-to-UART handler on exceptions and
abort()s. Users may have used set_write_stdout() to redirect output
elsewhere, which likely isn't going to work in the exception handler.


phkehl:
Thanks for you comments! That makes sense.
I've studied the code a bit more. I found that set_user_exception_handler() lets me do what I want already. I'll close this.

PR #597, opened 2018-03-29
uart: add ability to configure byte length

nicogrx:
This patch to optionally change data bits length of serial bus.

Signed-off-by: Jean-Nicolas Graux <nicogrx@gmail.com>


nicogrx:
Thank you!

PR #608, opened 2018-04-16
Fix for sdio_write_sectors

rerobika:
The issue was found by using `f_mkdir("directory_name")`.
The function always returned `FR_DISK_ERR` due to an unhandled case in `sdio_write_sectors()`.
According to this documentation: http://www.convict.lu/pdf/ProdManualSDCardv1.9.pdf#page=92 stop transmission command should be sent after writing multiple blocks.
This patch fixes this bug also adds a test case for it.


rerobika:
@UncleRus please take a look at it.

UncleRus:
Many thanks!
Merging it right now

PR #609, opened 2018-04-16
Initial HW PWM (Delta-Sigma)

Zaltora:
Hi, it is a HW PWM driver (WIP)
I will do test with signals soon.
- Hw pwm will be maybe renamed to delta-sigma, depends of the signal ?
- Frequency calculation need verification and be enhanced.


flannelhead:
Looks like a great addition to the OS.

One nitpick: before merging, you might want to pay attention to naming. AFAICT the HW functionality that is used here is actually a delta-sigma modulator. Delta-sigma modulation is not pulse width modulation but rather pulse *frequency* modulation where the pulse width is constant and the frequency is altered. See e.g. [this figure](https://en.wikipedia.org/wiki/Delta-sigma_modulation#/media/File:Pulse-density_modulation_1_period.gif).

It would be good to check the actual output signals. I could do it later this month if my logic analyzer is sufficient for that.

Zaltora:
Yeah, i think i will rename thing to be clear. (was planned, now i valid!)
I will check tomorrow with an oscilloscope.

When you set duty, in fact you change frequency. (the target mentioned here:
```
// Freq = (80,000,000/prescale) * (target / 256) HZ           (0   < target < 128)
// Freq = (80,000,000/prescale) * ((256 - target) / 256)  HZ  (128 < target < 256)
```
I will rename functions name too

Zaltora:
After test, I see Delata-sigma signal is very sensible to capacity line:
Test with a low target and a prescale of FF:
img_20180417_132945

Change target from 0 to 255 change my voltage ouptut from 0V to 3.3V. Measurement done with (x1) sensor. I will do test next week with (x10). It is feel  easy do use it like a DAC.
Tested on 2 different pin. The signals change between them because of different capacity. peek-to-peek is increase for one pin.
the signal is not proper, increase prescale will linearized more the signal i guess... I understand why the max target i can get is 127 to control my motor. I got a transisor with Vgs between 1 and 3 volt.

Don't test it with a logic analyzer, you will get strange result.

I am not sure if the formula here is good:
```
// Freq = (80,000,000/prescale) * (target / 256) HZ           (0   < target < 128)
// Freq = (80,000,000/prescale) * ((256 - target) / 256)  HZ  (128 < target < 256)
```
because voltage increase close to a linear way . Or maybe the frequency decrease after 128 but voltage still increase.

Zaltora:
Second test:
- i was wrong last week, the test was with a prescale of 0.  that why the signal is quite smooth.
- Test with prescale of FF look like a PWM square signal.
- Change duty 0 to 128 increase frequency and change duty 128 to 255 decrease frequency 
- Voltage (mean) increase with duty 0 to 255.

Zaltora:
I will do a last test tomorrow to see what happen when the CPU clock is at 160MHz

PR #612, opened 2018-04-20
Fix SDIO send_command

rerobika:
This patch fixes the problem, while receiving `R1b` type of response from the card.
In case of `R1b` response the continuous stream that the card sends must be checked whether the signal indicates busy status and wait until a non zero response.


rerobika:
cc: @UncleRus

UncleRus:
Thank you!

PR #621, opened 2018-04-30
sdk_wifi_station_dhcpc_stop: ensure the client is flagged as stopped.

ourairquality:
Need to flag the dhcp client as stopped, even if the netif is not yet initialized, because the flag is used to control the starting of the dhcpc.


UncleRus:
Thank you!

PR #634, opened 2018-06-05
sysparam: always create the semaphore on init.

ourairquality:
Otherwise when initialization was not successful it may fail with an assertion in FreeRTOS rather than returning SYSPARAM_ERR_NOINIT.


UncleRus:
Thank you!

PR #635, opened 2018-06-05
ssd1306: Allow SPI3 support to be disabled

quietboil:
At the moment SPI3, unlike SPI4 or I2C, cannot be disabled and the driver always includes SPI3 protocol code. This change allows a program to exclude SPI3 if it is not needed.

For example, when only SPI4 is needed, a program would set these variables in the Makefile:
```make
EXTRA_COMPONENTS = extras/ssd1306
SSD1306_SPI4_SUPPORT = 1
SSD1306_SPI3_SUPPORT = 0
SSD1306_I2C_SUPPORT = 0

PROGRAM_CFLAGS = $(ssd1306_CFLAGS)
```
The change in `ssd1306.c` is to suppress "unused variable" warning as `j` is used only by SPI3.


UncleRus:
Thank you!

PR #661, opened 2018-07-26
fix http ota

apiel:
I had some issue to flash my esp8266 (sonoff) with http client ota.

sdk_spi_flash_write requires a word aligned
Assuming size is always a multiple of 4 bytes.

See ./extras/rboot-ota/ota-tftp.c

### DIFF
--- a/extras/http_client_ota/http_client_ota.c
+++ b/extras/http_client_ota/http_client_ota.c
@@ -68,17 +68,21 @@ static unsigned int ota_firmaware_dowload_callback(char *buf, uint16_t size)
         return -1;
     }
 
-    // Ready for flash device, the erase NANDFLASH Block
-    if (flash_offset % SECTOR_SIZE == 0) {
-        unsigned int sector;
-
-        sector = flash_offset / SECTOR_SIZE;
-        sdk_spi_flash_erase_sector(sector);
+    int offset = 0;
+    if(size && ((uint32_t)buf % 4)) {
+        // sdk_spi_flash_write requires a word aligned
+        // Assuming size is always a multiple of 4 bytes.
+        uint32_t first_word;
+        memcpy(&first_word, buf, 4);
+        sdk_spi_flash_write(flash_offset, &first_word, 4);
+        memmove(LWIP_MEM_ALIGN(buf),&buf[1],size-4);
+        buf = (uint32_t *)LWIP_MEM_ALIGN(buf);
+        offset += 4;
+        size -= 4;
     }
-
-    // Write into Flash
     sdk_spi_flash_write(flash_offset, (uint32_t *) buf, size);
     flash_offset += size;
+
     return 1;
 }
 
--- a/extras/http_client_ota/http_client_ota.c
+++ b/extras/http_client_ota/http_client_ota.c
@@ -80,6 +80,8 @@ static unsigned int ota_firmaware_dowload_callback(char *buf, uint16_t size)
         offset += 4;
         size -= 4;
     }
+
+    // Write into Flash
     sdk_spi_flash_write(flash_offset, (uint32_t *) buf, size);
     flash_offset += size;
 
--- a/extras/http_client_ota/http_client_ota.c
+++ b/extras/http_client_ota/http_client_ota.c
@@ -76,7 +76,7 @@ static unsigned int ota_firmaware_dowload_callback(char *buf, uint16_t size)
         memcpy(&first_word, buf, 4);
         sdk_spi_flash_write(flash_offset, &first_word, 4);
         memmove(LWIP_MEM_ALIGN(buf),&buf[1],size-4);
-        buf = (uint32_t *)LWIP_MEM_ALIGN(buf);
+        buf = LWIP_MEM_ALIGN(buf);
         offset += 4;
         size -= 4;
     }


extras/http_client_ota/http_client_ota.c
```diff
@@ -68,17 +68,23 @@ static unsigned int ota_firmaware_dowload_callback(char *buf, uint16_t size)
         return -1;
     }
 
-    // Ready for flash device, the erase NANDFLASH Block
-    if (flash_offset % SECTOR_SIZE == 0) {
-        unsigned int sector;
-
-        sector = flash_offset / SECTOR_SIZE;
-        sdk_spi_flash_erase_sector(sector);
```
UncleRus: Why did you remove call to sdk_spi_flash_erase_sector()? It's nesessary to erase flash page before writing.

apiel: Hmm, in my case, it is actually working without. I will try tomorrow to put it back and update the PR.

AndreaGreco: Try to write a with customized firmware a page of 0.
Then write your firmware it without erase.
This must show you that erase is required. If it works without we got a problem.
Because flash does not work without erase.

apiel:
I try to upload another firmware on the rboot_config.roms[1] -> 0x102000 

esptool.py -p /dev/ttyUSB0 --baud 115200 write_flash -fs 16m -fm dout -ff 40m 0x102000 ./firmware/firmware.bin

And it does like I upload it on 0x2000
It reboot on this new firmware, but rboot_config.current_rom is 0 and the address is 0x2000 even if I uploaded it to 0x102000

Any idea?

AndreaGreco:
This is a wdog Reset, that means that your IDLE Task does not ping wdog.
The perfect is a API for manually ping wdog but at now is not available.
Checkout that you do not remove delay. Or consider increase delay.

apiel:
I actually wonder if it is not a problem with rboot. I try to change the flashsize setting to 0x80000, 0x40000 even 0x20000 but none of them work. 

esptool.py -p /dev/ttyUSB0 --baud 115200 write_flash -fs 16m -fm dout -ff 40m 0x0 ../esp-open-rtos/bootloader/firmware/rboot.bin 0x1000 ../esp-open-rtos/bootloader/firmware_prebuilt/blank_config.bin 0x2000 ./firmware/firmware156.bin 0x42000 ./firmware/firmware157.bin

I have to stop for today cause my head is explading :p 
Hopefully, I will find the solution soon. I will try to take in consideration your comments to make it work :D

AndreaGreco:
Do you make a sleep before reset ESP. Make a large sleep before reset is important for ping wdog.

apiel:
Actually, it's even more a question of OTA or not, I just have to find the right setting for flashing my ROMs. I tried so much different setting but I don't manage to get it work :-/

esptool.py -p /dev/ttyUSB0 --baud 115200 write_flash -fs 8m -fm dout -ff 40m 0x0 ../esp-open-rtos/bootloader/firmware/rboot.bin 0x1000 ../esp-open-rtos/bootloader/firmware_prebuilt/blank_config.bin 0x2000 ./firmware/firmware.bin 0x82000 ./firmware/firmware164.bin

Here, I manage to upload the 2 firmwares correctly but when I switch to the second room, the boot fails.

apiel:
I spent the whole day on my issue. I even made a new version of rBoot, I start to understand more and more all the logic of this booting system. But it seem that there is an issue when we boot from another address than 0x2000. I cleaned up all the flash and then I run the following command (with original rBoot). I took care to leave 0x2000 that rBoot switch to the next rom address. Everything is fine, but when it come to boot, nothing happen, no error, nothing, just stuck there.
```
 #:~/dev/esp/sonoff$ esptool.py -p /dev/ttyUSB0 --baud 115200 write_flash -fs 8m -fm dout 0x0 ../esp-open-rtos/bootloader/firmware/rboot.bin 0x82000 ./firmware/firmware.bin && cu -l /dev/ttyUSB0 -s 115200
esptool.py v1.2
Connecting...
Running Cesanta flasher stub...
Flash params set to 0x0320
Writing 4096 @ 0x0... 4096 (100 %)
Wrote 4096 bytes at 0x0 in 0.4 seconds (89.6 kbit/s)...
Leaving...
Connected.
load 0x40100000, len 2412, room 16 
tail 12
chksum 0xf0
ho 0 tail 12 room 4
load 0x3ffe8000, len 800, room 12 
tail 4
chksum 0xdc
csum 0xdc

rBoot v1.4.0 - richardaburton@gmail.com
Flash Size: 8 Mbit
Flash Mode:   DOUT
Flash Speed:  40 MHz
rBoot Option: Big flash
rBoot Option: RTC data

Rom 0 at 2000 is bad.
Booting rom 1 from 82000, run at ram c0620.

```

oremic:
hi 
resolve problem ??

apiel:
Long time I didn't touch sonoff and especially OTA, but from what I remember I had lot of issue with erasing the flash before to write the new rom. So the best way was to prepare the flash when the sonoff start. Also I had some issue to make the sonoff boot at the second rom position, so to solve this issue I wrote my own booting system. However, OTA with esp-open-rtos is a pain and if you really need OTA, I would highly recommand you to use Arduino.

PR #671, opened 2018-08-07
Shrink gpio<>iomux maps

quietboil:
Both maps are in RAM, so there is really no need to allocate extra memory for each entry - no aligned reads.

The assembler for `gpio_to_iomux` and `iomux_to_gpio` differs only in how array entry address (offset) is calculated:
- current: `addx4 a2, a2, a3`
- changed: `add.n a2, a3, a2`

### DIFF
--- a/core/esp_iomux.c
+++ b/core/esp_iomux.c
@@ -7,8 +7,8 @@
 #include "esp/iomux.h"
 #include "common_macros.h"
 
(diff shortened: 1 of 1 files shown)


ourairquality:
This might be a problem as an 8 bit access to iram generates a load exception, very slow, and perhaps some code will expected gpio access to be fast. So it might be best to either revert this, or change it to place these in data RAM?

quietboil:
> an 8 bit access to iram generates a load exception

Agree. My fault. I managed to miss the section attribute on those arrays :disappointed:

UncleRus:
Reverted

PR #676, opened 2018-08-21
Fixed sysparams flash area address calculation

quietboil:
As the `addr` is not a pointer adding a size of the `buffer` in words calculates a wrong offset in flash (+8 instead of +32). This eventually leads to the loop overstepping the boundary of the sysparam area and include the beginning of the esp system parameters sector - the last read from 0x003fafe8 to 0x003fb007 with a 4M flash. The latter is not empty, so the sysparam area checker would always see the sysparam area as not empty.


UncleRus:
Nice catch!

PR #678, opened 2018-08-22
Add ADS1015 support to ads111x lib.

k-korn:
ADS101x is a faster version of ADS111x ADC, with lower precision.
This change provides basic support for device, by defining sample rates used by ADS101x and providing new ads101x_get_value() function with proper 4-bit shift to obtain actual ADC value.


UncleRus:
Thank you!

PR #683, opened 2018-10-11
Fixed command and address in LSB mode

quietboil:
Not even sure if anyone cares :smiley: I, for instance, have not seen a device that uses LSB bit ordering, but... for the sake of completeness and what if...

Something I noticed when I was "playing" with different transfer modes. To show what it is I wrote a small test program:
```c
enum { LSB, MSB };
enum { USING_CS, MINIMAL_PINS };
// -----8<----->8------
spi_init(1, SPI_MODE0, SPI_GET_FREQ_DIV(2000,40), LSB, SPI_LITTLE_ENDIAN, USING_CS);
for (;;) {
switch (getchar()) {
    case 'c': {
        spi_set_command(1,4,0xC);
        spi_set_dummy_bits(1,4,false);
        spi_transfer_8(1, 0x6a);
        spi_clear_dummy(1);
        spi_clear_command(1);
        break;
    }
    case 'a': {
        spi_set_address(1,4,0xA);
        spi_set_dummy_bits(1,4,false);
        spi_transfer_8(1, 0x6a);
        spi_clear_dummy(1);
        spi_clear_address(1);
        break;
    }
}
```
Current implementation assumes "always MSB" for commands and addresses and pushes wrongs bits out.
## Command and Address
image
> **Note** that SPI analyzer is set to decode 8-bit transfers with the least significant bit first.
> also... while I ran them separately and captured 2 images, one for the command and one for the address, they are virtually identical

The fixed `spi_set_command` and `spi_set_address` set those bits properly.
> **Note** this time I set decoder to 4-bits per transfer to illustrate the result a bit better

## Command
image

## Address
image

### DIFF
--- a/core/include/esp/spi.h
+++ b/core/include/esp/spi.h
@@ -293,9 +293,18 @@ static inline void spi_set_command(uint8_t bus, uint8_t bits, uint16_t data)
 {
     if (!bits) return;
 
-    SPI(bus).USER0 |= SPI_USER0_COMMAND;                           //enable COMMAND function in SPI module
-    uint16_t command = data << (16 - bits);                        //align command data to high bits
-    command = ((command >> 8) & 0xff) | ((command << 8) & 0xff00); //swap byte order
+    SPI(bus).USER0 |= SPI_USER0_COMMAND;                                //enable COMMAND function in SPI module
+    uint16_t command;
+    // Commands are always sent using little endian byte order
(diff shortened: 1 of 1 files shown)


UncleRus:
@Zaltora please look

Zaltora:
It is look like good, unfortunately, I can't  test it right now.

PR #723, opened 2019-06-12
Http download issue when connection lost

Metalus:
I was testing HTTP OTA and I realized that OTA locks down when Wifi connection drops. When this happens, POSIX function ```read``` (located at ```HttpClient_download```) returns -1. Since it's implemented as unsigned, ```do { ... } while(read_byte > 0)``` loops forever. Therefore the ```read_byte``` variable must be signed type.


kanflo:
Thanks!

PR #732, opened 2019-10-03
Improve wificfg

zub2:
Several small improvements of extras/wificfg:
* add missing #includes to wificfg.h
* use #if instead of #ifdef when checking for configUSE_TRACE_FACILITY
* make default ssid, password and hostname overridable by making the definitions in wificfg.c weak
* fix typo in comment

### DIFF
--- a/extras/wificfg/wificfg.c
+++ b/extras/wificfg/wificfg.c
@@ -1540,7 +1540,7 @@ static int handle_wificfg_challenge_post(int s, wificfg_method method,
     return wificfg_write_string(s, http_redirect_header);
 }
 
-#ifdef configUSE_TRACE_FACILITY
+#if configUSE_TRACE_FACILITY
 static const char *http_tasks_content[] = {
 #include "content/tasks.html"
 };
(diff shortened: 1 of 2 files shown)


extras/wificfg/wificfg.c
```diff
@@ -54,11 +54,11 @@
 #endif
 
 
-const char *wificfg_default_ssid = "EOR_%02X%02X%02X";
-const char *wificfg_default_password = "esp-open-rtos";
-const char *wificfg_default_hostname = "eor-%02x%02x%02x";
+const char *wificfg_default_ssid __attribute__ ((weak)) = "EOR_%02X%02X%02X";
+const char *wificfg_default_password __attribute__ ((weak)) = "esp-open-rtos";
+const char *wificfg_default_hostname __attribute__ ((weak)) = "eor-%02x%02x%02x";
```
ourairquality: These don't need to be 'weak', they are variables that can be trivially modified, just set them in your app initialization code.

extras/wificfg/wificfg.c
```diff
@@ -1540,7 +1540,7 @@ static int handle_wificfg_challenge_post(int s, wificfg_method method,
     return wificfg_write_string(s, http_redirect_header);
 }
 
-#ifdef configUSE_TRACE_FACILITY
+#if configUSE_TRACE_FACILITY
```
ourairquality: Yes, that looks right, thanks.

extras/wificfg/wificfg.c
zub2: You are right. I didn't realize it's a `const char*`, not a `const char[]`.

I was motivated by several things:
* I thought I can't really change it (which was wrong)
* I didn't want the literals from the wificfg.c to be present in the binary at all (just a few bytes, but still needless if I'm using my own strings)
* when re-defining a weak symbol, there is no need to do anything at runtime, linker just handles that (yes, just 3 assignments, but when just a declaration can suffice, it seems nicer)

Because the type is `const char*`, even the second point was not fulfilled. If the type is changed to `const char[]`, then it seems to do what I wanted.

What do you think about keeping these weak and making them `const char[]`? Or are the last two points from the list above not interesting to you?

PR #737, opened 2019-11-29
Fix si7021 temperature/humidity sensor reading

WizBangCrash:
Raw reading modified according to the Silicon Labs sample code for the Si7021 sensor.

I've had issues trying to use the Si7021 sensor with Sonoff TH10 and TH16 devices, so looked into why it was always returning TEMP = 0, HUM = 2. The raw vales were not being interpreted as per the data sheet and Linux sample code that Silabs provide.

### DIFF
--- a/extras/dht/dht.c
+++ b/extras/dht/dht.c
@@ -139,6 +139,9 @@ static inline int16_t dht_convert_data(dht_sensor_type_t sensor_type, uint8_t ms
             data = 0 - data;       // convert it to negative
         }
     }
+    else if (sensor_type == DHT_TYPE_SI7021) {
+        data = msb * 256 + (lsb & ~0x3);
+    }
     else {
         data = msb * 10;
     }


WizBangCrash:
Apologies. Hold off reviewing this pull request until I update it.
I realise I haven't catered for negative temperatures (its cold in the UK today) :-)
I'll update my commit later today.

WizBangCrash:
Updates committed and and tested on Sonoff TH16 with DHT22, Si7021 & AM2301 sensors.

RavenSystem:
@WizBangCrash Did you test it with DHT11 too?

WizBangCrash:
I don't have a DHT11 sensor unfortunately, so I couldn't test that option.

WizBangCrash:
@RavenSystem One other thing I have noticed during testing is that the new Sonoff branded AM2301 sensors behave like the Si7021, but the older ASAIR branded AM2301 sensors behave like the DHT22. Both sensors are in identical plastics accept for the information printed on the rear.
This can get quite confusing, especially as the software uses the sensor name as the option. It might be wise to refactor this at some point in the future and have generic 'types' e.g. `SENSOR_TYPE_A`, `SENSOR_TYPE_B` and then list the sensors that can be used with each 'type'. Rather than `DHT_TYPE_DHT22`, `DHT_TYPE_SI7021`

RavenSystem:
Change sensor definitions is not a good idea because a lot of people uses this SDK, and everybody that uses DHT driver will must change code.

I think I have a DHT11. I will test it asap.

WizBangCrash:
> Change sensor definitions is not a good idea because a lot of people uses this SDK, and everybody that uses DHT driver will must change code.

I'm sure I could make it backwards compatible. I'm just forward thinking as more sensors come to the market.  I also looked at automatically switching between modes as I had in initial fix that tried the AM2301 using the old way and when it failed Phase B it would retry Phase A with a longer period. This worked, but I wasn't sure whether to submit it as a fix or not.

PR #760, opened 2020-08-16
Fix full on/off in pca9685.

mjkillough:
It's currently not possible to toggle between full off and full on, as
switching to full on leaves the full off control bit set. The full off
control bit takes precedence according to the datasheet, which means the
signal remains off. (See [datasheet] page 21 for register definitions and
page 24 for information about precedence).

[datasheet]: https://cdn-shop.adafruit.com/datasheets/PCA9685.pdf

The 'normal' branch does correctly clear full off/full on, as it writes
0s to the `LEDn_ON` registers unconditionally and writes 0 to the control
bit in `LEDn_OFF`.


UncleRus:
Thank you!

